PR #3, opened 2021-10-29
Add CICD publication step for PyPi

flaviuvadan:
Currently, users have to install Hera from this repository. Version management can be a bit cumbersome for that as the GitHub links are not super friendly. This PR adds a CICD step for publishing `hera-workflows` to PyPi. This will make installation accessibility easier as users can specify `hera-workflows` in the `install_requires` field of their applications' `setup.py`, `requirements.txt`, or the requirements field of any other dependency management tool they use.

### DIFF
--- a/.github/workflows/hera_build.yaml
+++ b/.github/workflows/hera_build.yaml
@@ -1,0 +1,42 @@
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - 'VERSION'
+jobs:
+  hera-main-build:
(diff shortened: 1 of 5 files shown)


.github/workflows/hera_build.yaml
```diff
@@ -0,0 +1,42 @@
+on:
+  push:
+    branches:
+      - main
+    paths:
```
alexec, line 5: I did not know about this

PR #5, opened 2021-11-01
Add service account token authentication from k8s config

rodrigobaron:
This show an example of how authenticate using kubernetes configuration file.

### DIFF
--- a/examples/k8s_sa.py
+++ b/examples/k8s_sa.py
@@ -1,0 +1,45 @@
+"""This example showcases the hello world example of Hera using k8s service account token"""
+from hera.v1.task import Task
+from hera.v1.workflow import Workflow
+from hera.v1.workflow_service import WorkflowService
+
+from kubernetes import client, config
+import base64
+
+
+def get_sa_token(service_account: str, namespace: str = "default", config_file=None):
+    """Get ServiceAccount token using kubernetes config.
+
+     Parameters
+    ----------
+    service_account: str
+        The service account to authenticate from.
+    namespace: str = 'default'
+        The K8S namespace the workflow service submits workflows to. This defaults to the `default` namespace.
+    config_file: str
+        The path to k8s configuration file.
+    """
+    config.load_kube_config(config_file=config_file)
+    v1 = client.CoreV1Api()
+    secret_name = (
+        v1.read_namespaced_service_account(service_account, namespace).secrets[0].name
+    )
+    sec = v1.read_namespaced_secret(secret_name, namespace).data
+    return base64.b64decode(sec["token"]).decode()
+
+
+namespace = "argo"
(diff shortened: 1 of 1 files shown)


examples/k8s_sa.py
```diff
@@ -0,0 +1,45 @@
+"""This example showcases the hello world example of Hera using k8s service account token"""
+from hera.v1.task import Task
+from hera.v1.workflow import Workflow
+from hera.v1.workflow_service import WorkflowService
+
+from kubernetes import client, config
+import base64
+
+
+def get_sa_token(service_account: str, namespace: str = "default", config_file=None):
```
flaviuvadan, line 10: I am aware this is an example but curious what you think about adding a type for the `config_file` field as a `str`, perhaps, or `Path`, and checking for its existence to throw an exception if the file is not there? I think it might help users out by covering a potential failure case.

examples/k8s_sa.py
```diff
@@ -0,0 +1,45 @@
...
+    v1 = client.CoreV1Api()
+    secret_name = (
+        v1.read_namespaced_service_account(service_account, namespace).secrets[0].name
+    )
+    sec = v1.read_namespaced_secret(secret_name, namespace).data
+    return base64.b64decode(sec["token"]).decode()
+
+
+namespace = "argo"
+token = get_sa_token("argo-server", namespace=namespace)
```
flaviuvadan, line 32: Any chance we can position these variables right before the `WorkflowService` instantiation? It might help keep everything relevant for the workflow submission example neatly in the same block 🙂

flaviuvadan (commented):
Thank you for submitting this! This is an awesome example. I didn't even know we could submit via K8S SA token 😄

examples/k8s_sa.py
rodrigobaron: yes, you are right it should be `Optional[str]`, the [expected input is 'str'](https://github.com/kubernetes-client/python-base/blob/master/config/kube_config.py#L48).. fixing it!

examples/k8s_sa.py
rodrigobaron: done :slightly_smiling_face:

examples/k8s_sa.py
```diff
@@ -0,0 +1,57 @@
...
+):
+    """Get ServiceAccount token using kubernetes config.
+
+     Parameters
+    ----------
+    service_account: str
+        The service account to authenticate from.
+    namespace: str = 'default'
+        The K8S namespace the workflow service submits workflows to. This defaults to the `default` namespace.
+    config_file: str
```
flaviuvadan, line 25: Small suggestion:

```suggestion
    config_file: Optional[str] = None
```

flaviuvadan (approved):
Thank you again for submitting this!

PR #8, opened 2021-11-02
Features/when

rodrigobaron:
This implement the [conditional feature](https://github.com/argoproj/argo-workflows/blob/master/examples/conditional-parameters.yaml) named `when` around the wrapper of `V1alpha1DAGTask` and include the coin-flip example similar as [couler](https://github.com/couler-proj/couler/blob/master/examples/coin_flip.py).


flaviuvadan:
First, this is a great addition so thank you for submitting the PR! 🙂

There are some aspects of the implementation of this PR that I have feedback on but I also want to mention a tiny backstory. One of the philosophies of Hera is simplicity. When @dynotx started using Workflows all the libraries were too complex to use... and they were complex because of this specific type of syntax:
- `{{inputs.parameters.param2}}`
- `{{task1.outputs.results}}`
- etc.

This type of specification is what prevented a lot of internal adoption of Argo Workflows. People really liked Argo Workflows but they did not like writing, submitting, reading, and reviewing workflows (in PRs and the codebase). So, Hera is focused on abstracting away those details and supports users in writing definitions that are very focused on the holistic concept of a task and workflow properties rather than very, very specific details, particularly syntax-wise. Having said that, what do you think are some simpler ways of implementing this interface? I am thinking about a few things:
- use `when` rather than `next_when`. AFAIK, there's nothing else in Argo Workflows that can act as another "when" conceptually; is this too generic?
- rather than have the conditions written out as strings we could implement the interface of `when` to be `when(task_result: str, operator: Operator, value: str)`. This requires:
  - adding a `result` property on a task that returns the `{{task.output.result}}` string
  - building an `Operator` that supports specific operators like `==`

What do you think about the above suggestions? The intent is to help users in specifying this conditional execution:
```python
ws = WorkflowService('my-argo-server.com', 'my-auth-token')
w = Workflow("coin-flip", ws)
r = Task("r", random_code)
h = Task("h", heads)
t = Task("t", tails)
h.when(r.result, Operator.equals, "heads")
t.when(r.result, Operator.equals, "tails")
```

I think this kind of flips the logic of the dependency setting, though, so perhaps a combination of what you proposed and the above would fit best? Thoughts?

flaviuvadan:
Another alternative is to expand the `next` function and add a condition field to it! 
```python
ws = WorkflowService('my-argo-server.com', 'my-auth-token')
w = Workflow("coin-flip", ws)
r = Task("r", random_code)
h = Task("h", heads)
t = Task("t", tails)
r.next(h, r.result, Operator.equal, "heads")
r.next(t, r.result, Operator.equal, "tails")
```

This is more in line with the idea of a "next" step after executing a current one.

rodrigobaron:
thanks for the complete explanation :slightly_smiling_face:
but not sure if could we evaluate the `r` task using `h.when(r.result, Operator.equals, "heads")` API, because the function don't have information about the condition dependency. So we can't reconstruct `'{{tasks.r.outputs.result == "heads"}}'`.

I think need something like `when(task='r', target='result', operator=Operator.equals, value='heads')`.

flaviuvadan:
> thanks for the complete explanation 🙂 but not sure if could we evaluate the `r` task using `h.when(r.result, Operator.equals, "heads")` API, because the function don't have information about the condition dependency. So we can't reconstruct `'{{tasks.r.outputs.result == "heads"}}'`.
> 
> I think need something like `when(task='r', target='result', operator=Operator.equals, value='heads')`.

Yes, we're almost 100% on the same page! What do you think about something like:
```Python
    def when(self, other: 'Task', operator: 'Operator', value: str):
        self.argo_task.when = f'{{{{tasks.{other.name}.outputs.result}}}} {operator} {value}'
        self.next(other)
```

I think the target `result` field is consistent in Argo Workflows? I do not have too much experience with conditional execution but, from the documentation, it seems that the result field is always named `result`, which allows Hera to always use the same target, as you suggested!

flaviuvadan:
@rodrigobaron any thoughts on the above? Wondering if we're going ahead with this PR 🙂  If it just happens that you do not have time for this, I can certainly implement it and have you review it!

rodrigobaron:
Sorry for the delay. I've implemented in the such way.. now just worrying about the 'Operator' implementation.

src/hera/v1/operator.py
```diff
@@ -0,0 +1,33 @@
+class Operator:
```
flaviuvadan, line 1: This is awesome! I do have a suggestion, though. I think we can simplify this to an enum, like so:
```python
class Operator(str, Enum):
    equals = '=='
    greater = '>'
    lower = '<'
    greater_equal = '>='
    lower_equal = '<='
    not_equal = '!='
``` 

Let me know what you think!

src/hera/v1/task.py
```diff
@@ -163,6 +164,28 @@ def __rshift__(self, other: 'Task') -> 'Task':
         """
         return self.next(other)
 
+    def when(self, other: 'Task', operator: 'Operator', value: str):
```
flaviuvadan, line 167: We can probably import the `Operator` here and use it as the type rather than use an annotation 🤔

src/hera/v1/task.py
```diff
@@ -163,6 +164,28 @@ def __rshift__(self, other: 'Task') -> 'Task':
...
+        Task
+            The other task that was specified.
+
+        Examples
+        --------
+        t1, t2, t3 = Task('t1'), Task('t2'), Task('t3')
+        t2.when(t1, Operator.equals, "t2")
+        t3.when(t1, Operator.equals, "t3")
+        """
+        self.argo_task.when = f'{{{{tasks.{other.name}.outputs.result}}}} {operator} {value}'
```
flaviuvadan, line 186: If we switch the `Operator` to an `Enum`, then this becomes: 
```suggestion
        self.argo_task.when = f'{{{{tasks.{other.name}.outputs.result}}}} {operator.value} {value}'
```

flaviuvadan (commented):
I tested the implementation and it works fine!

Screen Shot 2021-11-25 at 11 36 30

src/hera/v1/operator.py
rodrigobaron: I've tried that and got TypeError, because `when`function is expecting `Operator` and if we do `.when(t1, Operator.equals, "t2"` we give an `str` input instead.

rodrigobaron: I didn't used `Enum` in this case.. i'll try that!

rodrigobaron: I've updated the implementation, if we use `Operator(str, Enum):` then:
```bash
$ make verify
flake8 src
mypy --namespace-packages --explicit-package-bases -p hera
src/hera/v1/operator.py:17: error: Incompatible types in assignment (expression has type "str", base class "str" defined the type as "Callable[[str], str]")
Found 1 error in 1 file (checked 17 source files)
```
just `Enum` works fine..

src/hera/v1/operator.py
```diff
@@ -0,0 +1,20 @@
...
+     value: str
+         The value of an operator in str.
+    """
+
+    equals = '=='
+    greater = '>'
+    lower = '<'
+    greater_equal = '>='
+    lower_equal = '<='
+    not_equal = '!='
```
flaviuvadan, line 20: Very picky suggestion:
```suggestion
class Operator(Enum):
    """Operator is a representation of mathematical comparison symbols.

     This can be used on tasks that execute conditionally based on the output of another task.

    Notes
    ------
    The task that outputs its result needs to do so using stdout. See `examples` for a sample workflow.
    """

    equals = '=='
    greater = '>'
    less = '<'
    greater_equal = '>='
    less_equal = '<='
    not_equal = '!='
```

flaviuvadan (commented):
This looks awesome! Remember to add a CHANGELOG entry and bump the minor version in VERSION as this is a new feature 🙂

rodrigobaron:
I didn't follow, the next version is `v0.3.0` should skip that and go to `v0.4.0`?
obs.: I've added the the version `0.3.0`

rodrigobaron:
ok, I did the sync with main and merge into my branch.

src/hera/v1/task.py
```diff
@@ -163,6 +164,28 @@ def __rshift__(self, other: 'Task') -> 'Task':
         """
         return self.next(other)
 
+    def when(self, other: 'Task', operator: Operator, value: str):
```
flaviuvadan, line 167: Small:
```suggestion
    def when(self, other: 'Task', operator: Operator, value: str) -> 'Task':
```

flaviuvadan (approved):
Just a small suggestion left for a type specification in the function signature of `when`. Thank you so much for adding this @rodrigobaron, and all of your patience with the discussions! This is awesome 🙂

PR #9, opened 2021-11-03
cron workflows first sketch

NI1993:
Adds cron workflows support in hera


src/hera/v1/cron_workflow.py
```diff
@@ -0,0 +1,56 @@
...
+        service: WorkflowService,
+        cron_service: CronWorkflowService,
+        parallelism: int = 50,
+    ):
+        super().__init__(name, service, parallelism)
+        self.schedule = schedule
+
+        self.cron_workflow_spec = V1alpha1CronWorkflowSpec(schedule=self.schedule, workflow_spec=self.spec)
+        self.cron_workflow = V1alpha1CronWorkflow(metadata=self.metadata, spec=self.cron_workflow_spec)
+        self.cron_service = cron_service
```
flaviuvadan, line 45: What do you think about overwriting the `service`, `workflow`, and `spec` properties of the `Workflow`? I think combining properties specific to `Workflow` and `CronWorkflow` (conceptually) into `CronWorkflow` sort of breaks the unique responsibility of one vs. another. If `CronWorkflow` overwrites those properties, then `submit` can maintain the same interface, but execute with a different underlying service, which helps the client avoid potential problems with the submission of the workflow! I don't know the internals of each service (workflow vs. cron workflow) but I assume there are some specific fields different between them that _might_ break functionality if we combine the classes. What do you think? 🙂

src/hera/v1/cron_workflow_service.py
```diff
@@ -0,0 +1,138 @@
...
+        self.service = CronWorkflowServiceApi(api_client=api_client)
+
+    def submit(self, cron_workflow: V1alpha1CronWorkflow, namespace: Optional[str] = None) -> V1alpha1CronWorkflow:
+        """Submits the given cron workflow to the given namespace.
+
+        Parameters
+        ----------
+        cron_workflow: V1alpha1CronWorkflow
+            The cron workflow to submit.
+        namespace: optional str
```
flaviuvadan, line 43: Any chance the type specifications can use the `typing` package references? It provides neat consistency and clarity in documentation. It might also be nice to use something like auto-generated docs in the future to show these, and the types are nice 🙂 
```suggestion
        namespace: Optional[str] = None
```

src/hera/v1/cron_workflow_service.py
```diff
@@ -0,0 +1,138 @@
...
+            The K8S namespace of the Argo server to submit the cron workflow to.
+
+        Raises
+        ------
+        argo.workflows.client.ApiException
```
flaviuvadan, line 48: Any chance a brief explanation of the circumstances under which this exception is raised be provided?

src/hera/v1/cron_workflow.py
```diff
@@ -0,0 +1,56 @@
...
+from argo.workflows.client import (
+    V1alpha1CronWorkflow,
+    V1alpha1CronWorkflowSpec,
+)
+from hera.v1.cron_workflow_service import CronWorkflowService
+from hera.v1.workflow_service import WorkflowService
+from hera.v1.workflow import Workflow
+
+
+class CronWorkflow(Workflow):
```
flaviuvadan, line 14: I would add some simple tests to ensure that the right fields are set on this class!

flaviuvadan:
Awesome! Thank you for submitting this 🙂 I left a few questions to get your thoughts on some parts of the implementation and some suggestions on documentation.

src/hera/v1/cron_workflow_service.py
NI1993: I may need your help here as it seems to me that only http related errors may be raised by this implementation.
One edge case I could think of is supplying a namespace which does not exist.

src/hera/v1/cron_workflow.py
NI1993: Interesting point :)
as the dos of argo states: _"In essence, CronWorkflow = Workflow + some specific cron options"_
(https://argoproj.github.io/argo-workflows/cron-workflows/)
That being said - I'm not sure what's the best way of translating that relation into code.

I understand & agree with your comments, but perhaps there is a better way.
How about composition?

src/hera/v1/cron_workflow.py
NI1993: Will do as soon as we agree on implementation

NI1993:
> Awesome! Thank you for submitting this 🙂 I left a few questions to get your thoughts on some parts of the implementation and some suggestions on documentation.

Thanks for your time :)
Will fix the remaining open points as soon as we can agree on acceptable implementation.

src/hera/v1/cron_workflow.py
flaviuvadan: Ok, very cool! I did not notice that previously, thank you for sharing! One thing I am wary of, though, is allowing the explanation of `CronWorkflow = Workflow + cron options` dictate code patterns. I view it as more of a very accessible explanation of the cron type as I am not sure whether the Argo Workflows code actually reflects that structure.

I see a few options going forward:
- use composition to set the workflow type inside the cron. This requires extracting specific fields, like tasks, off of the workflow type, which adds unnecessary complexity and complicates the cron type a bit
- use inheritance with field overwrites. This maintains a pretty clean interface and avoids some level of code duplication by taking advantage of the provided functionality of the workflow type. This masks away the behavior of the parent workflow type, though, and it might confuse contributors at first as it _might_ be challenging to find exactly which field is used by one over another
- avoid inheritance of workflow by cron and add a base class that acts as the DAG task container, which workflow and cron inherit the `add_task/s` and `add_head/tail` from. This requires a bit of a rewrite that we should probably not do until V2 perhaps.
- avoid inheritance and composition completely. Duplicate the `add_task/s` and `add_head/tail` to cron workflow. Duplication is ok, but when Hera ends up supporting `Step`, for example, it'll need to duplicate the adjustments of `add_step/s` perhaps

I am most in favor of the second option because of verbosity and some savings of code duplication to favor additions of other constructs (future `Step`). In my XP using very verbose implementations, along with some level of code duplication, makes it very easy to pick up something, and my hope is that Hera can showcase that as well. The inheritance, in this case, is quite accessible, and makes it decently clear that some methods are overwritten in order to adjust the used types. Thoughts?

NI1993: Thanks for clearing this up.
I'm not sure though that option #2 is safe.
Perhaps I do not fully understand it, but If the code would go like:

```
class CronWorkflow(Workflow):
    def __init__(
        self,
        name: str,
        schedule: str,
        service: WorkflowService,
        cron_service: CronWorkflowService,
        parallelism: int = 50,
    ):
        super().__init__(name, service, parallelism)
        self.schedule = schedule

        self.workflow = V1alpha1CronWorkflow(metadata=self.metadata, spec=self.spec)
        self.spec = V1alpha1CronWorkflowSpec(schedule=self.schedule, workflow_spec=self.spec)
        self.service = cron_service

    def suspend(self, name: str, namespace: Optional[str] = None):
        self.service.suspend(name, namespace)

    def resume(self, name: str, namespace: Optional[str] = None):
        self.service.resume(name, namespace)

    def submit(self, namespace: Optional[str] = None) -> None:
        self.service.submit(self.workflow, namespace)
```

Then the implementation of `CronWorkflow.add_tasks` would break - since its implementation would refer to `self.spec`. Since it expects it to be a `V1alpha1WorkflowSpec` but got `V1alpha1CronWorkflowSpec` it would crash.

Again, maybe I can't see what you do...
Can u demonstrate your suggestion by code?

src/hera/v1/cron_workflow_service.py
NI1993: Certainly, will update that

src/hera/v1/cron_workflow.py
flaviuvadan: Ah, yes, you are _definitely_ correct! My apologies for the confusion 🙁  this discussion was extremely valuable! I've tried this myself and ended up going with the 3rd option enumerated above. It makes the code very clean, actually, and as mentioned I think duplication is ok for now until we learn better how to structure these. I keep going back to the explicit vs. implicit and simplicity principles behind Hera and I think the verbosity of an independent cron workflow implementation is very nice! Here's what I wrote and tested in an Argo cluster:

```python
from datetime import datetime, timezone
from typing import Optional
from uuid import uuid4

from argo.workflows.client import *

class CronWorkflowService:

    # TODO: this should go in cron_workflow_service.py
    def __init__(self, domain: str, token: str):
        self.domain = domain
        api_client = Client(Config(domain), token).api_client
        self.service = CronWorkflowServiceApi(api_client=api_client)

    def submit(self, cron_workflow: V1alpha1CronWorkflow, namespace: str = 'default') -> V1alpha1CronWorkflow:
        return self.service.create_cron_workflow(
            namespace, V1alpha1CreateCronWorkflowRequest(cron_workflow=cron_workflow, namespace=namespace)
        )

    def delete(self, name: str, namespace: str = 'default') -> Tuple[object, int, dict]:
        return self.service.delete_cron_workflow(namespace, name)

    def suspend(self, name: str, namespace: str = 'default') -> Tuple[object, int, dict]:
        return self.service.suspend_cron_workflow(
            namespace, name, body=V1alpha1CronWorkflowSuspendRequest(name=name, namespace=namespace)
        )

    def resume(self, name: str, namespace: str = 'default') -> Tuple[object, int, dict]:
        return self.service.resume_cron_workflow(
            namespace, name, body=V1alpha1CronWorkflowResumeRequest(name=name, namespace=namespace)
        )

    def get_cron_workflow_link(self, name: str, namespace: str = 'default') -> str:
        return f'https://{self._domain}/cron-workflows/{namespace}/{name}'


class CronWorkflow:
    # TODO: this should go in cron_workflow.py
    def __init__(
        self,
        name: str,
        schedule: str,
        service: CronWorkflowService,
        parallelism: int = 50,
    ):
        self.name = f'{name.replace("_", "-")}-{str(uuid4()).split("-")[0]}'
        self.schedule = schedule
        self.service = service or CronWorkflowService()
        self.parallelism = parallelism

        self.dag_template = V1alpha1DAGTemplate(tasks=[])
        self.template = V1alpha1Template(name=self.name, steps=[], dag=self.dag_template, parallelism=self.parallelism)
        self.metadata = V1ObjectMeta(name=self.name)
        self.spec = V1alpha1WorkflowSpec(templates=[self.template], entrypoint=self.name)

        self.cron_spec = V1alpha1CronWorkflowSpec(schedule=self.schedule, workflow_spec=self.spec)
        self.workflow = V1alpha1CronWorkflow(
            metadata=self.metadata,
            spec=self.cron_spec,
            status=V1alpha1CronWorkflowStatus(
                active=[], conditions=[], last_scheduled_time=datetime.now(timezone.utc)
            ),
        )

    def add_task(self, t: Task):
        # same implementation as the one in Workflow...
        ...

    def add_tasks(self, *ts: Task):
        # same implementation as the one in Workflow...
        ...

    def add_head(self, t: Task, append: bool = True):
        # same implementation as the one in Workflow...
        ...

    def add_tail(self, t: Task, append: bool = True):
        # same implementation as the one in Workflow...
        ...

    def suspend(self, name: str, namespace: str = 'default'):
        """Suspends the cron workflow"""
        self.service.suspend(name, namespace)

    def resume(self, name: str, namespace: str = 'default'):
        """Resumes execution of the cron workflow"""
        self.service.resume(name, namespace)

    def submit(self, namespace: str = 'default') -> None:
        try:
            self.service.submit(self.workflow, namespace)
        except ValueError as e:
            # the deserialization protocol of the V1alpha1CronWorkflow object creates a V1alpha1CronWorkflowStatus
            # that does not set the `active`, which cannot be `None`. Deserialization happens after this receives 
            # the HTTP response back and it raises a ValueError. This captures that specific error message and simply
            # returns as it's safe for now. Otherwise, it raises the same exception to propagate it to the client
            if str(e) == 'Invalid value for `active`, must not be `None`':
                return
            else:
                raise e
```

What do you think? IMHO, I find this _very_ easy to understand. I makes maintenance and optimization of cron workflows, independently, very easy. It also makes it very clear upon first sight, I think, which should facilitate usage.

Excited to hear what you think! 🙂

NI1993: Honestly, I like more the composition option over the code duplication.
But, as you kindly (😄 ) mentioned above you prefer the duplicates.

It's totally reasonable and respect your decision!

Nevertheless - please allow me to demonstrate just for the sake of it the composition option I mentioned:
```
class CronWorkflow:
    def __init__(
        self,
        name: str,
        schedule: str,
        service: WorkflowService,
        cron_service: CronWorkflowService,
        parallelism: int = 50,
    ):
        self._workflow = Workflow(name, service, parallelism)
        self.schedule = schedule

        self.cron_workflow_spec = V1alpha1CronWorkflowSpec(schedule=self.schedule, workflow_spec=self._workflow.spec)
        self.cron_workflow = V1alpha1CronWorkflow(metadata=self._workflow.metadata, spec=self.cron_workflow_spec)
        self.cron_service = cron_service

    @property
    def workflow(self) -> Workflow:
        return self._workflow

    def suspend(self, name: str, namespace: Optional[str] = None):
        """Suspends the cron workflow"""
        self.cron_service.suspend(name, namespace)

    def resume(self, name: str, namespace: Optional[str] = None):
        """Resumes execution of the cron workflow"""
        self.cron_service.resume(name, namespace)

    def submit(self, namespace: Optional[str] = None) -> None:
        self.cron_service.submit(self.cron_workflow, namespace)

    # These are optional - and matter of taste. User can always type cwf.workflow.add_* ...
    def add_task(self, *_, **__) -> None:
        """proxy method"""
        return self._workflow.add_task(*_, **__)

    def add_tasks(self, *_, **__) -> None:
        """proxy method"""
        return self._workflow.add_tasks(*_, **__)

    def add_head(self, *_, **__) -> None:
        """proxy method"""
        return self._workflow.add_head(*_, **__)

    def add_tail(self, *_, **__) -> None:
        """proxy method"""
        return self._workflow.add_tail(*_, **__)
```

Just to clarify: I'm using your draft as the final version.
Perhaps the snippet above may be useful in the future 🤔 ...
Thank you for listening to my feedback.

src/hera/v1/cron_workflow.py
NI1993: @flaviuvadan added test_cwf_valid_field_set

NI1993:
@flaviuvadan added also isort to the makefile - I think it will improve readability of the code and it plays well with black.
Still have not ran it against the whole project

src/hera/v1/cron_workflow.py
flaviuvadan: Thank you so much for illustrating the composition example! It's wonderful. I particularly love the optional proxies. Also, I must say I really appreciate the kindness of your message and understanding 🙂 As we add more feature parity with Argo Workflows, we'll eventually end up building a high level V2 of Hera. I will ping you, if you don't mind, for a discussion on how to organize the new version in the future - I think you'll have great input on the organization of the new version, composition, inheritance patterns, etc.

flaviuvadan:
> @flaviuvadan added also isort to the makefile - I think it will improve readability of the code and it plays well with black. Still have not ran it against the whole project

Awesome! We can run it in a different PR. Thank you for adding that as well!

flaviuvadan (commented):
Looks great! One last question before we merge - did you have the chance to run it in a cluster to functionally test it? If not, I can do it and report back.

NI1993:
> Looks great! One last question before we merge - did you have the chance to run it in a cluster to functionally test it? If not, I can do it and report back.

@flaviuvadan I just now tested it and noticed a bug in terminology I had - the submit method I implemented is actually creating the cron.
Submission is done in an other api call, which is not part of the `CronWorkflowServiceApi` so I'll leave it for now

src/hera/v1/cron_workflow_service.py
```diff
@@ -0,0 +1,134 @@
...
+        cron_workflow: V1alpha1CronWorkflow
+            The cron workflow to create.
+        namespace: str = 'default'
+            The K8S namespace of the Argo server to create the cron workflow in.
+
+        Raises
+        ------
+        argo.workflows.client.ApiException: Raised upon any HTTP-related errors
+        """
+        return self.service.create_cron_workflow(
```
flaviuvadan, line 51: I just realized the Argo SDK uses `create` vs. `submit`. Thank you for pointing it out via this implementation. We're going to have a `submit` on `workflow` and a `create` on cron now... that's fine, we'll change it in V2 🙂

flaviuvadan:
@NI1993 Sign the CLA and we're good to go :)

NI1993:
> @NI1993 Sign the CLA and we're good to go :)

I'm not sure what is that 🤷‍♂️

flaviuvadan (approved):
Awesome! Thank you so much for adding this and the willingness to engage in discussions! 🙂

PR #14, opened 2021-11-17
ran make isort

NI1993:
Simply executed `make isort` on terminal and pushed the result


flaviuvadan (approved):
Was just about to do this so thank you for being ahead of the game 🙂

PR #23, opened 2021-12-05
fix(examples): add missing import to coin_flip example

mobin102:


flaviuvadan (approved):
Oh, good catch! Should definitely make it easier to take this example "off the shelf". Thank you for the fix! 🙂

Issue #7, opened 2021-11-01
Migrate Hera to newly generate Argo Workflows OpenAPI client


flaviuvadan:
Currently, Hera builds on top of the now deprecated/archived [argo-client-python](https://github.com/argoproj-labs/argo-client-python). This client has moved to the main Argo Workflows repository [here](https://github.com/argoproj/argo-workflows/tree/master/sdks/python). Hera's constructs and imports should migrate to the new client. A new minor version should be generated ~- 0.0.0 -> 0.1.0 -~ as a dependency will be removed, another one introduced, but the interface does not change, only internals.

flaviuvadan:
Now that https://github.com/argoproj/argo-workflows/pull/7363 was merged, we can start thinking about migrating to the new SDK when a new Argo release is performed.

bobh66:
I could really use this change as I have a problem with the current SDK that won't be fixed.   It validates the CronWorkflowStatus object that is returned in the create response for CronWorkflow, and it "fails" validation because the "active" attribute is None (since no cronjobs have run yet) which causes a KeyError exception even though the object was created successfully.

Is the new SDK available in pypi?  I didn't see anything from a quick search.  If you can point me in the right direction I'll see what I can put together.

bobh66:
I should have looked further to see that you started it - let me know if I can be of any help.

flaviuvadan:
Hey @bobh66! There's a whole story to the initial effort to migrate. [Here's](https://github.com/argoproj-labs/hera-workflows/pull/46) the PR that attempted a second migration. The problem is that PyPi does not allow setting a dependency on a repo (as is currently the case with the new SDK). After a discussion with @terrytangyuan it's a bit unclear when the next release of Argo Workflows will take place, which inhibits the official release of the SDK. However, we've discussed using my fork of Argo Workflows, after it's up to date with the latest main branch of [Argo Workflows](https://github.com/argoproj/argo-workflows), to create a release candidate as `6.0.0rc0`, install that in Hera, and perform the migration. This will give us all the objects that are required, allow us to fix the problem you noticed with cron workflows, and allows us to just relock Hera to use the official SDK post publication to PyPi. Note that both `6.0.0rc0` and `6.0.0` would be officially released to PyPi, not to some projects that I am the sole owner of 🙂  I will try my best to get this done this upcoming weekend.

flaviuvadan:
[...]

PR #25, opened 2021-12-06
fix(cron-workflow): fix active and conditions nullity

mobin102:
Cron workflow example can not be executed properly due to `ValueError: Invalid value for active, must not be None`. I have added  `V1LocalObjectReference` to  `active` and `conditions` empty list of `V1alpha1CronWorkflowStatus`.


flaviuvadan:
Thank you so much for contributing again @mobin102 🚀  Is the `ValueError` thrown right by the `cron_workflow_service.create(...)` or the initialization phase of the `CronWorkflow` object?

flaviuvadan:
@mobin102 I have a PR in Argo Workflows to fix the autogenerated SDK: https://github.com/argoproj/argo-workflows/pull/7363 I think that will help solve the issue without us modifying Hera. We're going to be dependent on migrating to `argo-workflows==6.0.0`, which is https://github.com/argoproj-labs/hera-workflows/issues/7, but that should not be _that_ big of a change 🙂

flaviuvadan:
@mobin102 check out this line: https://github.com/argoproj-labs/hera-workflows/pull/38/files#diff-40b2cae22ec120b51dde27b367365d8ae1aa08f425cfd186da5c1712b31e5cefR60 That's the one that makes the SDK _not_ check the return type validity (it is a valid response, it just doesn't have _all_ the expected fields but that's fine).

flaviuvadan:
@mobin102 I just merged https://github.com/argoproj-labs/hera-workflows/pull/38! I am going to close this for now. Thank you for looking into this issue in the first place! I hope the new releases fixes the problems you noticed 🙂 New issues are always welcome!

PR #38, opened 2022-01-08
Upgrade underlying Argo SDK

flaviuvadan:
This PR tackles issue #7. In addition, it fixes some of the type issues that we encountered in the past with deserializing responses post workflow submission. It's also worth noting that the new underlying Argo SDK requires `limit` rather than `duration` and `max_duration` for retries. This introduces a `get_limit` in the `Retry` object that computes the limit based on `max_duration // duration`, which should typically correspond to the intended limit. There are some leftover TODOs that will be added as GitHub Issues after this is merged. 

It would be ideal to add some more tests to this prior to release but leaving it open for any watchers.


flaviuvadan:
I tested all the examples in an Argo cluster with version 3.2.4. Retries are fine, along with task parallelism and all currently supported features.

PR #46, opened 2022-01-10
Revert "Revert "Revert "Revert "Upgrade underlying Argo SDK""""

flaviuvadan:
The `dependency_links` adjustments did not fix the release. I deleted 0.4.1 quickly and now going to explore other ways of fixing the dependency problem. Perhaps it might be prudent to wait for the official release of the new SDK so it can be installed from PyPy as 6.0.0.

Reverts argoproj-labs/hera-workflows#45


harelwa:
Hi @flaviuvadan 

As there are features available to us only with the SDK, can we perhaps setup an `rc` branch [ perhaps a `2.0rc` ] we can reintroduce the changes you already made, and work in fixing them ?

Than we can either publish the **2.0rc** to pypi, or at least install from source and help stabilizing it.

Thanks,

Harel

Issue #40, opened 2022-01-09
Expose `verify_ssl` and `scheme` on `Configuration`


flaviuvadan:
Multiple users wish to use Hera but have control over certificate checking behavior by controlling the scheme and SSL verification configuration of Hera. Hera needs to expose the `verify_ssl` flag, along with allowing users to specify their preferred scheme, and domain fully.

dixon1e:
@flaviuvadan I am so sorry for dropping off the radar on this. COVID related events have hit our world very hard, as well as the Colorado fires. We appreciate your efforts so much.

A quick favor to ask? Can you add "hello world" examples to the repo demonstrating how to use the options we've just created?

dixon1e:
Just did a quick smoke test and it worked. Working up some more complex use cases. Brilliant, thank you!

flaviuvadan:
> @flaviuvadan I am so sorry for dropping off the radar on this. COVID related events have hit our world very hard, as well as the Colorado fires. We appreciate your efforts so much.
> 

No worries at all! Yes, they have, and with everything happening around COVID and the fires I am only sending good thoughts to Colorado ❤️ 

> A quick favor to ask? Can you add "hello world" examples to the repo demonstrating how to use the options we've just created?

Yes, definitely!

> Just did a quick smoke test and it worked. Working up some more complex use cases. Brilliant, thank you!

Glad this is settled! My pleasure 🙂 

> The imports are changed. Also showed usage of a different namespace to execute, and the verify_ssl flag

👍🏻

Issue #41, opened 2022-01-09
Remove internal `v1` module


flaviuvadan:
It is not necessary to version the internal modules of Hera for a few reasons:
- it makes semantic versioning more complicated than it needs to be (should Hera release a major version update if the underlying SDK changes? what if a single object has to change, such as the `Config` to modify the domain and `verify_ssl` flag?). Major package updates will reflect breaking changes, and the CHANGELOG will showcase what's available in a specific release that can be installed via PyPi;
- the introduction of `v2` would mean maintaining two internal version, or more in the future, and a public 3.x version potentially, which can be very confusing for users of all levels of experience. Hera needs to keep in mind that the purpose of it is to make it easy to adopt Argo Workflows. Understanding what a version of Hera means is critical for adoption and clarity/simplicity of installation
- reduce maintenance burden - when "v1" is conceptually eradicated, Hera can slowly move towards v2, rather than introduce a partially developed v2, which would not be a great experience for users (simplicity is key, so Hera should keep it simple)

flaviuvadan:
@dixon1e and @bartcode - I made this decision in order to facilitate making the changes you requested. In its current state I realized that Hera would _force_ us to release a V2 in order to introduce the interface changes that your uses cases require. Removing v1 (and the concept of internal library versioning) introduces a breaking change in which we can safely deploy the necessary SSL verification changes you requested. I am opening a PR soon to remove `v1`. Then I will open a PR to adjust the `Configuration` object in a branch off of the v1 removal branch and tag you in it.

flaviuvadan:
Expect a Hera 2.0.0 with a new `Config` that fixes the SSL thing and a package structure that is easier to understand 🙂

PR #24, opened 2021-12-05
feat: Add verify flag to workflow/cron workflow service

flaviuvadan:
There are use cases in which users have deployments with self-signed certificates. When port forwarding to an Argo deployment with such a certificate, Hera should instantiate the Argo config to skip certificate validation. By default, validation will be performed, but the `verify` flag allows users to control this behavior. 

This fixes #16.

Issue #16, opened 2021-11-22
How to ignore certificate errors?


tachyus-ryan:
When I attempt to port-forward Argo running in GKE, I get a certificate error. I tried the [suggestion from the first comment here](https://stackoverflow.com/a/36601223), but I still get the certificate error.

```
C:\Users\ryanr\Code\hera-workflows\examples [main ≡ +0 ~1 -0 !]> python .\diamond.py
Traceback (most recent call last):
  File "C:\Users\ryanr\.virtualenvs\hera-workflows-FA8T1hb6\lib\site-packages\urllib3\connectionpool.py", line 706, in urlopen      
    chunked=chunked,
  File "C:\Users\ryanr\.virtualenvs\hera-workflows-FA8T1hb6\lib\site-packages\urllib3\connectionpool.py", line 382, in _make_request
    self._validate_conn(conn)
  File "C:\Users\ryanr\.virtualenvs\hera-workflows-FA8T1hb6\lib\site-packages\urllib3\connectionpool.py", line 1010, in _validate_conn
    conn.connect()
  File "C:\Users\ryanr\.virtualenvs\hera-workflows-FA8T1hb6\lib\site-packages\urllib3\connection.py", line 426, in connect
    tls_in_tls=tls_in_tls,
  File "C:\Users\ryanr\.virtualenvs\hera-workflows-FA8T1hb6\lib\site-packages\urllib3\util\ssl_.py", line 450, in ssl_wrap_socket
    sock, context, tls_in_tls, server_hostname=server_hostname
  File "C:\Users\ryanr\.virtualenvs\hera-workflows-FA8T1hb6\lib\site-packages\urllib3\util\ssl_.py", line 493, in _ssl_wrap_socket_impl
    return ssl_context.wrap_socket(sock, server_hostname=server_hostname)
  File "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.7_3.7.2544.0_x64__qbz5n2kfra8p0\lib\ssl.py", line 423, in wrap_socket
    session=session
  File "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.7_3.7.2544.0_x64__qbz5n2kfra8p0\lib\ssl.py", line 870, in _create
    self.do_handshake()
  File "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.7_3.7.2544.0_x64__qbz5n2kfra8p0\lib\ssl.py", line 1139, in do_handshake
[...]

PR #48, opened 2022-01-10
Refactor `Config` to allow host and SSL specifications

flaviuvadan:
There are multiple use cases in which users would get value out of choosing their own host/scheme combo, along with SSL verification behavior. This PR refactors the `Config` object:
- removes `domain` with `host` full specifications, so users can have complete control over this
- adds scheme/domain/port inference conditionally
- adds a `verify_ssl` flag
- makes host spec optional so that it's clear users can rely on the env variables exposed by Argo in their own K8S cluster

Fixes: #40, merging into branch for #41

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,9 @@ The general format is:
 
 - location of all files from `v1` up one folder to `hera`. Now everything will take the import form
   of `from hera.module import Object` rather than `from hera.v1.module import Object`
+- interface of services to take a full host rather than a single domain and put in effort to compute the final host.
+  This will offer more freedom to users to select their own host scheme, for example. A flag for SSL verification was
+  also introduced
 
 # 0.4.0 - DATE (15/12/2021)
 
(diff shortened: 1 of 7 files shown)


flaviuvadan:
@bartcode @dixon1e let me know what you think about this one!

dixon1e:
This looks great, looking to give it a try right away so I understand it better.

bartcode:
Looks good to me! Haven't managed to try it yet, though.

flaviuvadan:
Alright, let's merge it with this structure. I hope this gives back the freedom to cover all the cases of HTTP vs. HTTPS, and whether to perform SSL/TLS verification. However, you should _definitely_ submit an issue if you notice any odd behavior!

flaviuvadan:
> Looks good to me! Haven't managed to try it yet, though.

I will leave this here until tomorrow afternoon EST in case you do have the chance to check out this branch and try out Hera against your setup!

bartcode:
Can't do it before that time, but that's okay. Happy to see you're actively working on solving issues like these!

flaviuvadan:
> Can't do it before that time, but that's okay. Happy to see you're actively working on solving issues like these!

Alright, and thank you 🙂  again, definitely submit an issue if it does not work in your setup! My hunch is that it'll be ok given the discussions we've had on #16 and #24 🤞🏻

PR #50, opened 2022-01-11
added security_context in Workflow constructor

bayerasif:
Hi!
I added the ability to provide a security context object (V1PodSecurityContext) to a Workflow constructor.
Hopes it meets your standards and you can find it useful as well.


CHANGELOG.md
```diff
@@ -28,6 +28,12 @@ The general format is:
 - C from D
 
 ```
+# 1.0.1 - DATE (11/01/2022)
```
harelwa, line 31: Hi @bayerasif 

`1.0.1` was already published, so you can rebase main, and change that to `1.0.2`

flaviuvadan:
Thank you for opening the PR @bayerasif! 🙂  I am confident the introduction of security contexts will be valuable for many users. 

One of the things Hera aims to do is abstract away uses of Argo Workflows contextual elements, such as objects from the underlying Argo SDK. Therefore, Hera objects never expose constructs such as `V1ObjectMeta`, `V1alpha1Artifact`, etc. Instead, Hera exposes specific fields of these objects and sets them for users either directly or indirectly via some computation over that field. For instance, the string necessary for identifying the task output to take as input via an artifact used by `V1alpha1Artifact` is assembled for the user to reduce the complexity of Argo Workflows usage. In light of this, how do you think we could expose elements of `V1PodSecurityContext` so that the context can be set on `Workflow`/`CronWorkflow` but not have the object itself exposed? I see a test for an FS group on this PR. Perhaps we could expose just that field (`fs_grup: Optional[int] = None`), for now, and create the context inside the `Workflow`? Also, with something like security contexts we have to be mindful to not introduce biases in Hera that would prevent a subclass of users from using it. For instance, [the Argo docs](https://argoproj.github.io/argo-workflows/workflow-pod-security-context/) mention that the Docker executors needs `privileged: true`, which means the workflow pods need `root` access. What do you think? 🙂

CHANGELOG.md
flaviuvadan: Since this PR aims to add functionality in a backwards compatible manner (new feature), and Hera follows semantic versioning, I suggest we bump to `1.1.0` rather than `1.0.2`. Using `1.0.2` would indicate a patch/bug fix, which is not the aim of the PR. https://semver.org/ is a great resource for more information! 🙂

bayerasif:
@flaviuvadan Thanks! Got it! sounds reasonable :smile: I figured you might want to abstract the Argo SDK but still wanted to expose the entire security context in one shot. For the meanwhile I exposed only the fs_group parameter for my use and committed it now, hope you might find it useful :)

flaviuvadan:
@bayerasif [another PR](https://github.com/argoproj-labs/hera-workflows/pull/85) came through that implements the security context feature in Hera. That one exposes all of the security fields in a fairly concise way! Any chance you can take a look at it? It provides the `fs_group` feature you'd like to see, and the other ones too 🙂 If it solve the problem you are faced with, any chance we can close this PR?

bayerasif:
@flaviuvadan Sure! looks great!

PR #55, opened 2022-01-13
Add timezone to CronWorkflowSpec

bobh66:
This PR adds support for the timezone attribute to the CronWorkflow object.

Please let me know if additional changes are needed.


src/hera/cron_workflow.py
```diff
@@ -64,13 +69,11 @@ def __init__(
             templates=[self.template], entrypoint=self.name, service_account_name=self.service_account_name
         )
 
-        self.cron_spec = V1alpha1CronWorkflowSpec(schedule=self.schedule, workflow_spec=self.spec)
+        self.cron_spec = V1alpha1CronWorkflowSpec(schedule=self.schedule, timezone=timezone, workflow_spec=self.spec)
```
flaviuvadan, line 72: Minor suggestion: 
```suggestion
        self.cron_spec = V1alpha1CronWorkflowSpec(schedule=self.schedule, timezone=self.timezone, workflow_spec=self.spec)
```

flaviuvadan (commented):
Any chance you had the opportunity to functionally test this against an Argo cluster? 🙂  Also, need to:
- increment `VERSION`
- add a `CHANGELOG` entry

flaviuvadan:
I've been thinking about the way we have validators in `Resource` and `Task` and I am wondering whether we should implement one for a `Workflow`/`CronWorkflow` as well. One simple validation is checking that the specified `timezone` field is valid on a `CronWorkflow`. This can be done by introducing `pytz`. I took a look through that package and doing a `pip show pytz` lists no additional dependencies (balance between implementing vs. bringing in a library). Perhaps on a different PR/issue, what are your thoughts on adding `pytz` so we can check whether `timezone in pytz.all_timezones`? Would that offer users an additional level of safety?

src/hera/cron_workflow.py
bobh66: I'll update the PR

bobh66:
I was able to test these changes against an Argo server and verified that the timezone attribute shows up in the CronWorkflow object.  I didn't validate the timezone _functionality_ on the server (yet) but that would be a bug against the server if it's not working.

I'm doing the timezone validation in my application API (using a pydantic validator with pytz) but we could put it here too if you like.  I agree it's a balancing act and it's not always clear where the validation should happen.

I'll update the PR with the requested changes.

Thanks!

CHANGELOG.md
```diff
@@ -29,6 +29,12 @@ The general format is:
 
 ```
 
+# 1.1.0 - DATE (13/01/2022)
+
+### Added
+
+- add support for the timezone attribute of CronWorkflow and validate the specified timezone
```
flaviuvadan, line 36: What do you think about including a small note about `pytz` so we know when it was introduced? 

```suggestion
- add support for the timezone attribute of CronWorkflow and validate the specified timezone
- introduce `pytz` dependency for timezone validation
```

Pipfile
```diff
@@ -5,6 +5,7 @@ name = "pypi"
 
 [packages]
 hera = { editable = true, path = "." }
+pytz = "*"
```
flaviuvadan, line 8: There's no need to include it in the `Pipfile` as it is automatically brought in, and locked, through the `setup.py` specification 🙂  
```suggestion
```

tests/test_cron_workflow.py
```diff
@@ -134,5 +136,16 @@ def test_cwf_overwrites_head_and_tail(cw, no_op):
...
+
+
+def test_cwf_valid_timezone_set(cws):
+    cw = CronWorkflow('cw', "* * * * *", timezone="UTC", service=cws)
+    assert cw.timezone == "UTC"
+
+
+def test_cwf_invalid_timezone_set(cws):
+    with pytest.raises(ValueError):
+        cw = CronWorkflow('cw', "* * * * *", timezone="foobar", service=cws)
```
flaviuvadan, line 151: Perhaps include a check for the error message to ensure we get a failing test in case it changes? 
```suggestion
def test_cwf_invalid_timezone_set(cws):
    with pytest.raises(ValueError) as e:
        cw = CronWorkflow('cw', "* * * * *", timezone="foobar", service=cws)
    assert str(e) == 'foobar is not a valid timezone'
```

src/hera/cron_workflow.py
```diff
@@ -1,8 +1,10 @@
 """The implementation of a Hera cron workflow for Argo-based cron workflows"""
-from datetime import datetime, timezone
+from datetime import datetime
+from datetime import timezone as tz
 from typing import Optional
 from uuid import uuid4
 
+import pytz
```
flaviuvadan, line 7: I suggest we run a `make format` prior to merging so we get consistent formatting across the introduced changes! 🙂

flaviuvadan:
Amazing stuff, thanks again for contributing! Left some minor suggestions that mostly have to do with cleanup and a test. We'll be ready to merge right after those are addressed 🚀

flaviuvadan (approved):

flaviuvadan:
For the `Library stubs not installed for "pytz"` build error, we can add: 
```
[mypy-pytz.*]
ignore_missing_imports = True
```
to the `mypy.ini` file.

flaviuvadan:
@bobh66 circling back to this 🙂  any chance we can get this shipped soon? After we adjust the `mypy` config, along with the `VERSION` and `CHANGELOG`!

bobh66:
@flaviuvadan the tests fail when pytz is in setup.cfg but not in Pipfile.

Should I run pipenv install pytz locally and add Pipfile and Pipfile.lock into the PR?

flaviuvadan:
> @flaviuvadan the tests fail when pytz is in setup.cfg but not in Pipfile.
> 
> Should I run pipenv install pytz locally and add Pipfile and Pipfile.lock into the PR?

Right! The build uses a virtual environment that has the packages listed in `Pipfile.lock` so we need to update the lockfile. The easiest way to do that is:
```shell
> deactivate // leave the current venv, if you are told you already have a shell, use `exit`
> pipenv --rm
> rm Pipfile.lock
> pipenv install --dev --pre
```
This will relock the project and bring in the specified dependencies from `setup.py` 🙂  Happy to explain more if needed! I think this is the last step we need.

bobh66:
Thanks @flaviuvadan  - do I include Pipefile in the PR or just Pipfile.lock?

flaviuvadan:
> Thanks @flaviuvadan - do I include Pipefile in the PR or just Pipfile.lock?

Just the `Pipfile.lock` changes!

flaviuvadan (approved):

flaviuvadan:
Thank you for adding this @bobh66! 🚀  I am super thankful for all your patience in going back and forth on the design, review, and discussion!

PR #57, opened 2022-01-13
allow task to run as daemon

GuyPozner:
Add the possibility to run task as a daemon.

I've fixed the user name issues with the commit so I can conform with the CLA.


flaviuvadan:
Thank you for submitting this! 🙂  Just a reminder to update the `VERSION` and add a `CHANGELOG` entry but it depends on the order of merges you'd like to perform!

VERSION
```diff
@@ -1 +1 @@
-1.0.1
+1.0.2
```
flaviuvadan, line 1: Thank you for adjusting the version! 🙂 One minor thing - Hera uses [semantic versioning](https://github.com/argoproj-labs/hera-workflows/blob/main/CONTRIBUTING.md#contributing-to-hera). This indicates a PATCH, which implies a bug fix, but this is a feature! So, I suggest we bump the MINOR version, to something like `1.1.0` perhaps.

src/hera/task.py
```diff
@@ -90,6 +92,7 @@ def __init__(
         input_artifacts: Optional[List[InputArtifact]] = None,
         output_artifacts: Optional[List[OutputArtifact]] = None,
         image: str = 'python:3.7',
+        daemon: bool = False,
```
flaviuvadan, line 95: I was looking through the `V1alpha1Template` object and it seems the default value is `None`. What do you think about making this: 
```suggestion
        daemon: Optional[bool] = None,
```
This way Hera is (1) aligned with the SDK default value and (2) avoid potentially unexpected behavior via setting the value to `False` (I don't know what this could be, tbh, but it has happened before in my XP that setting a value actually changes behavior rather than leaving stuff to default 🤷🏻‍♂️).

GuyPozner: agree

bchalk101:
Hi @flaviuvadan,
Is there something blocking the MR being merged? We are currently blocked by it and hoping to get it in :)
Thnx

flaviuvadan:
> Hi @flaviuvadan, Is there something blocking this PR from being merged? We are currently blocked by it and hoping to get it in :) Thnx

Hi @bchalk101! Apologies for not circling back to this! Let's ship this once the `VERSION` and `CHANGELOG` conflict is resolved 🙂 I will keep an eye out for my GH notifications so I get to it ASAP.

flaviuvadan (approved):
Again, thank you for the contribution! 🚀

PR #59, opened 2022-01-17
make value in tolerations optionals

bchalk101:
As per the Kubernetes API, make the value in a Toleration optional. 
https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#taint
https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/


src/hera/toleration.py
```diff
@@ -25,7 +27,7 @@ class Toleration(BaseModel):
     key: str
     effect: str
     operator: str
-    value: str
+    value: Optional[str] = ""
```
flaviuvadan, line 30: First, thank you for submitting this PR! Now that you mentioned it as a K8S characteristic, I recalled it's possible to have taints on nodes without a requirement for a value 🙂  What do you think about this:
```suggestion
    value: Optional[str] = None
```
This is more in line with the default values of the `argo-workflows` SDK, which uses `None` pretty much everywhere.

bchalk101: updated with None as defaul value

flaviuvadan (commented):
What do you think about adding a tiny, tiny test to verify that a `None` is indeed passed to the respective toleration on a `Task`? 🙂 I think that would increase our confidence in a merge, which we will 100% do soon!

bchalk101:
I've added a test for it.
Although in general, I'm not sure the value of a test in this situation, as I would consider this more a refactor, than a feature addition. Given though there was no testing around a Toleration inside a Task I did add a test anyways.

flaviuvadan (approved):

flaviuvadan:
Thank you for the patch @bchalk101 🚀

PR #63, opened 2022-01-19
fixed trailing quotation marks

### LINKED ISSUE #62: EnvSpec bug?
Not sure if intended but setting environment variables via EnvSpec leaves unwanted quotation marks in the string when resolved in python script as shown below.

Steps to reproduce: Execute the below task and see that it prints false, while it should be true

`def test_env_spec():

    ressource = Resources(
            min_cpu=1,
            max_cpu=2,
            min_mem="2Gi",
            max_mem="4Gi"
    )
    envs = [
        EnvSpec(
                name="ENV_STRING",
                value='test'
        ),
    ]
    task = Task(
            'test',
            lambda: _,
            None,
            image='python:3.7',
            command=[
                "python",
                "-c",
                "import os; print(os.environ['ENV_STRING']);print(os.environ['ENV_STRING']=='test')",
            ],
            resources=ressource,
            env_specs=envs
    )
    return task

KukumavMozolo:
CLOSES #62


flaviuvadan (changes requested):
First, thank you so much for taking the time to add this! 🚀  I think it exposed something very important in the design of Hera, and we should discuss it here: https://github.com/argoproj-labs/hera-workflows/issues/62#issuecomment-1017492159

flaviuvadan (commented):
Just a note to increment the version of the project 🙂 in this case I'd go with a `PATCH` since this fixes a specific use case for environment variables. Also, gentle reminder to include a `CHANGELOG` entry for this change. Once that's done we can :shipit:

KukumavMozolo:
> Just a note to increment the version of the project slightly_smiling_face in this case I'd go with a `PATCH` since this fixes a specific use case for environment variables. Also, gentle reminder to include a `CHANGELOG` entry for this change. Once that's done we can :shipit:

added suggested changes :)

flaviuvadan (approved):
Thank you for the valuable discussion on #62, your patience, and contribution @KukumavMozolo! 🙂

Issue #36, opened 2022-01-08
Make `func: Callable` an optional parameter on `Task`


flaviuvadan:
If users want to submit a container and execute a command in it, they currently need to supply a `lambda: _` (no op) placeholder for `func`. This should not be required as it also forces Argo to insert an empty script, rather than just execute the container, along with the provided command.

PR #66, opened 2022-01-21
Optional func

bchalk101:
Addresses #36 

Uses a container template when there is no func. 

Updated container example to not require the func, and tested against local argo instance.


src/hera/task.py
```diff
@@ -220,6 +218,10 @@ def validate(self):
                 self.output_artifacts
             ), 'output artifact names must be unique'
 
+        if self.func:
+            self._validate_func()
+
+    def _validate_func(self):
```
flaviuvadan, line 224: I think this can be a public method 🤔

src/hera/task.py
```diff
@@ -341,7 +342,14 @@ def get_parameters(self) -> List[V1alpha1Parameter]:
             args = set(inspect.getfullargspec(self.func).args).intersection(set(self.input_from.parameters))
             for arg in args:
                 parameters.append(V1alpha1Parameter(name=arg, value=f'{{{{item.{arg}}}}}'))
+        if self.func:
+            parameters += self._get_func_parameters()
+
+        return parameters
 
+    def _get_func_parameters(self):
```
flaviuvadan, line 350: Similarly to above, I think this can be public 🙂

src/hera/task.py
```diff
@@ -513,15 +522,34 @@ def get_script_def(self) -> V1alpha1ScriptTemplate:
...
             env=self.env,
             resources=self.argo_resources,
+        )
+
+    def get_container(self) -> V1Container:
+        """Assembles and returns the container for the task to run in.
+
+        Returns
+        -------
+        V1alpha1ScriptTemplate
```
flaviuvadan, line 543:
```suggestion
        V1Container
```

tests/test_task.py
```diff
@@ -341,3 +341,17 @@ def test_task_template_has_correct_labels(op):
...
+
+    assert not tt.container
+
+def test_task_should_create_task_with_container_template():
+    t = Task('t', command=["cowsay"])
+    tt = t.get_task_template()
+
+    assert tt.container.image == "python:3.7"
+    assert tt.container.command[0] == "cowsay"
+    assert tt.container.resources.requests["memory"] == '4Gi'
```
flaviuvadan, line 357:
```suggestion
    assert tt.container.resources.requests["memory"] == '4Gi'

```

flaviuvadan (commented):
I almost cannot believe someone contributed this to Hera! It's amazing 🚀 Thank you for the invested effort! 

I left some brief comments that are not critical to the functional aspect of Hera but, given the number of necessary changes, I am going to clone this and test it out myself against an Argo cluster, then come back with either comments or an approval. Please mention me if I do not come back to this in a timely fashion that allows your use case to take advantage of containers 🙂

src/hera/task.py
bchalk101: Addressing this and the one above, wondering what the justification would be to make this public? Outside of the task class what benefit would there be to access these?

flaviuvadan: The main reason is exposing parts of Hera's internals in order to offer some flexibility to users who build their own `Task` on top of the Hera one. For instance, if there are users who want to inject something in their template script, in addition to what Hera injects already, they can access the public method to obtain the pre-built one, and modify that one, rather than write the script assemble method themselves. This kind of applies generally - users can use any of the public functions to use the "nice" aspect of Hera, which sets up a bunch of stuff already, so users save some work and only have to be concerned with "what do _I_ want to inject further into Argo object definitions that helps my problem".

bchalk101: My concern with this would be that it is introducing a certain level of brittleness to the code base, for example, if at a certain point we needed to change the method signature, we would be introducing a breaking change.
Perhaps, for the getter `_get_func_parameters` I hear, but for the validate method, would it not be `validate` is the public method, and than internal calls are private?

flaviuvadan: I definitely see where you're coming from. Perhaps the balance between a public `get_func_parameters` and a private `validate` that you suggested works best for now. So, let's go with your suggestion! 🙂

src/hera/task.py
```diff
@@ -341,7 +342,14 @@ def get_parameters(self) -> List[V1alpha1Parameter]:
             args = set(inspect.getfullargspec(self.func).args).intersection(set(self.input_from.parameters))
             for arg in args:
                 parameters.append(V1alpha1Parameter(name=arg, value=f'{{{{item.{arg}}}}}'))
+        if self.func:
+            parameters += self.get_func_parameters()
+
+        return parameters
 
+    def get_func_parameters(self) -> List[V1alpha1Parameter]:
```
flaviuvadan, line 350: Last thing we need is docstrings for `get_func_parameters` and then we can ship! :shipit:

flaviuvadan (approved):

flaviuvadan:
Thank you again @bchalk101! 🚀

PR #68, opened 2022-01-24
add ability to run task as privileged

bchalk101:
- Adds the ability to run the task container as privileged


bchalk101:
@flaviuvadan would u be able to take a look at this one - currently blocking our usage of hera.
Thnx :)

tests/test_task.py
```diff
@@ -341,3 +341,9 @@ def test_task_template_has_correct_labels(op):
     tt = t.get_task_template()
     expected_labels = {'foo': 'bar'}
     assert tt.metadata.labels == expected_labels
+
+
+def test_task_set_as_privaliged(no_op):
```
flaviuvadan, line 346: My meaningless feedback: 
```suggestion
def test_task_set_as_privileged(no_op):
```

🙂

src/hera/task.py
```diff
@@ -522,6 +527,7 @@ def get_script_def(self) -> V1alpha1ScriptTemplate:
             resources=self.argo_resources,
             volume_mounts=self.get_volume_mounts(),
             working_dir=self.working_dir,
+            security_context=V1SecurityContext(privileged=self.privileged),
```
flaviuvadan, line 530: What do you think about isolating this into a class method? I think there is some interest in extending the usage of security contexts so that FS groups can also be mentioned, for instance, and it might be nice to have that started here 🙂 It's also slightly more consistent with the other uses, such as `self.get_command()` and `self.get_volume_mounts()`. I'd understand if you do not want to over-engineer at first!

flaviuvadan (commented):
I think security contexts can be set at the workflow level as well. I like that this offers granularity (context set at the task level) but I imagine the use case for `privileged=True` has to do with the Docker executor. Since that is now deprecated, do you imagine we'll remove this in a future breaking change release of Hera? Just bouncing some thoughts 🙂

bchalk101:
> Since that is now deprecated, do you imagine we'll remove this in a future breaking change release of Hera? 
This is very true, I opened this pr and it came up in conversation straight away with someone else. I'm torn here between closing this issue as it does have to do with a deprecated feature vs hera representing the current state of the python api, and what is currently available to users via the templates.

flaviuvadan:
> > Since that is now deprecated, do you imagine we'll remove this in a future breaking change release of Hera?
> 
> This is very true, I opened this pr and it came up in conversation straight away with someone else. I'm torn here between closing this issue as it does have to do with a deprecated feature vs hera representing the current state of the python api, and what is currently available to users via the templates.

At Dyno we switched from the `docker` executor to `pns` for the [workflow controller](https://argoproj.github.io/argo-workflows/workflow-controller-configmap.yaml) without much hassle/follow-up intervention. 

> deprecated feature vs hera representing the current state of the python api

Yea, this is quite challenging. Do you have Argo deployed to a cluster? Would switching the executor to, perhaps, `pns` help here? We use GCP at Dyno and switching to `pns` actually solved a few snags we've encountered with Argo fetching pod logs post-completion and some network errors 🤔

bchalk101:
definitely in our backlog to upgrade the executor - im going to close this issue and will solve the need for this on our side, seeing as docker is going to be deprecated anyway.

PR #71, opened 2022-01-25
Add default name/namespace handling to CronWorkflow create/suspend/re…

bobh66:
Refactored the way the CronWorkflow name and namespace attributes are handled by the create/suspend/resume methods so that the name and namespace can be specified at object creation and the methods can be invoked with no arguments.

It is still possible to call the methods with arguments as before and the same functionality will continue to work.


src/hera/cron_workflow.py
```diff
@@ -52,6 +54,7 @@ def __init__(
         parallelism: int = 50,
         service_account_name: Optional[str] = None,
         labels: Optional[Dict[str, str]] = None,
+        namespace: Optional[str] = 'default',
```
flaviuvadan, line 57: I think `Optional` represents a `Union[{chosen_type}, None]` so perhaps this can be:
```suggestion
        namespace: str = 'default',
```

src/hera/cron_workflow.py
```diff
@@ -41,6 +41,8 @@ class CronWorkflow:
         The name of the service account to use in all workflow tasks.
     labels: Optional[Dict[str, str]] = None
         A Dict of labels to attach to the CronWorkflow object metadata
+    namespace: Optional[str]
```
flaviuvadan, line 44: If we end up changing the type to a primitive one exclusively, then this can also work:
```suggestion
    namespace: str = 'default'
```

flaviuvadan: Never mind, actually. I see the usages on `create/suspend/resume` 🙂 
```suggestion
    namespace: Optional[str] = None
```

src/hera/cron_workflow.py
```diff
@@ -41,6 +41,8 @@ class CronWorkflow:
...
+        The namespace to use as a default when calling create/suspend/resume
```
flaviuvadan, line 45: Should make a note that this falls back on the usage of `default`.

src/hera/cron_workflow.py
```diff
@@ -175,14 +179,24 @@ def add_tail(self, t: Task, append: bool = True) -> None:
         free_tasks = set(task_name_to_task.keys()).difference(dependencies)
         t.argo_task.dependencies = list(free_tasks)
 
-    def create(self, namespace: str = 'default') -> None:
+    def create(self, namespace: str = None) -> None:
```
flaviuvadan, line 182:
```suggestion
    def create(self, namespace: Optional[str] = None) -> None:
```

src/hera/cron_workflow.py
```diff
@@ -175,14 +179,24 @@ def add_tail(self, t: Task, append: bool = True) -> None:
...
         """Creates the cron workflow in the server"""
+        if namespace is None:
+            namespace = self.namespace
         self.service.create(self.workflow, namespace)
 
-    def suspend(self, name: str, namespace: str = 'default'):
+    def suspend(self, name: str = None, namespace: str = None):
```
flaviuvadan, line 188:
```suggestion
    def suspend(self, name: Optional[str] = None, namespace: Optional[str] = None):
```

src/hera/cron_workflow.py
```diff
@@ -175,14 +179,24 @@ def add_tail(self, t: Task, append: bool = True) -> None:
...
         """Suspends the cron workflow"""
+        if name is None:
+            name = self.name
+        if namespace is None:
+            namespace = self.namespace
         self.service.suspend(name, namespace)
 
-    def resume(self, name: str, namespace: str = 'default'):
+    def resume(self, name: str = None, namespace: str = None):
```
flaviuvadan, line 196:
```suggestion
    def resume(self, name: Optional[str] = None, namespace: Optional[str] = None):
```

flaviuvadan (commented):
Thanks again @bobh66 🚀 do you plan to upgrade the `Workflow` object in the same PR?

flaviuvadan:
> @flaviuvadan I wasn't sure if you wanted it in the same PR or in a separate one. I can certainly add it to this PR if that's OK.

Same PR is definitely fine in this case 🙂

flaviuvadan (approved):
Awesome work @bobh66 🚀

PR #76, opened 2022-01-30
WIP: Add option to override unique name generation for Workflow and CronWorkflow

bobh66:
@flaviuvadan Let me know what you think of this idea.   It preserves the existing unique name functionality and provides the ability to keep the name specified by the caller unchanged if so desired.  If you agree I'll add the VERSION and CHANGELOG entries once I know where in the flight pattern it will land.

Thanks

### DIFF
--- a/src/hera/cron_workflow.py
+++ b/src/hera/cron_workflow.py
@@ -43,6 +43,10 @@ class CronWorkflow:
         A Dict of labels to attach to the CronWorkflow object metadata
     namespace: Optional[str] = 'default'
         The namespace to use by default when calling create/suspend/resume.  Defaults to 'default'.
+    use_unique_name: Optional[bool] = True
+        Flag to indicate whether the name should be made unique by appending a uuid4() substring.
+        Defaults to True to preserve existing functionality.
+        Note that the V1alpha1Template name will continue to use the unique name.
     """
 
     def __init__(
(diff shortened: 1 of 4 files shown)


src/hera/cron_workflow.py
```diff
@@ -55,11 +59,14 @@ def __init__(
...
+        use_unique_name: Optional[bool] = True,
     ):
         if timezone and timezone not in pytz.all_timezones:
             raise ValueError(f'{timezone} is not a valid timezone')
-
-        self.name = f'{name.replace("_", "-")}-{str(uuid4()).split("-")[0]}'
+        self.name = name.replace("_", "-")
+        self.unique_name = f'{self.name}-{str(uuid4()).split("-")[0]}'
+        if use_unique_name:
+            self.name = self.unique_name
```
flaviuvadan, line 69: It strikes me that 
```python
self.name = name.replace("_",  "-")    # RFC1123
if use_unique_name:
    self.name = "-".join([self.name, str(uuid4()).split("-")[0]])
```
is a _bit_ clearer. Thoughts? 🙂

flaviuvadan:
Hi @bobh66! First, I apologize for the delay in reviewing this. Second, thank you for the PR! 🚀 I like the addition and understand the use case. However, one thing I am thinking about is.. should Hera just use whatever name is passed in? Perhaps allow the user to pass in a unique name based on the uniqueness protocol they choose? I am not sure whether this is worth a breaking change... but thought of asking. Let me know what you think! 

I left some minor feedback on the order of statements but otherwise LGTM!

src/hera/cron_workflow.py
bobh66: Hi @flaviuvadan  - It's clearer but it doesn't preserve the unique_name for use in the template.  I found that when I overrode the name, metadata.name and template.name I got an error when trying to create the workflow.  I had to leave the template.name with the generated unique name to avoid the error.  I can look into that again and see what the specific error was.

bobh66:
Hi @flaviuvadan  - I definitely prefer using the name provided by the user and letting them handle the uniqueness requirement.  Maybe that change can go in 2.0 or whenever the Argo API change lands?

flaviuvadan:
> prefer using the name provided by the user and letting them handle the uniqueness requirement. Maybe that change can go in 2.0 or whenever the Argo API change lands?

Let's make this change when we release Hera 3.0.0, perhaps?

bobh66:
@flaviuvadan That sounds fine - I'll get this change finished up and we can revisit when 3.0.0 is in the works.  Thanks

flaviuvadan:
Let's discuss this when we publish 3.0.0!

Issue #51, opened 2022-01-11
problems with v1.0.0 package published to pypi


harelwa:
Hi,

Something is wrong with the wheels / pypi package published.

Here is the way to inspect / reproduce:

```bash
> docker run -it --rm continuumio/miniconda3
> conda create -n py37-env python=3.7 pip -y
[...]
> conda activate py37-env
(py37-env) > pip --version
pip 21.2.2 from /opt/conda/envs/py37-env/lib/python3.7/site-packages/pip (python 3.7)
(py37-env) > pip install hera-workflows==1.0.0
[...]
(py37-env) > python -c "import hera"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named 'hera'
(py37-env) > ls /opt/conda/envs/py37-env/lib/python3.7/site-packages/ | grep hera
hera_workflows-1.0.0.dist-info
# here the actual package is missing, e.g. with `pydantic` you'll get 
(py37-env)  > ls /opt/conda/envs/py37-env/lib/python3.7/site-packages/ | grep pydantic
pydantic
pydantic-1.9.0.dist-info
```

The above issue is also the cause of the `conda-forge` feedstock CI failing - 

https://github.com/conda-forge/hera-workflows-feedstock/pull/2

Many thanks,
Harel

flaviuvadan:
Hey @harelwa! Thank you for submitting this. I will take a look soon!

flaviuvadan:
Please see: https://github.com/argoproj-labs/hera-workflows/pull/52

flaviuvadan:
@harelwa I just installed 1.0.1 in a virtual environment through `pip install hera-workflows==1.0.1`. Unfortunately 1.0.0 will remain broken on PyPi 🙁  What's the best way to help conda-forge in a case like this? 

I must apologize for not catching this earlier. It's evident that Hera needs a step that publishes it to Test PyPi prior to the official PyPi index.

harelwa:
Hi @flaviuvadan 

Thank you very much for the [so]fast reply and fix!

as for `conda-forge` -- 

no worries. it did not publish due to the tests failure. I will update the recipe to look at `v1.0.1`.

please note that this err was caught by `conda-forge` ci, as `conda build` runs tests *of the built package*. you can see more here - <https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#test-section>

[...]

PR #43, opened 2022-01-09
Revert "Revert "Upgrade underlying Argo SDK""

flaviuvadan:
This is the actual implementation of the migration of the underlying Argo SDK. The initial release process failed, so I need to look into how to make twine publish the package.

Reverts argoproj-labs/hera-workflows#42

PR #78, opened 2022-02-02
Migrate Hera to Argo Workflows 6.3.0rc2

flaviuvadan:
This PR migrates Hera's underlying SDK to Argo Workflows 6.3.0rc2. The new Argo Workflows Python SDK was generated with OpenAPI v5+, which generates objects differently than v4, which was used to generate the previous SDK.

One thing to note is the introduction of a custom object `_Item`. There was a very lengthy discussion [here](https://github.com/argoproj/argo-workflows/issues/7374) about how the primitive type used by OpenAPI does not seem to be serialized/set correctly upon submission. This happens with the `with_items` field as well. To get around this problem, this PR introduces `_Item`, which is a custom object that holds the `value` field necessary for correct serialization of primitive types. Please see the docstring of `_Item` for further explanations.

I think your input would be amazing here @bobh66 and @bchalk101! 

Background of this:
- https://github.com/argoproj-labs/hera-workflows/issues/7
- https://github.com/argoproj-labs/hera-workflows/pull/46
- https://github.com/argoproj-labs/hera-workflows/pull/45
- https://github.com/argoproj-labs/hera-workflows/pull/43
- https://github.com/argoproj-labs/hera-workflows/pull/42
- https://github.com/argoproj-labs/hera-workflows/issues/51


src/hera/config.py
```diff
@@ -3,7 +3,7 @@
 from typing import Optional
 
 import urllib3
-from argo.workflows.client import Configuration as ArgoConfig
+from argo_workflows.api_client import Configuration as ArgoConfig
```
flaviuvadan, line 6: Make this `_ArgoConfig`.

src/hera/toleration.py
```diff
@@ -20,14 +20,12 @@ class Toleration(BaseModel):
...
         The value to scan for matching against a taint.
-
-    return V1Toleration(key='nvidia.com/gpu', effect='NoSchedule', operator='Equal', value='present')
     """
 
     key: str
     effect: str
     operator: str
-    value: Optional[str] = None
+    value: Optional[str] = ""
```
bchalk101, line 28: Is there a reason for using a string default rather than a None?

src/hera/client.py
```diff
@@ -1,5 +1,5 @@
 """Holds client configurations for communicating with Argo APIs"""
-from argo.workflows.client import ApiClient as ArgoApiClient
+from argo_workflows.api_client import ApiClient as _ArgoApiClient
```
bchalk101, line 2: It appears slightly inconsistent I feel. Sometimes the imports from argo are used as is ie `from argo_workflows.models import IoArgoprojWorkflowV1alpha1Artifact` and other times as above with an `as`. Is there a reason for this? Also why now make these "private" imports - I believe it takes away somewhat from the readability.

src/hera/cron_workflow.py
```diff
@@ -65,29 +65,32 @@ def __init__(
         self.service = service
         self.parallelism = parallelism
         self.service_account_name = service_account_name
-        self.labels = labels
-        self.namespace = namespace
+        self.labels = labels or {}
```
bchalk101, line 68: Doing defaults like this could have unintended consequences, as was the case with the command in a `Task` being impossible to override and always being forced to be `python`. I feel either we use defaults via the `labels ={}` or not at all, unless absolutely required.

src/hera/toleration.py
flaviuvadan: Default changes from `None` to `""` are motivated by the mechanism of initialization on the new objects. The generated `Toleration` object expects a `str` type for the `value` field. For `Toleration`, `value` is an _optional_ field but expects a specific field value type - optional `value` but the actual value must be a `str`. Perhaps it might be better to check for `None` and not set the field at all?

src/hera/client.py
flaviuvadan: Not all of the generated objects have the `IoArgoprojWorkflow` prefix and some objects removed versioning from their name (`V1`). That caused some clashes with the Hera objects, as is the case for `Toleration` (used to be `V1Toleration` and now migrated to `Toleration`, motivating the `import ... as ArgoToleration`). As for the private imports, I can definitely see how it affects readability and consistency. Now that you pointed it out, we should probably stick to a consistent pattern - I will remove the private imports. As for the `as` statement, I think that is required, for now, so that we do not break Hera's simple interface. For instance, I would stand against a rename of `Toleration` to `HeraToleration`. I'd rather have Hera use alias imports so the client can avoid the need to do that with `import ... HeraToleration as Toleration`.

src/hera/cron_workflow.py
flaviuvadan: > either we use defaults via the labels ={} or not at all, unless absolutely required.

I am starting to share this sentiment. Perhaps we do it the same way we set the `service_account_name` rather than set default values? The new SDK objects use a bunch of `setattr` dynamically, so we can also use that for labels if they are set. I can also definitely see the potential for unintended consequences - perhaps K8S might use these empty values somehow (generally, not labels necessarily) and behave unexpectedly. 

What do you think about resorting to the use of conditional `setattr(...)` the same way we do it with the `service_account_name` but for labels, and basically any other field that moved from `None` to something else? I now think it's better to check for nullity and set accordingly, which also supports more consistency in usage of the SDK.

bchalk101: I think that's probably a good way to go about it, while it is definitely verbose, I think it is more explicit for the end-user. I am though not against using defaults in signatures in certain situations, for example with the python command.

I am intrigued though why the use of `setattr()` vs `self.template.service_account_name = self.service_account_name`. Although, I see that with the new OpenAPI spec, either one is not validated, and there would be no code editor benefit with either. Prob more a personal issue than anything else, I just find string name variables slightly on the edge of Magic Variables.

PR #83, opened 2022-02-08
Add support for multiple volume specifications (volumes, config maps, secrets, existing volumes)

flaviuvadan:
This PR adds support for multiple volume specifications. Users will now have access to specifying multiple volumes on the `Resources` object, which will either dynamically provision multiple volumes for specified volumes, mount multiple existing volumes, mount multiple secrets, and/or mount multiple config maps.

Closes: #33


bchalk101:
Looks good - really like how concise it is now!

PR #84, opened 2022-02-08
add s3 key only artifact

bchalk101:


src/hera/artifact.py
```diff
@@ -61,3 +65,80 @@ def get_spec(self) -> IoArgoprojWorkflowV1alpha1Artifact:
...
+    This artifact is used to represent a bucket object
+
+    Attributes
+    ----------
+    name: str
+        The name of the input artifact.
+    path: str
+        The path where to store the input artifact. Note that this path is isolated from the output artifact path
+        of the previous task artifact.
+    bucket_name: str
```
flaviuvadan, line 82:
```suggestion
    bucket: str
```

bchalk101:
getting this error:
`HTTP response body: {"code":2,"message":"templates.artifact-passing-1fcad5d8.tasks.consumer templates.consumer.inputs.artifacts.test.from not valid in inputs"}`

bchalk101:
yeah, so I removed the `_from` from the input artifact and that got it working. I pushed the working code, however, I feel it's a bit ugly, in the way that the knowledge of having to remove the `_from` lives inside the `Task` rather than inside the `Artifact`.

flaviuvadan:
> knowledge of having to remove the _from lives inside the Task rather than inside the Artifact.

This concern is very reasonable. What do you think about making a `get_input_spec` or `input_from_arg_spec` on the argument objects? Wondering if each artifact/parameter can specify it's own "form" of input.

bchalk101:
I went with `get_input_spec`, moving that to each `Artifact` as required. 

I've tested it against both minio, using the s3 artifact, and tested with the `artifact.py` example, and all is working. :)

PR #85, opened 2022-02-08
 Add security context

GuyPozner:


GuyPozner:
There are two tests that fail in adding security context on the task itself.
Apparently in the [python sdk the security_context is of type PodSecuiryContext](https://github.com/argoproj/argo-workflows/blob/032520f026f909ff110972aa50c4651280a68022/sdks/python/client/argo_workflows/model/io_argoproj_workflow_v1alpha1_template.py#L166) and not [SecurityContext ](https://github.com/argoproj/argo-workflows/blob/032520f026f909ff110972aa50c4651280a68022/sdks/python/client/argo_workflows/model/security_context.py#L95)which does not enable to specify Capabilities. Is it a bug in the argo python sdk?

src/hera/task.py
```diff
@@ -6,6 +6,7 @@
 import textwrap
 from typing import Any, Callable, Dict, List, Optional, Set, Union
 
+from argo_workflows.model.security_context import SecurityContext
```
flaviuvadan, line 9: What do you think about importing this in the block starting on line 16?

src/hera/security_context.py
```diff
@@ -0,0 +1,78 @@
+from typing import List, Optional
+
+from argo_workflows.model.capabilities import Capabilities
+from argo_workflows.model.pod_security_context import PodSecurityContext
+from argo_workflows.model.security_context import SecurityContext
+from pydantic import BaseModel
+
+
+class WorkflowSecurityContext(BaseModel):
```
flaviuvadan, line 9: I think there's some opportunity for restructuring here and I am curious what you think about the following suggestion:
* make a parent `SecurityContext` class that has the 4 permission fields (`run_as_user`, `run_as_group`, `fs_group`, `run_as_non_root`)
* make the `WorkflowSecurityContext` inherit from `SecurityContext` with no additional code added
* make the `TaskSecurityContext` inherit from `SecurityContext` with the extra `additional_capabilities` field

So, the code turns into:
```python
from argo_workflows.model import (
    Capabilities,
    PodSecurityContext,
    SecurityContext,
)

class SecurityContext(BaseModel):
    run_as_user: Optional[int] = None
    run_as_group: Optional[int] = None
    fs_group: Optional[int] = None
    run_as_non_root: Optional[bool] = None
    
    def get_context(...): 
        # current implementation

class WorkflowSecurityContext(SecurityContext):
    pass

class TaskSecurityContext(SecurityContext):
    additional_capabilities: List[str] = None

    def get_context(...): 
        # different implementation  
```

src/hera/security_context.py
```diff
@@ -0,0 +1,78 @@
...
+        security_context = SecurityContext()
+        if self.run_as_user:
+            setattr(security_context, 'run_as_user', self.run_as_user)
+        if self.run_as_group:
+            setattr(security_context, 'run_as_group', self.run_as_group)
+        if self.run_as_non_root is not None:
+            setattr(security_context, 'run_as_non_root', self.run_as_non_root)
+        if capabilities:
+            setattr(security_context, 'capabilities', capabilities)
+        return security_context
```
flaviuvadan, line 78: This [PR](https://github.com/argoproj-labs/hera-workflows/pull/91/files#diff-d0242458c4cfc2ecd6d4cd87b214d860abbb0c8ab7efc0fb03fd5a303869a72cL642) proposes a concise way of setting parameters that does not rely on `setattr`. What do you think about using something similar here? 🙂

flaviuvadan (commented):
This looks great so far! I left some suggestions and I look forward to reading your thoughts on them. Love the tests 🚀

src/hera/security_context.py
GuyPozner: As long as [this](https://github.com/kubernetes/kubernetes/blob/master/pkg/securitycontext/util.go#L163-L196) is true in k8s, we can do this abstraction. Though it might represent a false knowledge about the parent API, because the `SecurityContext` and `PodSecurityContext` in argo are not subclasses of any parent abstract class, what do you think?

src/hera/task.py
GuyPozner: I have put it there because `model` comes before `model_utils` and `models`, but I can change it, whatever you choose.

src/hera/security_context.py
GuyPozner: I like it, though there is one mistake in the pattern `script_kargs = {k: v for k, v in script_kargs.items() if v}`, it won't work for boolean types which take the value of False, so it should be `script_kargs = {k: v for k, v in script_kargs.items() if v is not None}`

src/hera/task.py
GuyPozner: The other comments I have fixed, please update me If you'd like me to change this one as well.

src/hera/task.py
```diff
@@ -647,12 +653,24 @@ def get_script_def(self) -> Optional[IoArgoprojWorkflowV1alpha1ScriptTemplate]:
...
         if self.env:
             setattr(template, 'env', self.env)
         return template
 
+    def get_security_context(self) -> SecurityContext:
+        """Assembles the security context for the task.
+        Returns
+        -------
+        SecurityContext
+            The security settings to apply to the task's container.
```
flaviuvadan, line 670:
```suggestion
        """Assembles the security context for the task.
        
        Returns
        -------
        SecurityContext
            The security settings to apply to the task's container.
```

flaviuvadan:
Looks amazing! Just one last tiny suggestion then we're good to ship 🚢

flaviuvadan:
This PR looks ready to ship! Let's push a MINOR version bump in `VERSION` and add a `CHANGELOG.md` entry then we're set! 🚀

flaviuvadan (approved):

flaviuvadan:
Thank you for the contribution @GuyPozner! 🚀

PR #91, opened 2022-02-09
Allow args without a command

bchalk101:
Breaking change as it removed the default of `python` from the `Task` command.

Potentially could do a check for either `args` or `command` and if neither place `python` as the command


bchalk101:
I will bump the version and changelog, once I know after which change it's going to be merged. Although it prob should be merged into https://github.com/argoproj-labs/hera-workflows/pull/83, otherwise we will have to bump first to 2 for https://github.com/argoproj-labs/hera-workflows/pull/83 and than  to 3 for this MR.

flaviuvadan:
Thank you for the PR! 🙂 I completely appreciate the need to do this - Hera should support running a container with args using the baked in `CMD` of the container image. I wonder, though, if this can be delivered in a non-breaking change manner. Perhaps Hera can offer the flexibility to set any `command` if the client wants to use one, but defaults to using `["python"]` when a `func` is passed with no `command`? This way containers are supported as users don't need to pass `func` and `command` only the optional `args` for container, and if a scientist does specify a `func` (general case) but no `command` then it's likely safe to assume the operation occurs within the context of Python so `["python"]` can be used as the command. Thoughts?

bchalk101:
So I've gone with what I believe to be a non-breaking change.

It will default to `python` if no args are supplied ie `Task('t')`, however, if args are supplied and no cmd has been set, then the command will be left empty ie `Task('t', args=["python"])`, will have no command set.

src/hera/task.py
```diff
@@ -172,9 +172,14 @@ class Task:
...
-        and invoked via `python script_file.py`. This command offers users the opportunity to start up the script in a
-        different way e.g `time python` to time execution, `horovodrun -p X` to use horovod from an image that allows
-        training models on multiple GPUs, etc.
+        and invoked via `python script_file.py`. Also, note that when neither command nor args are set, the command
+        will default to python. This command offers users the opportunity to start up the script in a different way
+        e.g `time python` to time execution, `horovodrun -p X` to use horovod from an image that allows training models
+        on multiple GPUs, etc.
+    args: Optionals[List[str]] = None
+        Optional list of arguments to run in the container. This can be used as an alternative to command, with the
+        advantage of not overriding the set entrypoint of the container. Note that when argo is using the emissary
```
flaviuvadan, line 181: This docstring seems incomplete? Also, perhaps we can add a `Notes` section for the emissary executor note?
```suggestion
        advantage of not overriding the set entrypoint of the container. Note that when argo is using the emissary...
```

src/hera/task.py
```diff
@@ -403,11 +410,19 @@ def get_outputs(self) -> IoArgoprojWorkflowV1alpha1Outputs:
     def get_command(self) -> List[str]:
         """
         Parses and returns the specified task command. This will attempt to stringify every command option and
-        raise a ValueError on failure.
+        raise a ValueError on failure. If not command is specified and no args, than this defaults to python.
```
flaviuvadan, line 413: Perhaps this can have a 
```python
Raises
-------
```
section for the `ValueError` specification?

tests/test_task.py
```diff
@@ -394,6 +394,12 @@ class SubclassTask(Task):
     assert t2.argo_task.dependencies[0] == 't'
 
 
+def test_supply_args():
+    t = Task('t', args=["arg"])
+    assert t.argo_template.container.args == ["arg"]
+    assert 'command' not in t.argo_template.container
+
+
```
flaviuvadan, line 402: What do you think about an assertion inside an existing test for the presence of `["python]` when no command and no args are specified? 🙂

flaviuvadan:
This looks like a solid addition! 🚀 Just some very minor feedback to address and then we're good to 🚢

tests/test_task.py
bchalk101: This gets tested in test `test_task_command_parses` on line 203

src/hera/task.py
bchalk101: Ive added some more info - would be good to get feedback on it

flaviuvadan:
Shipped with #95 in 2.2.0.

PR #92, opened 2022-02-11
Add tox support

bobh66:
Add tox.ini file to run tests on python 3.7, 3.8, 3.9, 3.10


bobh66:
Hi @harelwa - this PR was targeted to the development environment rather than the CI/CD environment.  I am mostly interested in being able to type "tox" on the command line and have it do all of the things.   The CI/CD problem is much larger, as you have seen, and is complicated by pipenv.  My personal opinion is that pipenv is fine for applications that only need to support one python version but it doesn't work well with libraries that need to support many versions.

harelwa:
hi @bobh66 

Thanks for replying.

In #93 I've added the run of `tox` to the `CI` flow as well, and it's [running successfully](https://github.com/argoproj-labs/hera-workflows/runs/5155415352?check_suite_focus=true) on all python versions. please see https://github.com/argoproj-labs/hera-workflows/pull/93/commits/895c4bc6c98d00cbfa78b96a0296775f586e998e

As for `pipenv` -

I think it's partly off topic as far as `tox` is considered. Or at least can be, for now.

In #93 I've tritted `pipenv` as what takes care of installing `tox` and nothing more. Hence not using `pipenv` in `tox.ini`. and the addition to `dev-packages`

Thus you get to run `tox` and have _do all the things for you_ locally regardless of `pipenv`, having `tox` taking care of dependencies. This is also one of the reasons I avoided using `pipenv` in `tox.ini` in #93 

Can please address the essential differences between `tox.ini` in #93 than the one proposed here ? [ choosing to use `pipenv` in `tox.ini` etc. ]

harelwa:
@bobh66 

As for the your observation regarding the difference between `py` _apps_ supporting a single version, as opposed to `py` _libs_ supporting multiple versions I tend do agree. 

As for alternatives, I don't know if `poetry` provides a good alternative with the required elasticity for our use case. Nevertheless, I made the necessary `pipenv` changes as it's currently `hera`'s project choice. 

On a different note, I see the recommended way of using `tox` + `pipenv` in `pipenv` docs is similar to what you proposed -

https://pipenv.pypa.io/en/latest/advanced/#tox-automation-project

I must say I'm a bit confused - 

The motivation is of not having to manage test deps in two places is clear and agreeable.

However, If your dev env is to be managed with `pipenv`, then `tox` needs to be listed as a `dev-package`. following their docs, doing -- 

```ini
commands=
    pipenv install --dev
```

will install `tox` in a venv created by `tox`. which is just a bit strange to me :\

bobh66:
Hi @harelwa 

I don't have a lot of experience with poetry, we came across it once when building dependencies for another project and found that it created a number of issues, so we have tried to stay away from it.  I would need to do more research to understand it better.

I think the general expectation for tox is that you don't install it as a dependency of the project, rather it is installed in your base python environment (like pip), so it is common to all projects and is just always there.  Otherwise you run into the chicken-egg problem of needing tox to install tox.

flaviuvadan:
`tox` was introduced in #93 so closing this one for now! Thank you for the efforts and the discussions @bobh66 🙂

Issue #73, opened 2022-01-29
Support latest Python version


hirosassa:
Hi! Thank you for developing such a great tool!
Currently, Hera supports only Python 3.7. The version is now security status, will become end of support soon.

reference: https://www.python.org/downloads/

So it is better that supporting Python 3.8+, IMO.
I'm happy to create PR for supporting newer version of Python, if it is OK.

Thank you.

flaviuvadan:
Hey @hirosassa! Thanks for the issue submission 🙂 Sure, this sounds great! I am curious, however, whether you foresee any potential future issue submissions that have to do with some potential backward incompatible changes? I have not looked at the release notes of 3.8 and I'd love to know whether you can share some potential drawbacks of this migration! My main concern is, to be perfectly honest, any syntax changes Hera might introduce, which can break existing application usage.

bobh66:
@flaviuvadan - have you considered using tox to provide test environment support for multiple python releases?   That would allow the test suite to run on every supported python release for every build and hopefully find any compatibility issues in the pipeline.   I can push a PR if you're interested.

hirosassa:
@flaviuvadan @bobh66 Thanks for the comments!

In my experiences, there's no significant changes (including syntax changes) in migration from Python 3.7 to 3.8+.
Only concern is dependent packages, but in Hera, dependencies are simple and all the dependencies supports Python 3.7+.
So this migration has no (or very small) concerns, IMO.

https://github.com/argoproj-labs/hera-workflows/blob/main/Pipfile
https://github.com/argoproj-labs/hera-workflows/blob/main/setup.py#L37-L44

hirosassa:
As @bobh66 says,  I also think using `tox` is the best way to support multiple versions of Python.

examples:
- https://github.com/spotify/luigi/blob/master/tox.ini (complex)
- https://github.com/m3dev/gokart/blob/master/tox.ini (simple)

bobh66:
Also note tox-pipenv:  https://github.com/tox-dev/tox-pipenv  to allow for tox and pipenv to work together cleanly

harelwa:
Hi,

pitching in with a bit of a different suggestion.

For CICD purposes I find [Github Actions matrices](https://ncorti.com/blog/howto-github-actions-build-matrix#:~:text=My%20favorite%20feature%20of%20Github,substitution%20when%20running%20your%20job.) more comfortable than `tox`.

Why ?

[...]

PR #93, opened 2022-02-11
Multiple Python Versions CI

harelwa:
## Purpose of PR

> related issues and PRs #73 #92 #94 

Provide multiple `python` versions **CI** ( **CD** will be handled in a different PR )

To achieve this, introduce usage of [tox](https://tox.wiki/en/latest/index.html) for testing *built* wheels in "isolated envs" of multiple python versions.

## Notes regarding similar open PRs

1. The difference between this PR and #92 is that it does not run `pipenv` in tox, but rather lets `tox` handle dependencies and venvs
2. The difference between this PR and #94 that it keep `pipenv` as the project's dev env manager 

## Tasks

- [x] add `tox.ini` with `python{3.7,3.8,3.9,3.10}` test envs
- [x] add build matrix of python versions to ci
   - ref docs: https://docs.github.com/en/actions/using-workflows/advanced-workflow-features#using-a-build-matrix
   - ref proj: https://github.com/spotify/luigi/
   - ref proj: https://github.com/samuelcolvin/pydantic/
- [x] remove caching of `Pipfile.lock`
   - `Pipfile.lock` is a `py37` lock file 
- [x] for `dev` purposes, update `Pipfile` and `Pipfile.lock`  
   - for ci, `tox` is used directly as `Pipfile.lock` is a `py37` lock file

## Notes for future changes

1. Evaluate usage of `pipenv` in `tox`
    1.1 See https://pipenv-fork.readthedocs.io/en/latest/advanced.html#tox-automation-project
    1.2 Evaluate [tox-pipenv](https://github.com/tox-dev/tox-pipenv)
2. Evaluate `poetry` to replace `pipenv`
    2.1 See #94 
    2.2 See related discussion of "who does what and why" regarding `tox` and `poetry`: https://github.com/python-poetry/poetry/discussions/4307

## Changes to Pipfile

To use `tox`, it was added to `pipfile` as a dev dependency, with the following command:

```bash
pipenv install pytest-cov "tox<4.0" tox-wheel --dev
```

`tox-wheel` was added as well to make `tox` build wheels rather than `zip`, so these can be used for publishing [ more below ].

After running `tox`, you'll find the wheels under `.tox/dist`:

```bash
❯ ls .tox/dist
hera_workflows-1.8.0rc7-py3-none-any.whl
```

`pytest-cov` was added as well so we can add a coverage reports and tests, and perhaps add a coverage test.

Even though this might be related to a different PR, I think it's also related to this PR as well, as it concerns major changes to how `hera` is tested. In this context, I've added the relevant `pytest` opt in `setup.cfg`.

## More on added `tox.ini`

> To use `tox` update  / recreate your `pipenv` virtual env according to the updated `Pipfile`

An initial `tox.ini` was created with a few `tox` envs.

The "main" env - `testenv` -- provides the first desired functionality discussed in #73 and tests 4 python versions, including `3.10` ( latest ).

To list all `tox` envs you can run:

```bash
❯ tox -a
python3.7
python3.8
python3.9
python3.10
lint
typecheck
```

And to run a specific one, e.g. `lint`, run -

```bash
❯ tox -e lint
```


harelwa:
In order to **already** have `tox` run as part of this PR checks, so changes in `tox.ini` are checked, I've changed `Run tests` step to use `tox`. This required updating `Pipfile.lock` with the additional dev-packages.

harelwa:
> Assuming we keep using `pipenv` rather than `poetry`

> The issues described below is also to an issue pointed out in #73 regarding pipenv and multiple python versions.

Regarding using `gh` actions matrices - 

If we want to use `gh` actions matrices, there is a problem with the current caching of  `Pipfile.lock` in [`.github/workflows/hera_pr.yaml`](https://github.com/argoproj-labs/hera-workflows/blob/main/.github/workflows/hera_pr.yaml) - 

```yaml
      - id: cache-pipenv
        uses: actions/cache@v1
        with:
          path: ~/.local/share/virtualenvs
          key: ${{ runner.os }}-pipenv-${{ hashFiles('**/Pipfile.lock') }}

      - name: Install dependencies
        if: steps.cache-pipenv.outputs.cache-hit != 'true'
        run: |
          pipenv sync --dev --pre
```

Since [`Pipfile.lock`](https://github.com/argoproj-labs/hera-workflows/blob/main/Pipfile.lock) in under version control, it's locked for a specific python version. However, the matrix will be for the various python versions.

I'm adding here a possible addition to [`.github/workflows/hera_pr.yaml`](https://github.com/argoproj-labs/hera-workflows/blob/main/.github/workflows/hera_pr.yaml) to demonstrate the issue - 

```yaml
[...]

jobs:
  hera-pr-build:
    timeout-minutes: 10
    runs-on: ubuntu-latest

    strategy:
      matrix:
        include:
          - python-version: 3.7
            tox-env: python3.7
          - python-version: 3.8
            tox-env: python3.8
          - python-version: 3.9
            tox-env: python3.9
          - python-version: 3.10
            tox-env: python3.10

    steps:
      - name: Set up python ${{ matrix.python-version }}
        uses: actions/setup-python@v2
        with:
          python-version: ${{ matrix.python-version }}

      - name: Checkout
        uses: actions/checkout@v2

      - name: Install pipenv
        run: |
          python -m pip install --upgrade pipenv wheel

      - id: cache-pipenv
        uses: actions/cache@v1
        with:
          path: ~/.local/share/virtualenvs
          key: ${{ runner.os }}-pipenv-${{ hashFiles('**/Pipfile.lock') }}

      - name: Install dependencies
        if: steps.cache-pipenv.outputs.cache-hit != 'true'
        run: |
          pipenv sync --dev --pre

[...]

      - name: Run tests
        env:
          TOXENV: ${{ matrix.tox-env }}
        run: |
          pipenv run tox

```

The cache will be _wrong_ for py version `!=` 3.7

One way to deal with it is - 

1. remove `pipfile.lock` from version control
2. check for cache
    2.1. if `false` create lock  file and cache ( first hit will always be `false` )
    2.2 if `true` skip install deps

What do you think ?

setup.cfg
```diff
@@ -1,3 +1,8 @@
+[pytest]
+addopts = --cov=hera -vv
+# for now, keep this commented out, as it's used explicitly in Makefile
+; testpaths = tests
+
```
hirosassa, line 5: This configuration can be put on tox.ini. 
ref: https://docs.pytest.org/en/6.2.x/customize.html#tox-ini

I prefer the test related configurations is placed in tox.ini. What do you think about this?

setup.py
```diff
@@ -25,6 +25,9 @@
         "Topic :: Scientific/Engineering",
         "License :: OSI Approved :: MIT License",
         "Programming Language :: Python :: 3.7",
+        "Programming Language :: Python :: 3.8",
+        "Programming Language :: Python :: 3.9",
+        "Programming Language :: Python :: 3.10",
```
hirosassa, line 30: Yeah! 🎉

tox.ini
```diff
@@ -0,0 +1,28 @@
+[tox]
+envlist = python{3.7,3.8,3.9,3.10}
```
hirosassa, line 2: [nits] I often see the expressions like below:

```suggestion
envlist = py{37,38,39,310}
```

The merit of this is less typing on command line like:
```
tox -e py39
```

tox.ini
```diff
@@ -0,0 +1,28 @@
...
+
+[testenv:typecheck]
+skipsdist = True
+deps =
+    mypy
+    {[testenv]deps}
+commands=
+    make typecheck
+
+
```
hirosassa, line 28: should remove trailing white space and new lines.
```suggestion
```

Pipfile
```diff
@@ -13,6 +13,9 @@ flake8 = "*"
 mypy = "*"
 build = "*"
 isort = "*"
+pytest-cov = "*"
+tox = "<4.0"
```
hirosassa, line 17: [ask] Why is this dependency strictly specified?

hirosassa (commented):
@harelwa Thank you for implementing this! I added small comments.

hirosassa:
@harelwa I agree with your solution!

> One way to deal with it is -
>    1. remove pipfile.lock from version control
>    2. check for cache
>    2.1. if false create lock file and cache ( first hit will always be false )
>    2.2 if true skip install deps

setup.cfg
harelwa: Hi @hirosassa !

I keep this in `setup.cfg` so when one run's the tests **not** using `tox`, he uses the same `pytest` configuration.

For example - 

```bash
> make test
> pytest ./tests/[specific-test]
```

Meaning - I want to keep `tox` for the purpose of isolated builds, not for the source of all of the configuration.
I'm aware of other projects [ also `luigi` ] using it to _also_ to replace `setup.cfg`. I just don't like it :) 

acceptable ?

tox.ini
harelwa: I also prefer the `py`[ver].

but I encountered the following error while developing it / running it -

```bash
ERROR:  py37: InterpreterNotFound: python3.7
```

And changing it to `python`[ver] solved it.

See also here -
https://stackoverflow.com/questions/63180349/error-py37-interpreternotfound-python3-7-when-running-tox-from-github

Pipfile
harelwa: As of `tox` version `4.0` ithe _entry point name_ is `tox4` and not `tox`.

From https://github.com/tox-dev/tox/blob/4.0.0b1/docs/changelog.rst

> ... 4.0.0a1 is a complete rewrite of the package

When it stabilizes, we can add a `>=4.0` and replace `tox` cli cmd with `tox4` cli cmd everywhere.

tox.ini
hirosassa: Oh, I didn't know this! I understood. Thanks!

Pipfile
hirosassa: Understood! Thanks!

setup.cfg
hirosassa: Yeah, I understand what you intended 👍 I agree with this.

> I want to keep tox for the purpose of isolated builds, not for the source of all of the configuration.

Just for your information, you can write configuration of `pytest` on tox.ini without using `tox`.

ref:
https://github.com/pytest-dev/pytest/blob/e86bf7fda0a73aebee7d3741e1106d58cf8dec85/src/_pytest/config/findpaths.py#L88-L111

harelwa: I was not aware `tox.ini` is a part of the [defaults configs list](https://github.com/pytest-dev/pytest/blob/e86bf7fda0a73aebee7d3741e1106d58cf8dec85/src/_pytest/config/findpaths.py#L93) !

The question now rises regarding other tools I guess.. 

Do you think it's worth the effort to check with the rest of the tools if they behave the same ?

I see [luigi](https://github.com/spotify/luigi/blob/master/tox.ini) indeed got rid of `setup.cfg` all together in favour of `tox` ..

flaviuvadan:
You folks have done a lot of work for this and had loads of valuable discussions. This looks _awesome_ 🚀 I like the way this structure delegates virtual environment creation to `tox` and still maintains `pipenv` as the default venv management tool for now; otherwise, things can get very complicated to deploy in the same PR as there are too many changes, as @hirosassa suggested [here](https://github.com/argoproj-labs/hera-workflows/issues/73#issuecomment-1034130260). I also love that there's already a PR for poetry's introduction. If you folks don't mind, I will approve this one, then look into poetry a bit more so I can perfectly understand #94 as well.

hirosassa:
@flaviuvadan Thanks for your comment! I agree with your opinion 👍

harelwa:
+ I'll edit the PR subject + description accordingly ( e.g. - it does not include CD .. )

flaviuvadan:
Thanks again @harelwa and @hirosassa 🙂 I'll wait until the docs are updated for development then ship this. Enjoy the time away @harelwa!

harelwa:
hi @flaviuvadan 

perhaps overlooked in my previous comment -docs are updated.

Pipfile.lock
```diff
@@ -35,10 +34,6 @@
             "editable": true,
             "path": "."
         },
-        "hera-workflows": {
```
flaviuvadan, line 38: Why is this no longer part of the lockfile?

Pipfile
```diff
@@ -13,6 +13,9 @@ flake8 = "*"
...
+tox-wheel = "*"
 
 [requires]
 python_version = "3.7"
```
hirosassa, line 21: @harelwa We should remove this to support multiple Python versions.

harelwa:
@flaviuvadan - my bad. i see the changes to README are not pushed for some reason. 

I can take of this only next wednesday :\
if someone wishes to pick up from where I left and complete the PR so it does not wait to next wednesday — by all means.

Pipfile
harelwa: thanks @hirosassa 

indeed

i did not manage to play with pipenv and get a clear “howto get started with development” with any python version without having pipenv change the git state… hence I did not change this yet. i have a bit more digging to do

i’ll be able to cont. work on this PR next wednesday :\  If you’d like to pickup from where I left and complete the PR — by all means. o.w we’ll wait till next week.

flaviuvadan:
> I can take of this only next wednesday :

Please tag me whenever you have a chance to continue on this 🙂 I will immediately circle back to this so we can deliver it.

Pipfile.lock
flaviuvadan: @harelwa just a reminder about this question!

flaviuvadan (approved):

flaviuvadan:
Thank you again for the efforts @harelwa and @hirosassa! 🚀

hirosassa:
@harelwa @flaviuvadan Thanks for the great discussions!

PR #94, opened 2022-02-12
Support multiple python using tox with Poetry

### LINKED ISSUE #73: Support latest Python version

hirosassa:
fix #73 

## What is this PR?

Related to #93, I implemented support for multiple Python versions using `tox`.
Different from #93, this PR uses `Poetry` as a package management to simplify configurations of linters, tests and builds.


hirosassa:
[note] 

I said

> applying all of the solutions is not preferable for the developers and such a big change will cause bugs

but @harelwa implemented nice PR to fix same issue using pipenv.
So I implemented another version to compare.


@harelwa @bobh66 @flaviuvadan plz review!

harelwa:
Hi @hirosassa 

That's great !!

I think we [ as in `hera` project ] needs to decide on the following two things, and unite the focus into a single PR ( this + #93 + #92 ).

1. The `makefile`

In your changes here, you have basically replaced the `make` + `Makefile` with `tox` + `tox.ini`

I'd love if we can address what I wrote in the `Makefile` section of #93 description, which is also related to my answer to you here regarding `setup.cfg` - https://github.com/argoproj-labs/hera-workflows/pull/93#discussion_r805132881

2. Drop `Pipenv` In favour of `Poetry`

@flaviuvadan @bobh66 @bchalk101 @GuyPozner ?

flaviuvadan:
Please see my comment [here](https://github.com/argoproj-labs/hera-workflows/pull/93#issuecomment-1038459672). I think Hera should introduce the usage of `tox` at first to get a stable build, release 1-2 features using that, so we test that setup. The introduction of `tox` gives us the benefit of testing the package in multiple Python versions, which is an amazing start for the goal of supporting multiple Python versions! After 1-2 releases, I suggest we introduce this change 🙂 I am excited about the ideas you folks brought for improving Hera's testing and stability, so I am excited to get this rolling as well.

tox.ini
```diff
@@ -0,0 +1,36 @@
+[tox]
```
flaviuvadan, line 1: So now, rather than invoking `make test`, we invoke `tox -e test`, right? If so, this is great! What do you think about keeping the `Makefile` around but replacing the inner calls with `tox -e`? 🙂 Curious whether that might be nicer for people who may know `make` but not necessarily `tox`.

flaviuvadan:
@hirosassa I think we can merge this soon! Tested the CICD improvements @harelwa and everything deployed just fine 🚀

hirosassa:
@flaviuvadan Thanks for informing this!

tox.ini
hirosassa: > So now, rather than invoking make test, we invoke tox -e test, right?

There's no exact counterpart of `make test` in this branch.
`tox` is the task runner on each isolated environment.
If your want to run test for Python 3.8, then you should invoke `tox -e py38`. 
`make lint`, `make format` correspond to `tox -e lint` `tox -e format`

How about making Makefile below?
```Makefile
testall: # run all test in all Python version
	tox

test-pyXX: # run test on PythonXX
	tox -e pyXX
...
```

flaviuvadan:
@hirosassa now that we merged the CICD changes, I think we can pick this up again and deliver the migration to Poetry!

.github/workflows/cicd.yaml
```diff
@@ -160,28 +110,14 @@ jobs:
...
-          twine check dist/*
-
-      - name: publish package to pypi
-        uses: pypa/gh-action-pypi-publish@release/v1
-        with:
-          user: __token__
-          password: ${{ secrets.PYPI_API_TOKEN }}
-          verbose: true
-          skip_existing: true
+          poetry publish --build --username $TWINE_USERNAME --password $TWINE_PASSWORD
```
hirosassa, line 123: For poetry, `publish` sub-command builds package and uploads to PyPI with `py3-none-any` wheel.

.github/workflows/cicd.yaml
```diff
@@ -47,67 +48,15 @@ jobs:
...
-#          name: pypi-build-arts
-#          path: .tox/dist
-
       - name: upload coverage files
         uses: actions/upload-artifact@v2
         with:
           name: coverage-data
           path: ".tox/.coverage.*"
 
-  build-py3-none-any:
```
hirosassa, line 74: This jobs is removed because `poetry` publish this automatically

Makefile
```diff
@@ -4,15 +4,13 @@ EXAMPLES_FOLDER=examples
...
 lint:
-	flake8 $(SRC_FOLDER)
+	pflake8 $(SRC_FOLDER)
 
-format: isort
+format:
 	black --verbose $(SRC_FOLDER) $(TEST_FOLDER) $(EXAMPLES_FOLDER) ${CONFTEST}
-
-isort:
 	isort $(SRC_FOLDER) $(TEST_FOLDER) $(EXAMPLES_FOLDER) ${CONFTEST}
```
hirosassa, line 14: I integrate `isort` task with `format` task (isort is also format).

pyproject.toml
```diff
@@ -1,17 +1,87 @@
+[tool.poetry]
+name = "hera"
+version = "2.6.1"
```
hirosassa, line 3: I naively implemented `version`. I prefer using [poetry-dynamic-versioning](https://github.com/mtkennerly/poetry-dynamic-versioning) to add version by `git tag`.

@flaviuvadan Do you have any ideas on this?

hirosassa (commented):
@flaviuvadan I updated. Please review!

.github/workflows/cicd.yaml
hirosassa: For example, this [repo](https://github.com/m3dev/gokart) (I'm a mentainer of this) use poetry publish in [CI](https://github.com/m3dev/gokart/blob/master/.github/workflows/publish.yml#L27), then PyPI has [this files](https://pypi.org/project/gokart/#files)

pyproject.toml
flaviuvadan: I quite like the verbosity of `VERSION` thus far but I am definitely open to other approaches! Does the version need to be specified here? I wonder if we can leave the `setup.py` as it is?

.github/workflows/cicd.yaml
```diff
@@ -47,67 +48,15 @@ jobs:
           tox
 
       - name: tests
-        env:
-          ENABLE_BDIST_EXT_MODULE: 'yes'
-          TOXENV: python${{ matrix.python-version }}
```
flaviuvadan, line 52: Does tox run over the specified versions [here](https://github.com/argoproj-labs/hera-workflows/blob/main/tox.ini#L2)? Also, will it automatically run `testenv` in this case?

hirosassa: But, this maybe confusing for the users who are not familiar with tox and github actions

hirosassa: I agree with you! I'll fix this.

.github/workflows/cicd.yaml
```diff
@@ -13,11 +13,31 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ ubuntu-latest, macos-latest ]
-        python-version: [ '3.7', '3.8', '3.9', '3.10' ]
-    env:
-      PYVER: ${{ matrix.python-version }}
-      OS: ${{ matrix.os }}
+        include:
```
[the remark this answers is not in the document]
hirosassa: This verbosity comes from difference between [current main](https://github.com/argoproj-labs/hera-workflows/blob/main/tox.ini#L2) and [this branch](https://github.com/hirosassa/hera-workflows/blob/support-latest-python/tox.ini#L2).
The merit of this branch implementation is that if you'd like to run test, you type `tox -e py38` for example.
In the current main, you should type `tox -e python3.8`.
The frequency of changes of github actions configuration is low (once or twice in a year?), but running tests is frequent, so this verbosity is acceptable, IMO.

pyproject.toml
hirosassa: > I quite like the verbosity of VERSION

Yeah, I understand this 👍. This is very simple and easy to understand what the contributors should do. 
Currently we can't use VERSION file on pyproject.toml (pyproject.toml is `toml` so we could not load file dynamically).

> I think it's ok to leave this as is but update the contribution guideline to suggest that users increment this file rather than VERSION.

I updated this [here](https://github.com/argoproj-labs/hera-workflows/pull/94/files#diff-eca12c0a30e25b4b46522ebf89465a03ba72a03f540796c979137931d8f92055R15)

.github/workflows/cicd.yaml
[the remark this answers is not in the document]
flaviuvadan: Got it. Thank you so much for the thoughtful response 🚀

.github/workflows/cicd.yaml
```diff
@@ -127,13 +98,14 @@ jobs:
           name: coverage-data
           path: .tox
 
-      - name: list coverage files
-        run: |
-          find .tox -name ".coverage.*"
+      # - name: list coverage files
```
flaviuvadan, line 101: Any chance we can uncomment this? I think it was added for information purposes and also in preparation for adding the coverage badge 🤔 so we know what files can be accessed.

README.md
```diff
@@ -82,11 +82,10 @@ There are multiple ways to install Hera:
...
-If you plan to submit contributions to Hera you can install Hera in a virtual environment managed by `pipenv`:
+If you plan to submit contributions to Hera you can install Hera in a virtual environment managed by `poetry`:
 
 ```shell
-pipenv shell
-pipenv sync --dev --pre
+poetry install
 ```
 
 In you activated `pipenv` shell, you can utilize the tasks found in `Makefile`, e.g.:
```
flaviuvadan, line 91: Can we remove this section of the doc since the Makefile was removed in this PR?

tox.ini
```diff
@@ -26,43 +27,47 @@ setenv =
...
-    make typecheck
+    poetry install
+    poetry run make typecheck
+
+[gh-actions]
+python =
+    3.7: py37
+    3.8: py38
+    3.9: py39
+    3.10: py310, lint, typecheck, coverage #, format
```
flaviuvadan, line 73: What's the `#, format` comment? 🤔

flaviuvadan (commented):
Left some very minor questions and tasks to trim some part of the documentation that still mentions the `make` command. This is so cool @hirosassa. I am very thankful for the contribution. I know this has been open for a while and I am so glad we're finally converging on the final version 🙂 I am going to merge this after some of the small cleanup and main merge into this so I will be on the lookout.

tox.ini
hirosassa: @flaviuvadan These lines are for `tox-gh-actions`,  so I removed these.

README.md
hirosassa: Thanks this is I missed. Removed.

.github/workflows/cicd.yaml
hirosassa: Recovered!

hirosassa:
@flaviuvadan Thank you for your comments! I applied the changes. And I found conflicting on `VERSION` .
It looks I can't resolve because I don't have write access on this repo.

flaviuvadan (approved):

hirosassa:
@flaviuvadan Thank you for the fruitful discussions!

Issue #87, opened 2022-02-08
Add support for raw value task parameters


flaviuvadan:
Some users want to extract pod metadata, such as IP, from a running task and expose it to others. This can be achieved through something like:
```YAML
    arguments:
      parameters:
        - name: port
          value: "5051"  // raw values
        - name: whatever-task-field
          value: "{{steps.TASK.FIELD}}"  // step
```

https://argoproj.github.io/argo-workflows/variables/

Suggestion:
- have an enum with all available variables
- help users by only requesting a specification of the field and task name: ```Whatever(Task.name, Field.IP)``` (sample)

bchalk101:
I'm happy to jump on this one

GuyPozner:
Great Job Guys, Thanks:)

PR #95, opened 2022-02-13
Access task variables

bchalk101:
Address https://github.com/argoproj-labs/hera-workflows/issues/87


examples/variables.py
```diff
@@ -0,0 +1,63 @@
...
+from hera.workflow_service import WorkflowService
+
+
+def server():
+    pass
+
+
+def consumer(server_ip):
+    print(server_ip)
+
```
flaviuvadan, line 18: Leftover, perhaps?

examples/variables.py
```diff
@@ -0,0 +1,63 @@
...
+            self.send_response(200)
+            self.send_header("Content-type", "application/json")
+            self.end_headers()
+            self.wfile.write(bytes("{'name':'John'}", "utf-8"))
+
+    webServer = HTTPServer(("0.0.0.0", 8080), MyServer)
+    try:
+        webServer.serve_forever()
+    except KeyboardInterrupt:
+        pass
```
flaviuvadan, line 38: This is a very interesting example! I do not have experience with daemon tasks - how does this task know when to be "done"? 🙂

examples/variables.py
```diff
@@ -0,0 +1,63 @@
...
+    response = connection.getresponse()
+    print(response.read())
+
+
+namespace = "argo"
+# TODO: replace the domain and token with your own
+ws = WorkflowService(host='my-argo-server.com', token="token", namespace=namespace)
+w = Workflow('variables', ws)
+d = Task('daemon', server, daemon=True)
+variables = InputParameterAsEnv(name='SERVER_IP', value="{{tasks.daemon.ip}}")
```
flaviuvadan, line 57:
```suggestion
variables = InputParameterAsEnv(name='SERVER_IP', value=d.ip)
```

src/hera/input.py
```diff
@@ -48,3 +54,17 @@ def sleep(seconds: int, sentence: str = 'Hello, world!'):
...
+    value: str
+
+    def get_argument_parameter(self):
+        return IoArgoprojWorkflowV1alpha1Parameter(name=self.name, value=self.value)
+
+    def get_input_parameter(self):
+        return IoArgoprojWorkflowV1alpha1Inputs(name=self.name)
+
+    def get_env_spec(self):
+        return EnvSpec(name=self.name, value=f"{{{{inputs.parameters.{self.name}}}}}")
```
flaviuvadan, line 70: I love this concept. What do you think about calling this `Variable`? 🙂 The docstring can outline that the variable will be presented to the task in the form of an environment variable named `name`.

flaviuvadan:
This looks great so far! Left a few questions 🙂 I'd love to see this materialize as it strikes me to be a pretty powerful concept and has many applications for distributed training in the field of ML 💪🏻

flaviuvadan:
@bchalk101 may I help push this forward somehow? This is a very exciting piece of development 🙂

examples/variables.py
bchalk101: I cleaned this up as it implied something incorrect - Argo simply shuts down the daemon once the other Tasks have been completed.

bchalk101:
@flaviuvadan - got caught up with other things.... but should be good now :)

tests/test_task.py
```diff
@@ -3,6 +3,7 @@
 from argo_workflows.model.security_context import SecurityContext
 from argo_workflows.models import IoArgoprojWorkflowV1alpha1Inputs
 from argo_workflows.models import Toleration as _ArgoToleration
+from click import argument
```
flaviuvadan, line 6: Perhaps a tiny import mistake? 🙂

examples/variables.py
flaviuvadan: Ah, I see, so marking something as a daemon keeps the pod alive but Argo terminates the pod after it's done 👍🏻

examples/variables.py
```diff
@@ -0,0 +1,49 @@
...
+# The daemoned task operates as server, serving an example payload, with the second task operating as a client, making
+# http requests to the server
+
+from hera.task import Task
+from hera.variable import VariableAsEnv
+from hera.workflow import Workflow
+from hera.workflow_service import WorkflowService
+
+
+def server():
```
flaviuvadan, line 12: Amazing 🚀

src/hera/variable.py
```diff
@@ -0,0 +1,32 @@
+from argo_workflows.models import (
+    IoArgoprojWorkflowV1alpha1Inputs,
+    IoArgoprojWorkflowV1alpha1Parameter,
+)
+from pydantic import BaseModel
+
+from hera.env import EnvSpec
+
+
+class VariableAsEnv(BaseModel):
```
flaviuvadan, line 10: While not needed for this PR, I am wondering what do you think about naming this `TaskVariable`/`TaskVariableAsEnv`? Looking through the Argo docs it seems that we can expose many variable types. For instance, if Hera will support `Step` in the future, or global workflow variables, which I assume may have their respective properties to construct the required string `{{...}}`. Perhaps this is a bridge we'll cross at that point as this does not require over-engineering but really curious what you think! 🙂

flaviuvadan (commented):
Looks great! 🚀 I left two questions - one for an import removal, and a meta one about the future. Let's ship right after the removal of the `click` import. 

I noticed that this releases 2.2.0 and the #91 releases 2.3.0. What do you think about packaging these two together in a single release of 2.2.0? It may be a bit odd for some users to have 2 MINOR versions released at 10 minutes apart, or some other short interval 🙂  We can merge this one into #91, or vice versa (doesn't really matter), and ship 2.2.0 when the build is done!

tests/test_task.py
bchalk101: 🤭  ..... removed

src/hera/variable.py
bchalk101: I could be wrong, but my understanding is that the VariableAsEnv part is the generic part. What the task and step expose is the specific, ie in the future, a `Step` may have a property `ip` which will return `"{{{{steps.<STEPNAME>.ip}}}}`. And a step will still be able to take a `VariableAsEnv`.

flaviuvadan (approved):

flaviuvadan:
Thank you for the contribution and packaging the releases together! 🚀

PR #101, opened 2022-02-24
CICD Improvements

harelwa:
## About

`this is a work in progress; it aims to check feasibility for further discussion`

Following #93, further improve CI + CD

## Tasks

- [x] add coverage reports tasks and ci
   - [x] store all `py vers` X `OS` coverage files
   - [x] combine all `py vers` X `OS` coverage files
   - [x] create html cov
   - [x] store coverage report artifact  
- [x] add optional binary distribution to wheels via `setup.py`
   - [x] add a `none-any` build just in case and publish it
   - [x] add `sdist` build ( `.tar.gz` ) build for publishing 
- [x] store built wheels by tox
   - **reason** for publishing to tested wheels
- [x] build wheels for all distributions ( matrix = `py ver` + `os` )
- [x] build `py3-none-any` wheel + `sdist` ( .tar.gz )
- [x] consolidation hera CICD into a single `gh` workflow
   - **motivation** re-use `test / build` job as a pre-publishing job
   - **motivation** to leverage artifacts passing between `jobs` rather than between different workflows
   - [x] add `publish` job ( similar to exiting one )
   - [x] change event triggers 
   - [x] add `if:` to `publish` job
   - [x] **delete** file `.github/workflows/hera_build_and_publish.yaml`
   - [x] **rename** file `.github/workflows/hera_pr.yaml` -> `.github/workflows/cicd.yaml` ( or .. )
   - [x] **fix** jobs badge workflow name `hera_build_and_publish.yaml` -> `cicd.yaml` ( or .. ) in `README.md`
- [ ] **enable** upload to `pypi-test` for the purpose of validating entire `publish` job ( see `publish` job failure )
- [ ] **replace** upload to `pypi-test` with upload to `pypi` in `publish` job
- [ ] **setup**  [codecov](https://github.com/apps/codecov) **if** we want a `coverage` badge to be introduced in this PR ( or propose a different solution )

`inspiration credit` https://github.com/samuelcolvin/pydantic/blob/master/.github/workflows/ci.yml


harelwa:
`i love coverage and observability, what can I do`

`hera` has **95%** coverage now, which is great.

Unfortunately, it's not observable in CI and.or for developers and maintainers. So I gave a try of adding a report to CI...

To create a proper report, one needs to `combine` the coverage reports created _by all py versions builds tests_.

As you can see if you look [inside](https://github.com/argoproj-labs/hera-workflows/runs/5326727125?check_suite_focus=true) -- this fails .. why ?
 
As`tox` runs the tests in _isolated_ envs, with installation of `hera`, `coverage` runs on this installation .. 
thus a report can't be created, as in the `coverage-report` gh-wf job these `tox` envs never existed.

> uploading and downloading all these envs as artifacts is not an option `afaic`; they are too big

while this is annoying, it took me again to [luigi](https://github.com/spotify/luigi) to see how they dealt with it.

`luigi` uses [codecov](https://docs.codecov.com/docs) uploader to "solve this" but actually get much more from what I can see

> `note` they are using it in a way that will soon be deprecated, so the reference is relevant only in part

[codecov](https://docs.codecov.com/docs) is free for public projects, and from what I can tell can[ integrate easily](https://github.com/marketplace/actions/codecov), "solves" this issue, and provide with some quite cool visibility features built into github - 

https://docs.codecov.com/docs/github-checks

@flaviuvadan - I'd like to give `covecov` a try. what do you say ? 

https://github.com/apps/codecov

harelwa:
@flaviuvadan - for some reason I can't _mention_ https://github.com/hirosassa ; can you please add him to the discussion ?

flaviuvadan:
> @flaviuvadan - for some reason I can't _mention_ https://github.com/hirosassa ; can you please add him to the discussion ?

@hirosassa 🙂

hirosassa:
> hera has 95% coverage now, which is great.

Oh, this is great!

I'll take a look about failure CI.

harelwa:
@hirosassa @flaviuvadan — as for the coverage combine issue, the following looks like a the solution —

https://github.com/nedbat/coveragepy/issues/597

https://github.com/nedbat/coveragepy/issues/597#issuecomment-554766758

i’ll gove it a try. 

nevertheless— @flaviuvadan — would you like to install codecov so we can try it out ?

thanks,

harel

harelwa:
Hi @flaviuvadan 

I've pushed a solution for coverage combine issue described above, with no use of `codecov`.

For some reason newly pushed commits to this PR do not trigger _checks_ ( except the `license` check :\ ), and thus I can't verify it really works and continue with the rest of this PR tasks.

Do you maybe know why ?

Thanks,

Harel

harelwa:
hi @flaviuvadan @hirosassa 

> sorry for the previous "spam"; i've deleted the no longer irrelevant questions.

(1) coverage combine fixed

I've added a solution of the `coverage combine` issue. it took a bit longer then expected; eventually i used `tox`s own `tox.ini` as a ref - https://github.com/tox-dev/tox/blob/master/tox.ini which did the trick. you can see the result in https://github.com/argoproj-labs/hera-workflows/runs/5456858785?check_suite_focus=true

(2) next tasks and wrap up

> let's agree on these

I'd like to consolidate hera's CICD into `a single workflow file`, for purposes of artifacts re-use, eliminating the current duplication of `test/build` using `job` dependencies etc.

I've already added a `publish` job, which currently only downloads the build artifacts ( wheels ) and lists them - 

https://github.com/argoproj-labs/hera-workflows/runs/5456863640?check_suite_focus=true

These artifacts can be uploaded to `pypi` "as is" **without** first uploading to `pypi test` like the current CD. 
This is thanks to `tox` which tests those build via install in (so-called) isolated envs. 

Such a change will also require a - 

(2.1) `if` condition 

```yaml
  publish:
    needs: [coverage]
    runs-on: ubuntu-latest
    if: "success() && (startsWith(github.ref, 'refs/tags/') || github.ref == 'refs/heads/master')"
```

(2.2) change of the `workflow` event triggers

```yaml
on:
  push:
    branches:
      - main
  pull_request: {}
```

What say ya ?

hirosassa (changes requested):
Thanks for the great implementation! I added a few comments.

setup.cfg
```diff
@@ -1,7 +1,7 @@
-[pytest]
-addopts = --cov=hera -vv
-# for now, keep this commented out, as it's used explicitly in Makefile
-; testpaths = tests
+[coverage:run]
+relative_files = True
+source =
+    src/hera
```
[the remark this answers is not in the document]
harelwa: yep. and tox.ini indeed has pytest and coverage configuration. this is a left over. thanks for pointing it out. removed.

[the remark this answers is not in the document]
hirosassa: 👍

.github/workflows/hera_pr.yaml
```diff
@@ -3,7 +3,7 @@ name: Hera PR build
 on: pull_request
 
 jobs:
-  hera-pr-build:
```
flaviuvadan, line 6: Is the job name representative of the distribution that is going to be built? 🤔

.github/workflows/hera_pr.yaml
```diff
@@ -41,6 +41,93 @@ jobs:
...
+
+      - name: Check build
+        run: |
+          twine check .tox/dist/*
+
+      - name: upload build artifacts
+        uses: actions/upload-artifact@v2
+        with:
+          name: pypi-build-arts
+          path: .tox/dist
```
flaviuvadan, line 57: This seems to publish the constructed distributions as artifacts. To be clear, these are different than the actually published package, correct?

.github/workflows/hera_pr.yaml
```diff
@@ -41,6 +41,93 @@ jobs:
...
+
+      - name: upload coverage files
+        uses: actions/upload-artifact@v2
+        with:
+          name: coverage-data
+          path: ".tox/.coverage.*"
```
flaviuvadan, line 63: Does the publication of coverage files allow one to construct a test coverage GitHub badge? I am a bit unfamiliar with the necessary setup of coverage so I am very excited this is added, and I would love to see the report on the main repo README! I must say a direct thank you here, as a comment, for the efforts 🚀

.github/workflows/hera_pr.yaml
```diff
@@ -41,6 +41,93 @@ jobs:
...
+        run: |
+          ls -la dist/
+
+      - name: install twine
+        run: |
+          python -m pip install --upgrade twine
+
+      - name: Check build
+        run: |
+          twine check dist/*
```
flaviuvadan, line 133: This is a check of the built distribution published via the PR, correct?

flaviuvadan:
This looks awesome 🙂 I left some questions mostly to clarify my own understanding of the contribution.

flaviuvadan:
I still think there's a lot of value in first publishing to the Test PyPI as it ensures the whole installation and usage workflow works as expected. I am open to be convinced of this not being the case, though. I recall an issue in which a broken Hera package was published - the publication step and tests were successful because of accessibility to Hera via the editable mode, but the installed package from PyPI was broken because of some `setup.py` changes. Any thoughts @harelwa and @hirosassa?

flaviuvadan:
Oh, another question if I may - does the consolidation in a single GHA workflow file support differentiation between PRs and `main` pushes to the degree that is necessary in order to support publication of PR artifacts, such as coverage, and then publish the Hera package independently when the PR is merged?

flaviuvadan:
@hirosassa do you think we should merge this first and then https://github.com/argoproj-labs/hera-workflows/pull/94, maybe? I wonder what's the best order and mix of PRs that minimizes the effort you and @harelwa need to put in to merge these (due to conflicts).

harelwa:
Hi @flaviuvadan 

> I still think there's a lot of value in first publishing to the Test PyPI as it ensures the whole installation and usage workflow works as expected. I am open to be convinced of this not being the case, though. I recall an issue in which a broken Hera package was published - the publication step and tests were successful because of accessibility to Hera via the editable mode, but the installed package from PyPI was broken because of some setup.py changes. Any thoughts @harelwa and @hirosassa?

If you recall, I was the one to point the `publish issue` #51 where I also suggested a solution https://github.com/argoproj-labs/hera-workflows/issues/51#issuecomment-1010464186 and ended up hopefully contributing the suggestion in this 
PR !

I'll try and explain why publishing to Test PyPi is an overhead with the CI flows suggested in this `PR` - 

As we are using `tox` now, the actual wheels that are going to be published (90% true; I'll explain more below) are _installed_ and _tested_ thus you do not expect pitfalls like we saw in #51 usually involving using `pytest` on sources, rather on build due to [PyTest PYTHONPATH Fun](https://docs.pytest.org/en/6.2.x/pythonpath.html). I've we also `twine check` explicitly to verify pypi artifacts health.

having said that - of course we can add such a job  / step if you are still don't feel comfortable with that. 

What I did want to to use Test PyPi is to "test" the real publish flow in in single GH workflow in this PR  ( more below; it's not pushed yet )

> Oh, another question if I may - does the consolidation in a single GHA workflow file support differentiation between PRs and main pushes to the degree that is necessary in order to support publication of PR artifacts, such as coverage, and then publish the Hera package independently when the PR is merged?

(1) `coverage` 

If we wish to add a `coverage` shield ( and we do :), we can have such "external" uploads taken care of in a different work flow, but actually also in the same one. 

There are a few options of the `coverage`, e.g. - 

`luigi` use [codecove](https://github.com/apps/codecov); see https://github.com/spotify/luigi/blob/master/.github/workflows/pythonbuild.yml

`pydantic` uses [smokeshow](https://smokeshow.helpmanual.io/); see https://github.com/samuelcolvin/pydantic/blob/master/.github/workflows/upload-previews.yml

(2) `publishing` 

Indeed. I provided a snippet for how it's done in https://github.com/argoproj-labs/hera-workflows/pull/101#issuecomment-1061247450 (2.1) + (2.2)

`refs` 
- https://github.com/samuelcolvin/pydantic/blob/master/.github/workflows/ci.yml
- https://github.com/samuelcolvin/pydantic/blob/master/.github/workflows/upload-previews.yml

> @hirosassa do you think we should merge this first and then https://github.com/argoproj-labs/hera-workflows/pull/94, maybe? I wonder what's the best order and mix of PRs that minimizes the effort you and @harelwa need to put in to merge these (due to conflicts).

I tend to agree it would be wiser to merge this PR first

.github/workflows/hera_pr.yaml
harelwa: With pleasure. 

Yes, we are definitely _going for a n honorable shield_. 

See `coverage` in https://github.com/argoproj-labs/hera-workflows/pull/101#issuecomment-1065169028

.github/workflows/hera_pr.yaml
harelwa: Not quite. If we consolidate to a single GH workflow like I suggested, the `build` ( and test ) job will run as a dependency of `publish` job on push to `main` as well ( see more in https://github.com/argoproj-labs/hera-workflows/pull/101#issuecomment-1065169028 ), and thus provide us with the necessary Distribution builds ( wheels ) - As you can see _Distribution_ builds added to `setup.py` are `OS` + `Py` version specific.

HOWEVER ! what's usually done, is building `also` a `py-none-any` wheel + `tar.gz` and publishing it as well ( like `hera` publishes now ), e.g. ( from https://pypi.org/project/pydantic/ )

Screen Shot 2022-03-11 at 4 44 18 PM

setup.cfg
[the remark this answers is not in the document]
harelwa: this deletion actually fixed the `combine` ! the **combined** result before the deletion was `0%` 😖 . thank you @hirosassa for catching this

.github/workflows/hera_pr.yaml
flaviuvadan: Yes, thank you for clarifying! 🙂

flaviuvadan:
@harelwa thank you for the explanations 🙂 I feel like I have a much richer understanding now. I think this PR looks good and we can ship once the builds are fixed. Then we can focus on @hirosassa's Poetry contribution and make that fit with the contributions here!

README.md
```diff
@@ -9,6 +9,7 @@ and its crew were specially protected by the goddess Hera.
 
 [![Build](https://github.com/argoproj-labs/hera-workflows/actions/workflows/hera_build_and_publish.yaml/badge.svg)](https://github.com/argoproj-labs/hera-workflows/blob/main/.github/workflows/hera_build_and_publish.yaml)
 [![Pypi](https://img.shields.io/pypi/v/hera-workflows.svg)](https://pypi.python.org/pypi/hera-workflows)
+[![CondaForge](https://img.shields.io/conda/v/conda-forge/hera-workflows.svg)](https://anaconda.org/conda-forge/hera-workflows)
```
flaviuvadan, line 12: 🚀

.github/workflows/hera_pr.yaml
```diff
@@ -3,16 +3,19 @@ name: Hera PR build
...
+  build:
+    name: build py${{ matrix.python-version }} on ${{ matrix.os }}
     timeout-minutes: 10
-
-    runs-on: ubuntu-latest
-
     strategy:
       fail-fast: false
       matrix:
+        os: [linux, macos]
```
hirosassa, line 12: This line should be 
```suggestion
        os: [ubuntu-latest, macos-latest]
```
or we can specify version like `ubuntu-20.04, macos-11`

ref
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idruns-on

harelwa: done

hirosassa (approved):
Looks good to me! Thanks for your great contribution.

> question do you want to add windows-latest to the os matrix ?

I think it is unnecessary because `Argo` and `Hera` is not running on Windows and Windows has `WSL` currently, IMO.

harelwa:
> Hi @flaviuvadan + @hirosassa
> 
> I think we are pretty much done here.
> 
> Here are the final leftovers:
> 
> * [x]  **delete** file `.github/workflows/hera_build_and_publish.yaml`
> * [x]  **rename** file `.github/workflows/hera_pr.yaml` -> `.github/workflows/cicd.yaml` ( or .. )
> * [x]  **fix** jobs badge workflow name `hera_build_and_publish.yaml` -> `cicd.yaml` ( or .. ) in `README.md`
> * [x]  **remove** the `|| true` from `publish` job `if:` ( added so `publish` runs as part of this PR for wf validation )
> * [ ]  **enable** upload to `pypi-test` for the purpose of validating entire `publish` job ( see `publish` job failure )
> * [ ]  **replace** upload to `pypi-test` with upload to `pypi` in `publish` job
> * [ ]  **setup**  [codecov](https://github.com/apps/codecov) **if** we want a `coverage` badge to be introduced in this PR ( or propose a different solution )
> * [ ]  **question**  do you want to add `windows-latest` to the `os` matrix ?
> 
> Thanks,
> 
> Harel

Hi @flaviuvadan - 

Handing this over to you as all remaining tasks require maintainer permissions, access to secrets etc. 

As for the merge process to `main`, here are  my 3 cents -- 

Integrate **as is** -> [ it will ] publish to `pypi-test` -> validate upload manually -> change to `pypi` official

Harel

flaviuvadan:
Thank you for the efforts @harelwa 🙂 I was on vacation for a week and am just getting back! I am going to merge this PR today and add the necessary secrets.

flaviuvadan (approved):

harelwa:
Hi @flaviuvadan 

I see the unfortunate failures. 

Following - 

https://stackoverflow.com/questions/59451069/binary-wheel-cant-be-uploaded-on-pypi-using-twine

https://peps.python.org/pep-0513/#rationale

I see `auditwheel` can fix that.

I'll give it a try

flaviuvadan:
> Hi @flaviuvadan
> 
> I see the unfortunate failures.
> 
> Following -
> 
> https://stackoverflow.com/questions/59451069/binary-wheel-cant-be-uploaded-on-pypi-using-twine
> 
> https://peps.python.org/pep-0513/#rationale
> 
> I see `auditwheel` can fix that.
> 
> I'll give it a try

Thank you for being on top of this, @harelwa! I started reading about `manylinux` and [cibuildwheel](https://cibuildwheel.readthedocs.io/en/stable/setup/#github-actions) myself to diagnose the issue and think about a fix.

harelwa:
> > Hi @flaviuvadan
> > I see the unfortunate failures.
> > Following -
> > https://stackoverflow.com/questions/59451069/binary-wheel-cant-be-uploaded-on-pypi-using-twine
> > https://peps.python.org/pep-0513/#rationale
> > I see `auditwheel` can fix that.
> > I'll give it a try
> 
> Thank you for being on top of this, @harelwa! I started reading about `manylinux` and [cibuildwheel](https://cibuildwheel.readthedocs.io/en/stable/setup/#github-actions) myself to diagnose the issue and think about a fix.

Hi @flaviuvadan - 

The immediate fix with minimal changes is to **not** publish the _platform_ wheels. 

You can achieve this by removing / commenting out the upload step in `test` job.

Then only `none-any` will be published ( as did thus far ).

As for repairs etc. -- perhaps for later -- 

When trying to `repair` with `auditwheel` I get - 

```bash
>$ auditwheel repair ./tox/dist/hera_workflows-2.2.1-cp39-cp39-linux_x86_64.whl 
INFO:auditwheel.main_repair:Repairing hera_workflows-2.2.1-cp39-cp39-linux_x86_64.whl
INFO:auditwheel.main_repair:This does not look like a platform wheel
```

as for the discussion in https://github.com/pypa/auditwheel/issues/247 -- a bit odd to me as `libc` et. al. issues are a matter of python it self, not only compiled code in the package...

From what I can see `cibuildwheel` can also be used; note it requires `docker`

What do you think ?

flaviuvadan:
> You can achieve this by removing / commenting out the upload step in test job -

I agree with the proposed approach to comment out the upload steps for now! Let's do that, and then investigate how we can introduce `cibuildwheel` as a package to build wheels with.

harelwa:
> > You can achieve this by removing / commenting out the upload step in test job -
> 
> I agree with the proposed approach to comment out the upload steps for now! Let's do that, and then investigate how we can introduce `cibuildwheel` as a package to build wheels with.

great. will you take care of it or you'd rather me open a PR ?

JacobHayes:
Hi! Does `hera` have any compiled extensions in the wheel? If not, I think the stock `py3-none-any` wheel will cover all platforms automatically and we can skip any `cibuildwheel` stuff. Quoting from the [setuptools docs](https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#platform-wheels):

> Platform Wheels are wheels that are specific to a certain platform like Linux, macOS, or Windows, usually due to containing compiled extensions.
>
> The wheel package will detect that the code is not pure Python, and build a wheel that’s named such that it’s only usable on the platform that it was built on. For details on the naming of wheel files, see [PEP 425](https://www.python.org/dev/peps/pep-0425).

And testing a bit, everything looks good:
```sh
$ sh -c 'pip3 install wheel && pip3 install hera-workflows' | grep hera_workflows # Apple M1
  Downloading hera_workflows-2.2.1-py3-none-any.whl (41 kB)
$ docker run --rm -it --platform=linux/x86-64 python:3.7 sh -c 'pip3 install wheel && pip3 install hera-workflows' | grep hera_workflows
  Downloading hera_workflows-2.2.1-py3-none-any.whl (41 kB)
$ docker run --rm -it --platform=linux/x86-64 python:3.9 sh -c 'pip3 install wheel && pip3 install hera-workflows' | grep hera_workflows
  Downloading hera_workflows-2.2.1-py3-none-any.whl (41 kB)
$ docker run --rm -it --platform=linux/amd64 python:3.7 sh -c 'pip3 install wheel && pip3 install hera-workflows' | grep hera_workflows
  Downloading hera_workflows-2.2.1-py3-none-any.whl (41 kB)
$ docker run --rm -it --platform=linux/amd64 python:3.9 sh -c 'pip3 install wheel && pip3 install hera-workflows' | grep hera_workflows
  Downloading hera_workflows-2.2.1-py3-none-any.whl (41 kB)
```

--

I think this might also explain what @harelwa saw with:
```sh
>$ auditwheel repair ./tox/dist/hera_workflows-2.2.1-cp39-cp39-linux_x86_64.whl 
INFO:auditwheel.main_repair:Repairing hera_workflows-2.2.1-cp39-cp39-linux_x86_64.whl
INFO:auditwheel.main_repair:This does not look like a platform wheel
```
-> [there is no compiled code](https://github.com/pypa/auditwheel/issues/247#issuecomment-678790316)?

harelwa:
> Hi, apologies for jumping in! Does `hera` have any compiled extensions in the wheel? If not, I think the stock `py3-none-any` wheel will cover all platforms automatically and we can skip any `cibuildwheel` stuff. Quoting from the [setuptools docs](https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#platform-wheels):
> 
> > Platform Wheels are wheels that are specific to a certain platform like Linux, macOS, or Windows, usually due to containing compiled extensions.
> > The wheel package will detect that the code is not pure Python, and build a wheel that’s named such that it’s only usable on the platform that it was built on. For details on the naming of wheel files, see [PEP 425](https://www.python.org/dev/peps/pep-0425).
> 
> And testing a bit, everything looks good:
> 
> ```shell
> $ sh -c 'pip3 install wheel && pip3 install hera-workflows' | grep hera_workflows # Apple M1
>   Downloading hera_workflows-2.2.1-py3-none-any.whl (41 kB)
> $ docker run --rm -it --platform=linux/x86-64 python:3.7 sh -c 'pip3 install wheel && pip3 install hera-workflows' | grep hera_workflows
>   Downloading hera_workflows-2.2.1-py3-none-any.whl (41 kB)
> $ docker run --rm -it --platform=linux/x86-64 python:3.9 sh -c 'pip3 install wheel && pip3 install hera-workflows' | grep hera_workflows
>   Downloading hera_workflows-2.2.1-py3-none-any.whl (41 kB)
> $ docker run --rm -it --platform=linux/amd64 python:3.7 sh -c 'pip3 install wheel && pip3 install hera-workflows' | grep hera_workflows
>   Downloading hera_workflows-2.2.1-py3-none-any.whl (41 kB)
> $ docker run --rm -it --platform=linux/amd64 python:3.9 sh -c 'pip3 install wheel && pip3 install hera-workflows' | grep hera_workflows
>   Downloading hera_workflows-2.2.1-py3-none-any.whl (41 kB)
> ```
> 
> --
> 
> I think this might also explain what @harelwa saw with:
> 
> ```shell
> >$ auditwheel repair ./tox/dist/hera_workflows-2.2.1-cp39-cp39-linux_x86_64.whl 
> INFO:auditwheel.main_repair:Repairing hera_workflows-2.2.1-cp39-cp39-linux_x86_64.whl
> INFO:auditwheel.main_repair:This does not look like a platform wheel
> ```
> 
> -> [there is no compiled code](https://github.com/pypa/auditwheel/issues/247#issuecomment-678790316)?

Thank you @JacobHayes 

Yes. I agree. I think we can drop the platform wheels all together.

PR #104, opened 2022-03-17
Support env_from option for task

kitagry:
Hi maintainers, thank you for the great library.

I tried some task, but this library can't specify `env_from` option. So, I added `env_from`.

Thank you.


CHANGELOG.md
```diff
@@ -33,7 +33,8 @@ The general format is:
 
 ### Added
 
-- ability to set the `access_modes` for dynamically privisioned volumes 
+- ability to set the `access_modes` for dynamically privisioned volumes
+- support for `env_from` option
```
flaviuvadan, line 37: Hera follows semantic versioning so, for a new feature, I suggest we bump the minor version, which means adding a new entry in the change log! So, I think the entry `- support for `env_from` option` should move to a block that looks like: 
```
# 2.4.0 - DATE (27/03/2022)

### Added

- support for `env_from` option
```

Happy to provide more information about semantic versioning! There are some more details in the [contribution guide](https://github.com/argoproj-labs/hera-workflows/blob/main/CONTRIBUTING.md)!

flaviuvadan (commented):
Since a new version should be published with the new features, I suggest this PR bumps the VERSION specified in the VERSION file as well! 🙂

CHANGELOG.md
kitagry: Thank you for the review!

I fixed `CHANGELOG.md` and upgrade `VERSION`

kitagry:
Thank you!

PR #110, opened 2022-03-27
Remove artifact uploads

flaviuvadan:
[Recent CICD improvements](https://github.com/argoproj-labs/hera-workflows/pull/101) resulted in upload failures because the built wheels do not use a `manylinux` type. @harelwa [suggested](https://github.com/argoproj-labs/hera-workflows/pull/101#issuecomment-1079996290) commenting out the upload part for now until we figure out how to create the necessary wheel types.

### DIFF
--- a/.github/workflows/cicd.yaml
+++ b/.github/workflows/cicd.yaml
@@ -100,7 +100,7 @@ jobs:
 
       - name: twine check
         run: |
-          twine check .tox/dist/*
+          twine check dist/*
 
       - name: upload build artifacts
         uses: actions/upload-artifact@v2
--- a/.github/workflows/cicd.yaml
+++ b/.github/workflows/cicd.yaml
@@ -96,7 +96,7 @@ jobs:
 
       - name: list artifacts
         run: |
-          ls -l .tox/dist
+          ls -l dist
 
       - name: twine check
         run: |
--- a/.github/workflows/cicd.yaml
+++ b/.github/workflows/cicd.yaml
@@ -57,11 +57,13 @@ jobs:
         run: |
           twine check .tox/dist/*
 
-      - name: upload build artifacts
-        uses: actions/upload-artifact@v2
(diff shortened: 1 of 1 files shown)


harelwa:
it might have been simpler to leave all paths as they were :)

harelwa:
> it might have been simpler to leave all paths as they were :)

although misleading. `+1` for refactor.

.github/workflows/cicd.yaml
```diff
@@ -57,11 +57,13 @@ jobs:
         run: |
           twine check .tox/dist/*
 
-      - name: upload build artifacts
-        uses: actions/upload-artifact@v2
-        with:
-          name: pypi-build-arts
-          path: .tox/dist
+# TODO: wheels for linux_x86_64 cannot be uploaded. Rather, this needs manylinux distributions, which can be installed
```
harelwa, line 60:
```suggestion
# TODO: wheels for linux_x86_64 cannot be uploaded. Rather, this needs manylinux distributions, which can be built
```

.github/workflows/cicd.yaml
```diff
@@ -57,11 +57,13 @@ jobs:
...
+# TODO: via cibuildwheel
```
harelwa, line 61:
```suggestion
# TODO: using `cibuildwheel` in a separate step / job, and not as part of tox test task
```

.github/workflows/cicd.yaml
```diff
@@ -90,21 +92,21 @@ jobs:
 
       - name: build sdist and wheel ( no test )
         run: |
-          python -m build --sdist --wheel --outdir .tox/dist/
+          python -m build --sdist --wheel
```
harelwa, line 95: even though it's the `default`, perhaps better explicit ?

```suggestion
          python -m build --sdist --wheel --outdir dist
```

PR #125, opened 2022-04-07
Add support for field env vars

xiang9156:


src/hera/env.py
```diff
@@ -98,3 +99,28 @@ def argo_spec(self) -> EnvVar:
...
+    field_path: str
+    api_version: Optional[str] = 'v1'
+
+    @property
+    def argo_spec(self) -> EnvVar:
+        """Constructs and returns the Argo environment specification"""
+        return EnvVar(
+            name=self.name,
+            value_from=EnvVarSource(
+                field_ref=ObjectFieldSelector(field_path=self.field_path, api_version=self.api_version)
```
flaviuvadan, line 124: I wonder whether there's a way to help users access a specific field path. For instance, one can add a path for: 
* metadata.labels['<KEY>']
* metadata.annotations['<KEY>']
* spec.nodeName
* spec.serviceAccountName
* status.hostIP
* status.podIP
* status.podIPs

Do you think it would be worth adding an object that helps uses access those field paths? Something like:
```python
class FieldPath(str, Enum):
    node_name: str = "spec.nodeName"
    service_account_name: str = "spec.serviceAccountName"
    host_ip: str = "status.hostIP"
    ....
 
    def label(self, key: str) -> str:
        return f"metadata.labels['key']"

    def annotation(self, key: str) -> str:
        return f"metadata.annotations['key']"
```

flaviuvadan (commented):
This is awesome @xiang9156! Thank you for the contribution! I added a question 🙂

src/hera/env.py
xiang9156: thanks for the comment @flaviuvadan ! I think using a FieldPath class helps to minimize the risk of typo or any other mistakes made by developers, but would it incur extra code maintain effort? e.g. if there's new field added by k8s in future versions, the contributor has to make update of this class before referencing the new field. Using a string type may gives extra flexibility for developers, what do you think?

flaviuvadan: Now that you explained it like this I had to rethink the perspective I had.. Going back to the original mission of Hera: Hera is designed to make Argo workflow construction and submission easy for everyone. It is primarily oriented to usage by people who may not be very familiar with K8S and Argo in the first place. However, I imagine usage of field paths from K8S is, really, a power user feature. Someone who wants to use something like this likely already know about the field paths, and know how to use them, set them up, etc. So, I agree with you - let's leave it as a dictionary for now at least 🙂

flaviuvadan (approved):

flaviuvadan:
Thank you for the contribution @xiang9156! 🚀

src/hera/env.py
xiang9156: Thank you Flaviu! Hera is an awesome tool! We'll keep contributing to it!!

PR #126, opened 2022-04-07
prevent json.loads error when the json string contains single quote

bnuzhanyu:


flaviuvadan:
Thank you for the submission @bnuzhanyu! 🙂 This change actually has the potential to break a lot of things so I want to be super cautious with the deployment and testing of this. Any chance you can create a Python example with a use case you had in which the `json.loads` failed? String serialization is a bit challenging in Python, and it's very finicky to get right, especially in a remote setting, as it is with Argo, which performs a `json.loads` operation.

bnuzhanyu:
@flaviuvadan  
Thanks for the reply. Here is my example.
```python
from hera.input import InputFrom
from hera.task import Task
from hera.workflow import Workflow
from hera.workflow_service import WorkflowService

def context_config():
    import json, sys
    day = '2022-04-01'
    job_config = {
        "sql": f"select * from table where day='{day}'"
    }
    json.dump([{'context' : job_config}], sys.stdout)
    
def sql(context):
    print(context['sql'])

def start():
    ws = WorkflowService('localhost:xxx', verify_ssl=False, token=get_sa_token(service_account="default", namespace="argo"))
    w = Workflow('cv-train-flow', ws, namespace='argo')
    tasks = []
    ctx_task_name = 'ctx-gen'
    ctx_gen = Task(ctx_task_name, context_config)
    tasks.append(ctx_gen)
    sql_task = Task('sql', sql, input_from=InputFrom(name=ctx_task_name, parameters=['context']))
    tasks.append(sql_task)

    ctx_gen.next(sql_task)
    w.add_tasks(*tasks)
    w.create()

start()

```

And got error:
```
  File "/argo/staging/script", line 2
    context = json.loads('{"sql":"select * from table where day='2022-04-01'"}')
                                                                    ^
SyntaxError: invalid syntax
Error: exit status 1
```
`json.dump / dumps` will change \ to \\, and " to \" , 
no matter how many \ added , it won't help.
like  
```python
f"select * from table where day=\'{day}\'"
f"select * from table where day=\\'{day}\\'"
f"select * from table where day=\\\'{day}\\\'"
f"select * from table where day=\\\\'{day}\\\\'"
f'select * from table where day="{day}"'
```
all of them got failed.

flaviuvadan:
@bnuzhanyu I did some testing against a cluster and everything looks good to me 🙂 Thank you for catching this error and submitting the PR. Any chance you know exactly why the serialization/deserialization fails in this case? I am aware that `json.dumps` will add the escape character `\` but I wonder what's the reasoning? Is it because of the way strings are encoded as strings and wrapper with other quotes, which necessitates escaping the internal quotes of the dumped string?

flaviuvadan (commented):
Need to bump the VERSION and add a CHANGELOG entry so a new version is published and the change is captured. I think it should be 2.9.1.

bnuzhanyu:
> @bnuzhanyu I did some testing against a cluster and everything looks good to me 🙂 Thank you for catching this error and submitting the PR. I will approve and merge after a question, if you do know the answer.
> 
> Any chance you know exactly why the serialization/deserialization fails in this case? I am aware that `json.dumps` will add the escape character `\` but I wonder what's the reasoning? Is it because of the way strings are encoded as strings and wrapper with other quotes, which necessitates escaping the internal quotes of the dumped string?

json.dumps will not add the escape for `'` ,  and the  produced python code will mistake for sth like `'{"x":"a'a"}'`. Also, my change will fail if the string has `'''`, I think the best way is to replace `'` to `\'` ahead, cause `'` would only appear in a key or a string value inside a jsonstring.

bnuzhanyu:
> Need to bump the VERSION and add a CHANGELOG entry so a new version is published and the change is captured. I think it should be 2.9.1.

Would this done by me or you guys will do it ?

flaviuvadan:
> > Need to bump the VERSION and add a CHANGELOG entry so a new version is published and the change is captured. I think it should be 2.9.1.
> 
> Would this done by me or you guys will do it ?

By you! 🙂 you can certainly skip this step but if we don't update the version in this PR then a new version of Hera will not be published immediately for you to access the bug fix.

flaviuvadan (approved):
Thank you @bnuzhanyu! 🚀

PR #127, opened 2022-04-07
Support TTLStrategy

linyao-reevo:
Allow to explicitly declare the TTL of workflow, at this moment the completed pods remained indefinitely.

@flaviuvadan 

https://argoproj.github.io/argo-workflows/fields/#ttlstrategy

### DIFF
--- a/src/hera/ttl_strategy.py
+++ b/src/hera/ttl_strategy.py
@@ -1,4 +1,5 @@
 from pydantic import BaseModel
+from typing import Optional
 from argo_workflows.models import (
     IoArgoprojWorkflowV1alpha1TTLStrategy
 )


src/hera/workflow.py
```diff
@@ -71,6 +72,7 @@ def __init__(
         security_context: Optional[WorkflowSecurityContext] = None,
         image_pull_secrets: Optional[List[str]] = None,
         workflow_template_ref: Optional[str] = None,
+        ttl_strategy: Optional[Dict[str, int]] = None,
```
flaviuvadan, line 75: What do you think about making a custom Hera object representation of the TTL strategy? I think that will allow Hera to construct the TTL strategy using something like a `get_argo_spec`, similar to [Env](https://github.com/argoproj-labs/hera-workflows/blob/main/src/hera/env.py), which is a nice abstraction that supports creating custom logic that is aimed at helping the user have an easier time creating workflows!

src/hera/workflow.py
```diff
@@ -83,6 +85,16 @@ def __init__(
         self.image_pull_secrets = image_pull_secrets
         self.workflow_template_ref = workflow_template_ref
 
+        if ttl_strategy:
```
flaviuvadan, line 88: Maybe this should be set using `setattr`? 🤔 Thoughts? Similar to [line 107](https://github.com/argoproj-labs/hera-workflows/pull/127/files#diff-74e2ee6a97c31386ee31e5ff0f31cca730e28e83f429fb64b421bc14b9197db1L107).

flaviuvadan:
Thank you for the PR @linyaoli! This is going to be a very, very nice addition 🚀 I left some questions and suggestions.

linyaoli:
Hi @flaviuvadan, I'm not sure if I fully grasp your idea regarding the abstraction here, but I've added a `TTLStrategy` Model for better representation. Also `CronWorkflow` is supported now.  Thanks! (as I'm not a daily Python user, if there's any issue, pls lemme know)

src/hera/ttl_strategy.py
```diff
@@ -0,0 +1,37 @@
...
+    seconds_after_failure: int
+        The number of seconds to live after failure.
+    seconds_after_success: int
+        The number of seconds to live after success.
+
+    """
+
+    seconds_after_completion: int
+    seconds_after_failure: int
+    seconds_after_success: int
```
flaviuvadan, line 28:
```suggestion
    seconds_after_completion: Optional[int]
    seconds_after_failure: Optional[int]
    seconds_after_success: Optional[int]
```

src/hera/ttl_strategy.py
```diff
@@ -0,0 +1,37 @@
+from pydantic import BaseModel
```
flaviuvadan, line 1:
```suggestion
from pydantic import BaseModel
from typing import Optional
```

flaviuvadan:
@ linyaoli what you added was almost exactly what I had in mind 🚀  I left some small suggestions for making the times optional 🙂  Last few things before we merge:
* add the TTL strategy to [workflow template](https://github.com/argoproj-labs/hera-workflows/blob/main/src/hera/workflow_template.py#L23) as well
* bump the version to 2.9.0 [here](https://github.com/argoproj-labs/hera-workflows/blob/main/pyproject.toml#L3)
* add an entry for 2.9.0 to the [CHANGELOG](https://github.com/argoproj-labs/hera-workflows/blob/main/CHANGELOG.md)

flaviuvadan:
Reminder to sign the contributor license agreement! 🙂

linyaoli:
Hi @flaviuvadan, please review https://github.com/argoproj-labs/hera-workflows/pull/132, sorry had to close this one (got issue with my fork and had to start over)

Issue #128, opened 2022-04-07
[BUG] Wrong order of next() in `Task.when()`, `Task.on_success`, `Task.on_failure`


RunningIkkyu:
Shall we change this line to  `return self.next(other)` ?  
https://github.com/argoproj-labs/hera-workflows/blob/ebd7ac27ce26328f1de4bc669b4d0c710f11dd5e/src/hera/task.py#L375

So as the return statements in `when()`, `on_failure()`.

flaviuvadan:
Addressed in https://github.com/argoproj-labs/hera-workflows/pull/129

PR #129, opened 2022-04-07
Fix wrong dependencies for on_success(), on_failure(), on_error() functions of Task

xiang9156:
While I was using on_success, on_failure and on_error functions inside Task, I found that the dependencies looked incorrect.
According to the description of the functions, the correct logic should be like this:
```
If A.on_success(B), it means that A >> B and execute B when A.status == success
if A.on_failure(B), it means that A >> B and execute B when A.status == failure
If A.on_error(B), it means that A >> B and execute B when A.status == error
```
However, according to my understanding of the existing code, the logic looks like this:
```
If A.on_success(B), B >> A and execute A when B.status == success
If A.on_failure(B),  B >> A and execute A when B.status == failure
If A.on_error(B), B >> A and execute A when B.status == error
```
@flaviuvadan  Correct me if I'm wrong, hopefully this PR is the correct fix. :)


CHANGELOG.md
```diff
@@ -29,6 +29,12 @@ The general format is:
 
 ```
 
+# 2.9.0 - DATE (04/08/2022)
```
flaviuvadan, line 32: Can this be a PATCH since it's a bug fix? 🙂

VERSION
```diff
@@ -1 +1 @@
-2.8.0
+2.9.0
```
flaviuvadan, line 1:
```suggestion
2.8.1
```

flaviuvadan:
Thank you so much for catching this @xiang9156! 🚀  It was recently reported in https://github.com/argoproj-labs/hera-workflows/issues/128 by @RunningIkkyu 🙂 I left a minor request for the version!

VERSION
xiang9156: updated :)

CHANGELOG.md
xiang9156: Sure thing! Thanks for the comment :)

xiang9156: done

PR #130, opened 2022-04-08
Use separate poetry build and twine upload step 

flaviuvadan:
The package publish failed in [this build](https://github.com/argoproj-labs/hera-workflows/runs/5885199764?check_suite_focus=true) because the project aimed to publish Hera at https://pypi.org/project/hera rather than https://pypi.org/project/hera-workflows. This PR changes the name of the project to fix that. Also, the publication step will fail for versions that do not change the project version. To get around that, this adds a skip existing flag to the publication process. Note that poetry does not support a skip existing flag so the command was separated out into a poetry build step and a twine upload step.

### DIFF
--- a/.github/workflows/cicd.yaml
+++ b/.github/workflows/cicd.yaml
@@ -139,7 +139,8 @@ jobs:
 
       - name: Build and publish
         env:
-          TWINE_USERNAME: __token__
-          TWINE_PASSWORD: ${{ secrets.PYPI_API_TOKEN }}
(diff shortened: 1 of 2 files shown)


flaviuvadan:
@hirosassa this fixed the publication step https://github.com/argoproj-labs/hera-workflows/runs/5885768306?check_suite_focus=true

hirosassa:
@flaviuvadan Thanks! 👍

PR #134, opened 2022-04-10
CronWorkflow update command

kitagry:
resolve #133

PR #140, opened 2022-04-20
Fix CronWorkflow update command

kitagry:
relate https://github.com/argoproj-labs/hera-workflows/pull/134

I'm sorry :bow: I noticed some bugs and fixed it.

### DIFF
--- a/src/hera/cron_workflow.py
+++ b/src/hera/cron_workflow.py
@@ -165,11 +165,15 @@ class CronWorkflow:
             namespace = self.namespace
         return self.service.create(self.workflow, namespace)
 
-    def update(self, namespace: Optional[str] = None) -> IoArgoprojWorkflowV1alpha1CronWorkflow:
+    def update(
+        self, name: Optional[str] = None, namespace: Optional[str] = None
+    ) -> IoArgoprojWorkflowV1alpha1CronWorkflow:
         """Updates the cron workflow in the server"""
         if namespace is None:
             namespace = self.namespace
-        return self.service.update(self.workflow, namespace)
+        if name is None:
+            name = self.name
+        return self.service.update(self.workflow, name, namespace)
 
     def suspend(self, name: Optional[str] = None, namespace: Optional[str] = None) -> Tuple[object, int, dict]:
         """Suspends the cron workflow"""
--- a/src/hera/cron_workflow_service.py
+++ b/src/hera/cron_workflow_service.py
@@ -74,7 +74,10 @@ class CronWorkflowService:
         )
 
     def update(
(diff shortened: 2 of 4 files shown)


kitagry:
And, I got following error. I think argo_workflows package cause this. Do you know about it?

```
Reason: Bad Request
HTTP response headers: HTTPHeaderDict({'Content-Type': 'application/json', 'Date': 'Wed, 20 Apr 2022 07:53
:42 GMT', 'Via': '1.1 google', 'Transfer-Encoding': 'chunked', 'Alt-Svc': 'clear'})
HTTP response body: {"code":3,"message":"cronworkflows.argoproj.io \"argo-batch\" is invalid: metadata.r
esourceVersion: Invalid value: 0x0: must be specified for an update"}
```

flaviuvadan:
No worries @kitagry 🙂  these things happen. Any chance the PR can include a test for the two logical branches? The test can mock the service and make it raise an error when name is not passed, an assertions that the service.update is called with expected namespace and name. I believe the addition of the tests will make the codecov check pass!

flaviuvadan:
> And, I got following error. I think argo_workflows package cause this. Do you know about it?

This type of error seems to have been reported in ArgoCD as well 🤔  See [this issue](https://github.com/argoproj/argo-cd/issues/3657) for instance. Seems like the update might benefit from the following workflow:
- get the cron workflow first
- update spec, schedule, etc
- keep existing metadata? 

I will attempt to test this today.

flaviuvadan:
@kitagry any chance you tried the update request with a `metadata.resourceVersion`? I believe [this](https://github.com/argoproj-labs/hera-workflows/blob/main/src/hera/cron_workflow.py#L136) is the object and it can be accessed via `wf.metadata` 🤔

kitagry:
I haven't try it. I'll try this tomorrow, Thank you!

kitagry:
@flaviuvadan Thank you for your advice! I confirmed the operation.

flaviuvadan:
@kitagry, this looks awesome! [This PR](https://github.com/argoproj-labs/hera-workflows/pull/143) delivers version 3.1.0. Any chance you are open to merging main, after we merger the 3.1.0 into main, into this branch and update the project version to 3.1.1 (it's a bug fix/patch) and add an entry to the changelog? 🙂

flaviuvadan:
Actually, I am going to merge this first, then merge the [other PR](https://github.com/argoproj-labs/hera-workflows/pull/143) so we deliver a single version rather than 2. I will upgrade the CHANGELOG afterwards!

flaviuvadan (approved):

PR #145, opened 2022-05-02
fix: template_ref type in comment

szdr:
The type of template_ref is Optional[TemplateRef].


flaviuvadan (approved):
Thank you for the catch @szdr! 🚀

PR #147, opened 2022-05-03
Support VolumeClaimGC in Workflow

szdr:
Support VolumeClaimGC
https://github.com/argoproj/argo-workflows/blob/master/sdks/python/client/docs/IoArgoprojWorkflowV1alpha1VolumeClaimGC.md

### DIFF
--- a/src/hera/volume_claim_gc.py
+++ b/src/hera/volume_claim_gc.py
@@ -1,8 +1,5 @@
 from enum import Enum
 
-from argo_workflows.models import IoArgoprojWorkflowV1alpha1VolumeClaimGC
-from pydantic import BaseModel
-
 
 class VolumeClaimGCStrategy(Enum):
     """A representation of the strategy to use when deleting volumes from completed workflows
@@ -19,18 +16,3 @@ class VolumeClaimGCStrategy(Enum):
     OnWorkflowSuccess = "OnWorkflowSuccess"
     """delete volume claim when workflow is successful
     """
-
-
-class VolumeClaimGC(BaseModel):
-    """VolumeClaimGC describes how to delete volumes from completed Workflows
-
-    Notes
-    -----
-        See sdk doc (IoArgoprojWorkflowV1alpha1VolumeClaimGC)
-    """
-
-    strategy: VolumeClaimGCStrategy
-
-    @property
-    def argo_volume_claim_gc(self) -> IoArgoprojWorkflowV1alpha1VolumeClaimGC:
-        return IoArgoprojWorkflowV1alpha1VolumeClaimGC(strategy=self.strategy.value)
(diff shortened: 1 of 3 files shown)


src/hera/volume_claim_gc.py
```diff
@@ -0,0 +1,36 @@
...
+    OnWorkflowCompletion = "OnWorkflowCompletion"
+    """delete volume claim when workflow is completed
+    """
+
+    OnWorkflowSuccess = "OnWorkflowSuccess"
+    """delete volume claim when workflow is successful
+    """
+
+
+class VolumeClaimGC(BaseModel):
```
flaviuvadan, line 24: The [docs](https://argoproj.github.io/argo-workflows/fields/#volumeclaimgc) for volume claim GC suggest there's a single field at the moment, which is `strategy`. What do you think about using the `VolumeClaimGCStrategy` enum directly in a workflow `__init__` rather than wrapping it with a `VolumeClaimGC` `BaseModel`? I think `ImagePullPolicy` is similar to this - we use the enum directly. My decision boundary between using a base model and not using one is generally asking the question: "does this have more than 1 field and are there opportunities to construct logic on the fields?" If the answer is no, which I think is the case for image pull policy, GC, etc, then I suggest we don't use a `BaseModel`. Thoughts, @szdr?

szdr: Thanks for your comment, I agree with the idea of using VolumeClaimGCStrategy directly in __init__ of workflow.

src/hera/workflow.py
```diff
@@ -70,6 +74,7 @@ def __init__(
         image_pull_secrets: Optional[List[str]] = None,
         workflow_template_ref: Optional[str] = None,
         ttl_strategy: Optional[TTLStrategy] = None,
+        volume_claim_gc_strategy: Optional[VolumeClaimGCStrategy] = VolumeClaimGCStrategy.OnWorkflowCompletion,
```
flaviuvadan, line 77: I am not sure this is a good assumption 🤔  maybe some users may want to keep the volumes around post workflow completion. The same logic was applied when the other optional parameters were set to `None`. There's no way to know whether the user does want a TTL strategy, so Hera offers the option to add it but makes no assumptions about it, for now at least. Any chance we can set this to `None` as well? I think we are ready to merge right after that into 3.2.0 🚀

szdr: You're right! I would leave the default value as None.

szdr: https://github.com/argoproj-labs/hera-workflows/pull/147/commits/ed486defdddd891d13b69244ca3e445c49e05161

flaviuvadan (approved):
🚀

PR #154, opened 2022-05-07
Add task template mechanism 

kromanow94:
## Description

This MR provides additional funcitonality for templating tasks with a class `TaskTemplate`.

## Example
```
task_template = TaskTemplate('myTemplate', say,
    [{
        'message': "default value for param 'message'",
        "other_message": {"default": {"value": {"for": "other_message"}}},
    }],
)
a = task_template.task("taskWithDefaultValues")
b = task_template.task("taskWithNonDefaultValues", [{"other_message": "hello again"}])

```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,8 @@ The general format is:
 ### Added
 
 - support `cluster_scope` in template reference
+- add implementation for `TaskTemplates` - this enables specifying task details with a template once and creating task
+  based on that template instead of redefining every time
 
 # 3.1.0 - DATE (28/04/2022)
 
--- a/README.md
+++ b/README.md
@@ -126,6 +126,7 @@ Currently, Hera is centered around two core concepts. These concepts are also us
 consistent with:
 
 - `Task` - the object that holds the Python function for remote execution/the atomic unit of execution;
+- `TaskTemplate` - allows specifying task once and reusing with different arguments;
 - `Workflow` - the higher level representation of a collection of tasks.
 
 # Examples
--- a/examples/diamond_task_template.py
+++ b/examples/diamond_task_template.py
@@ -55,13 +55,7 @@ b = task_template.task('B-task', [{'message': 'This is task B!'}])
 c = task_template.task('C-task', [{'other_message': 'This is task C!'}])
 d = task_template.task(
     'D-task',
-    [
-        {
-            'message': {
-                "dicted": "message with non-alpanumeric characters: !@#$%^&*()_-+=;|<.>/?"
-            }
-        }
-    ],
+    [{'message': {"dicted": "message with non-alpanumeric characters: !@#$%^&*()_-+=;|<.>/?"}}],
 )
 
 a.next(b)  # a >> b
--- a/examples/diamond_task_template.py
+++ b/examples/diamond_task_template.py
@@ -1,0 +1,73 @@
+"""
+This example showcases the classic diamond workflow that is used as an example by Argo documentation and
+other libraries but with the use of TaskTemplate.
+
+TaskTemplate provides the same functionality as Task with the exception that one TaskTemplate could generate
+other tasks that would have their definition stated once and just invoked with arguments.
+
+This makes the workflow size smaller by reusing task templates in tasks. Also, it improves consistency and
+allows specifying default values for tasks.
+"""
+from hera.task_template import TaskTemplate
+from hera.workflow import Workflow
+from hera.workflow_service import WorkflowService
+
+
+def say(message: str, other_message: str):
+    print(f"message: {message}")
+    print(f"other_message: {other_message}")
+
+
+# TODO: replace the domain and token with your own
+ws = WorkflowService(host='https://my-argo-server.com', token='my-auth-token')
+w = Workflow('diamond-task-template', ws)
+
+# NOTE: task_template is specified once and invocations of task_template.task(...)
+# provide only task name and args.
+task_template = TaskTemplate(
+    'A',
+    say,
+    [
+        {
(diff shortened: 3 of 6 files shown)


kromanow94:
@flaviuvadan Hello, can we run the tests again? Coverage for `workflow_editors.py` should better now.

kromanow94:
@flaviuvadan I'm glad to see that this MR is now fully passing all the checks.

I think this would be nice to include in 3.2.0 which you've opened a MR for.

I was wondering about next steps and would like your opinion if this is what's preferred:
* rebase this MR with branch `hera/3.2.0`
* change target branch to `hera/3.2.0`
* write changelog in that branch

I'm open to suggestions.

kromanow94:
@flaviuvadan I've rebased into `hera/3.2.0` and added info to CHANGELOG.md.

Also, I've noticed that the current codecov comment is not related to the latest commit. Running coverage locally with tox resulted in higher values.

Is there anything else I can do?

flaviuvadan:
Thank you so much for the PR @kromanow94! 🙂 Following is an explanation of my perspective. 

First, I merged the 3.2.0 branch and really did not expect it to close your PR automatically! I apologize for that!

I wanted to leave a high level comment on this idea; I think I understand the value of it! The point is to make it easier to create tasks from a given task template, conceptually. Users would get to define a task once and keep performing `.task` to keep creating tasks that are copies of the original one but with modified parameters. The idea here is to save the users from having to perform something like `task.argo_task = task.get_argo_task()` if they modify something on a task template. So, I definitely acknowledge the problem. However, I think it approaches the problem from the wrong angle. Here are the problems I see with the approach, which I will follow up with alternatives:
* Hera aims to facilitate access to Argo Workflows by providing an easy interface for scheduling workflows. Part of this mission is to maintain consistency with the language of Argo, so that Hera's user base has an easy time maintaining a 1:1 mapping between Argo concepts and Hera concepts. Hera aims to maintain this 1:1 association but make the concepts easily accessible, by doing some work for the user. The idea of a `TaskTemplate` does not exist in Argo, yet at least. Only `WorkflowTemplate`s are, currently, available
* I think there is a problem with the internal design of Hera, which motivated the approach presented by this PR. I think the `get_argo_spec` call should be, perhaps, performed by the workflow rather than the task. This way users can make copies of a task, modify fields/parameters, and only have the task "construct" when it passes through the workflow. This open the opportunity to reduce the "work" that a task does, and make it provide builder hooks for the workflow so the workflow constructs the argo spec of the task, and task becomes lighter as a consequence of this complexity reduction. Exposing something like a (deep) `copy()` on the task is perhaps an easier way, conceptually, to make a "template" - a user makes a (deep) copy of a task. modifies fields as desired, then when the task is injected into a workflow, then and only then, it is assembled into an Argo task representation.

Again, I must apologize for closing the PR! I really did not expect this branch to be closed! It makes sense now since the head branch was deleted.

kromanow94:
@flaviuvadan Thank you for your comment. No worries about closing this PR.

Although, I have other opinion on Argo not implementing the TaskTemplate: it's possible to create a template for a task inside a workflow and then reference it in other tasks. This is shown in the Argo examples:
* https://github.com/argoproj/argo-workflows/blob/master/examples/dag-diamond-steps.yaml#L52
* https://github.com/argoproj/argo-workflows/blob/master/examples/dag-diamond-steps.yaml#L62
* https://github.com/argoproj/argo-workflows/blob/master/examples/dag-diamond-steps.yaml#L67

In current hera implementation, every call to the `Task` class creates a template in `.spec.templates` and then calls it in task defined in the DAG template that is used as a workflow entrypoint. This task (that is defined as one of the tasks in the entrypoint DAG) references it with something like `spec.templates[#].tasks[#].template: A`.

What I had in mind for the value in that PR is that we could get rid of redefining the templates.

Imagine I want to have a 100 of tasks that are based on the same function. This function have 100 lines. This creates 100 templates. Each of these templates have exactly the same content in `source` key. At end, we use about 10 000 lines for the workflow definition only for the templates (not including parameters, name and other fields). If we were to use the task template, we would have only one template defined with a 100 lines and then referenced in 100 tasks defined at the DAG level. This brings considerate savings in terms of how big this workflow defined with hera can be.

Once you have a hundreds or thousands of tasks that each uses the same python function, it results in redefinition of the template for hundreds or thousands of tasks. It's not optimal and could create an issue when workflow size exceeds maximum object size that depends on ETCD.

kromanow94:
@flaviuvadan Thank you, much appreciated.

From what I've seen in the code - the task template generation is already handled for a `Task`. This PR I proposed contained _reusing_ into that, combined with testing if given template object is already present on the list of workflow templates. Still, I understand this may not be desired solution.

I offer my help for coding if required.

flaviuvadan:
@kromanow94 I've been thinking about this a bit more.. I still think the idea of saving templates is amazing but there are several cases in which resources, such as volumes, need their own dynamically provisioned volume IDs. I am wondering if you have any thoughts on how we could get around this aspect? The script template holds the resources associated with a pod running a task. If a user wants to provision an independent volume for a task the user can do something like 
```python
w = Workflow(...)
for i in range(5):
    t = Task(..., resources=Resources(volumes=[Volume('5Gi', '/mnt/vol')]))
    w.add_task(t)
```
The problem is that when `w` creates the task templates there will be independent volume IDs for each task, each corresponding to a single volume to attach. Maybe there's a balance to strike.. perhaps a task that has `func_params` set can use a shared template and tasks that have a single params do not? When `func_params` is passed Argo will create a task group rather than single tasks, and these will share resources, such as the same volume if a user specifies one. In this case the same template can be used I think. Thoughts?

PR #161, opened 2022-05-16
Feat workflow visualize

Matt9993:
This PR is about a new feature that is using the Graphviz library to visualize a Workflow or WorkflowTemplate.
The generated image that is a PDF by default can be used to better understand the execution order of the Tasks in a Workflow and/or can be used to put the generated image into a project documentation.


flaviuvadan:
This is awesome @Matt9993! 🚀 I will review this PR this weekend!

flaviuvadan:
@Matt9993 any chance you can upload a sample file visualization of the workflow? I think it would aid to the review of this PR!

Matt9993:
Sorry @flaviuvadan  for only solving this now but the coverage check was always messing up my PR. I hope now it looks good to you. If you have any suggestions please tell me and I`m happy to implement it.

flaviuvadan:
I looked through the documentation of the `graphviz` dependency and it seems that it may have a runtime dependency on the GraphViz engine [here](https://gitlab.com/graphviz/graphviz). Is that correct?

Matt9993:
Yeah I think you are right. Are you thinking about putting the import into the `tool.poetry.dev-dependencies` or maybe something separate just for the visualization functionality?

flaviuvadan:
My main concern with the extra requirement is that it will make Hera no longer a stand-alone package - users will have to install additional elements, such as the GraphViz rendering engine. Given that Hera aims to be simple and provide easy access to Argo Workflows I do not think this additional dependency fits the mission. If the dependency were only the PyPI `graphviz` package this would 100% be approved! Because of this limitation I will reject the PR. However, perhaps there might be different graph drawing approaches that can solve the visualization problem and use a PyPI package that can be installed without any secondary runtime dependencies, such as a rendering engine. That would be amazing contribution, such as this one! I am very thankful for the contribution @Matt9993 and I hope to see future ones as well 🙂

PR #168, opened 2022-05-26
Fix input_from behavior without func

### LINKED ISSUE #167: Cannot use input_from without func
I want use input_from without func. But, the following code doesn't work...

```python
Task(
    "test",
    command=['python'],
    args=['-c', 'print("{{inputs.parameters.value}}"'],
    input_from=InputFrom(name='generate', parameters=['value'])
)
```

Above code gets following error.

```
  File "/lib/python3.9/site-packages/hera/task.py", line 294, in __init__
    self.parameters = self.get_parameters()
  File "/lib/python3.9/site-packages/hera/task.py", line 701, in get_parameters
    args = set(inspect.getfullargspec(self.func).args).intersection(set(self.input_from.parameters))
  File "/lib/python3.9/inspect.py", line 1171, in getfullargspec
    raise TypeError('unsupported callable') from ex
TypeError: unsupported callable
```

kitagry:
fix #167

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,12 @@ The general format is:
 
 ```
 
+# 3.3.1 - DATE (26/05/2022)
+
+### Fixed
+
+- Fix Task raise error with input_from without func
+
 # 3.3.0 - DATE (23/05/2022)
 
 ### Added
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "hera-workflows"
-version = "3.3.0"
+version = "3.3.1"
 description="Hera is a Python framework for constructing and submitting Argo Workflows. The main goal of Hera is to make Argo Workflows more accessible by abstracting away some setup that is typically necessary for constructing Argo workflows."
 authors = ["Flaviu Vadan <flaviu.vadan@dynotx.com>"]
 license = "MIT"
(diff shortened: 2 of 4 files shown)


flaviuvadan:
Thank you for the issue report and quick fix @kitagry! 🙂 Gentle reminder to bump the version and document the fix in a changelog. 3.3.1, perhaps?

flaviuvadan:
@kitagry any chance you had the opportunity to test the example in the issue? Seems interesting to me to have a `python -c "print("{{input....}}")" work, actually, and I want to make sure we know it works as expected. I would be surprised as I expect Argo to only do replacements for variables in its own objects but not in used provided container scripts 🤔

flaviuvadan:
@kitagry I tested this myself really quick and it works just fine 🚀 I was wrong about the `{{...}}` as Argo replaced the `input.parameters` in the `-c` Python call as well. This is the script I ran successfully: 
```python
from hera import Task, Workflow, InputFrom, WorkflowService


def gen():
    import sys
    import json
    json.dump([{'value': i} for i in range(1)], sys.stdout)


ws = WorkflowService(...)
w = Workflow('fv-test', ws)
gen = Task('gen', gen)
con = Task(
    'con',
    command=['python'],
    args=['-c', 'print("{{inputs.parameters.value}}")'],
    input_from=InputFrom(name=gen.name, parameters=['value'])
)
gen >> con
w.add_tasks(gen, con)
w.create()
```

flaviuvadan (approved):

kitagry:
Oh, I'm sorry for late reply! Thank you for the test!!

PR #186, opened 2022-06-13
Add set_global_token_callback to handle refreshing expiring tokens

JacobHayes:
3.4.0 added support for setting a global token, making it possible to configure credentials once for easier Workflow(Service) instantiation later. :rocket: However, when working with expiring tokens in a long-running service, the caller may need to refresh the token with another call to `set_global_token`. Instead, this PR adds support for passing a callback to `set_global_token` that will be called with no parameters to fetch a token as needed.

The token returned by the callback will not be cached by Hera (we don't know the token's format / how to check for expiry), so callers may want to cache internally if fetching is "expensive".

### DIFF
--- a/src/hera/__init__.py
+++ b/src/hera/__init__.py
@@ -14,7 +14,12 @@ from hera.cron_workflow_service import CronWorkflowService
 from hera.env import ConfigMapEnvSpec, EnvSpec, FieldEnvSpec, SecretEnvSpec
 from hera.env_from import ConfigMapEnvFromSpec, SecretEnvFromSpec
 from hera.host_alias import HostAlias
-from hera.host_config import set_global_host, set_global_token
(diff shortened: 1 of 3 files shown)


JacobHayes:
tangent: I was thinking a bit about the possibility of a Hera config file (eg: `~/.config/hera.{ext}`) where we could set persistent defaults for the host and token/callback. This could remove the need for callers to ensure the `set_*` functions have been called before use (eg: in some common code or a Hera wrapper). For this callback, perhaps the config file could point to the importable name (eg: `corp.argo.get_token`) that Hera would import and `set_global_token_callback` upon loading the config.

If that'd be of interest, I'd be happy to take a look at adding!

flaviuvadan (approved):

PR #198, opened 2022-06-23
Support task exit hook

michaelvay:
Added exit hook for Task, which makes use of Argo's Lifecyclehook. A simple example was also added.

### DIFF
--- a/src/hera/task.py
+++ b/src/hera/task.py
@@ -129,8 +129,6 @@ class Task:
         Whether to continue task chain execution when this task fails.
     continue_on_error: bool = False
         Whether to continue task chain execution this task errors.
-    exit_hook: Optional['Task'] = None
-        On the completion (exit) of this Task the given Task is executed.
     template_ref: Optional[TemplateRef] = None
         A template name reference to use with this task. Note that this is prioritized over a new template creation
         for each task definition.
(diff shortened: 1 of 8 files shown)


src/hera/task.py
```diff
@@ -124,6 +129,8 @@ class Task:
         Whether to continue task chain execution when this task fails.
     continue_on_error: bool = False
         Whether to continue task chain execution this task errors.
+    exit_hook: Optional['Task'] = None
```
flaviuvadan, line 132: the `__init__` was kind of expected to become fairly rich. So far Hera avoided adding action-related parameters on the `__init__`. Rather, it favored adding methods on the `Task` that set a specific reaction on some result, such as `on_failure`. To avoid the proliferation of parameters that are not action related in the `__init__` I wonder what you think about adding an `on_exit(..., t: `Task`) -> 'Task'`? Similar to [on_success](https://github.com/argoproj-labs/hera-workflows/blob/main/src/hera/task.py#L302)!

flaviuvadan (commented):
first, I apologize for the late review @michaelvay! Second, super thankful for this contribution! it's very exciting 🚀 I left a comment that changes the implementation at a pretty fundamental level so let's have a discussion on that thread. Also, note that this will be released as part of the 3.6.0 release, along with your other PR 🙂

flaviuvadan:
@michaelvay I ended up writing support for task exit hooks in #211. As part of the 3.6.0 release there were multiple changes applied, specifically as they pertained to context management of workflows, and I figured I'd rewrite it really quick as part of the release and comment on the PR for full credit to you! Check out the PR https://github.com/argoproj-labs/hera-workflows/pull/211 and the main commit https://github.com/argoproj-labs/hera-workflows/pull/211/commits/7acc8602c05d4ebe1322eecc21e2862151728175 

Thank you again for submitting this ❤️ Very grateful for the efforts. Closing this in favor of #211 there. We can discuss things on that PR or on the 3.6.0 release #207

PR #199, opened 2022-07-04
support sub_path for VolumeMount

szdr:
Allow subpath to be specified for VolumeMount.

https://github.com/argoproj/argo-workflows/blob/master/sdks/python/client/docs/VolumeMount.md

### DIFF
--- a/tests/test_volumes.py
+++ b/tests/test_volumes.py
@@ -29,3 +29,12 @@ def test_volume_created_with_multiple_access_modes():
 def test_volume_fails_creation_with_string_access_modes():
     with pytest.raises(ValidationError) as e:
         Volume(name='v', size='1Gi', mount_path='/test', access_modes=['ReadWriteTwice'])
+
+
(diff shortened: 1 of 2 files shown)


tests/test_volumes.py
```diff
@@ -29,3 +29,12 @@ def test_volume_created_with_multiple_access_modes():
 def test_volume_fails_creation_with_string_access_modes():
     with pytest.raises(ValidationError) as e:
         Volume(name='v', size='1Gi', mount_path='/test', access_modes=['ReadWriteTwice'])
+
+
+def test_volume_mount():
+    v = Volume(name="v", size="1Gi", mount_path="/test", sub_path="test.txt")
```
flaviuvadan, line 35: very small detail - I think the `sub_path` is a full path rather than a file name

flaviuvadan (approved):
Merging into the 3.6.0 branch. Thank you for the contribution @szdr!!

PR #203, opened 2022-07-08
Add HttpArtifact

Trollgeir:
This PR adds support for the HttpArtifact


flaviuvadan (commented):
Heyo @Trollgeir, thanks for the submission! 🙂 any chance this PR can have some tests so coverage checks pass? That's the only thing needed!

Issue #220, opened 2022-07-17
dynamic field allocation for obtaining outputs and parameters from other tasks


flaviuvadan:
Currently it's challenging/cumbersome to set the input parameters on a task since the signature of the input parameter is `def __init__(self, name: str, path: str, from_task: str, artifact_name: str) -> None:`. To facilitate access to this a task field should dynamically set and expose that is accessible once it's output parameters are set

CC: @trollgeir

flaviuvadan:
Discussed this approach with @Trollgeir and we decided to go with explicit accessors for artifacts and parameters

Issue #218, opened 2022-07-17
Add patch spec 


flaviuvadan:
See https://github.com/argoproj/argo-workflows/blob/master/examples/pod-spec-patch.yaml for an example

Issue #205, opened 2022-07-10
Resource Limits defaults


Trollgeir:
I've noticed how Resource limits - if not defined - default to the minimum requested. This was a gotcha for me as I expected there to be no limits in a resource if I didn't explicitly defined it. Also, is there a reason why the terminology differs form kubernetes (limit vs maximum)?

flaviuvadan:
> Resource limits - if not defined - default to the minimum requested. This was a gotcha for me as I expected there to be no limits in a resource if I didn't explicitly defined it

I can definitely see how this behavior can be confusing and warrants some adjustments. I think we can safely remove the limit as there are cases, like yours, in which users do not want to set an explicit limit

> reason why the terminology differs form kubernetes

Hera aimed to increase access to Argo Workflows for the scientific/DS community at the start. `min` vs `max` are terms that are more broadly accessible

> contradictory to the declaration in "About" for this project:

I see how this manifests as a contradiction but that statement refers to how Hera aims to maintain consistency in vocabulary with Argo, not necessarily Kubernetes. This refers to "workflow", "task", etc, which are higher level entities by comparison to the resource limits set for Kubernetes. Again, I definitely see your argument and perhaps there's space for improvement. 

How about  one of the following
```
Resources(cpu=Tuple[int, int], mem=Tuple[str, str], gpu=int)  # assumes that "nvidia.com/gpu" is the correct selector for GPUs, I prefer this because it's explicit, maybe add an optional field for "GPU" field with a "nvidia.com/gpu" default

Resources(requests: Dict[str, Union[int, str]], limits=Optional[Dict[str, Union[int, str]]])  # arbitrary resource setting, might be hard for users who are not familiar with requests/limits of K8S but this is by far the closest one can get to K8S and allows the removal of the `custom` resources field

Resources(cpu_req: int, mem_req: int, cpu_limit: Optional[int], mem_req: int, mem_limit: Optional[int], ...)  # very verbose and requires a lot of maintenance probably
```

Trollgeir:
If a developer has worked with kubernetes or Argo Workflows before, it would be a seamless transition to construct the `Resources` object as the arguments are the same. If a DS-developer (without background in the former) created a new `Resources` object, I would assume they don't have a preconception over how pod resources are generally defined, unless they're coming from a custom service which had its own API deciding to use different terminology (in other words, non-standard).

[...]

Issue #204, opened 2022-07-08
Dynamically mapping Task inputs to Artifact fields


Trollgeir:
I'm trying to rewrite some yml-files to hera, and I've stumbled upon some issues related to mapping an input variable (a google cloud storage file path) to the parameters of an `InputArtifact` (`GCSArtifact`). 

In the yml-workflow this is done by templating:
```yml
inputs:
    parameters:
      - name: bucket
      - name: filename
    artifacts:
      - name: file
        path: /file_to_process
        gcs:
          bucket: '{{inputs.parameters.bucket}}'
          key: '{{inputs.parameters.filename}}'
```

What's the intended way to use Hera to accomplish this? Should I include templating in the artifact fields, like this?
```python
GCSArtifact(name="file", 
            path="/file_to_process",
            bucket='{{inputs.parameters.bucket}}', 
            key='{{inputs.parameters.filename}}')
```

I'm asking primarily because I've yet to see any templating in the examples.

Trollgeir:
Is this the same as https://github.com/argoproj-labs/hera-workflows/issues/200 ?

Trollgeir:
Given the example shown [here](https://github.com/argoproj-labs/hera-workflows/issues/202#issuecomment-1177513077), I reckon this is actually the right way to implement this. I was under the naive assumption that templates wouldn't be needed with Hera, but I realize that there's really no other way to pass variables through the workflow dynamically without it.

flaviuvadan:
@Trollgeir what do you think about a `ParameterizedArtifact` that has fields set via `inputs.parameters` and users specify the key of that dictionary? I think Hera should be as wary as possible of requiring users to type out the `'{{inputs.parameters.filename}}'` syntax explicitly. It makes working with Argo a bit challenging. Also, there are many users who like Hera specifically because it does not require them to use that syntax and will set it for them.

Trollgeir:
> @Trollgeir what do you think about a `ParameterizedArtifact` that has fields set via `inputs.parameters` and users specify the key of that dictionary? I think Hera should be as wary as possible of requiring users to type out the `'{{inputs.parameters.filename}}'` syntax explicitly. It makes working with Argo a bit challenging. Also, there are many users who like Hera specifically because it does not require them to use that syntax and will set it for them.

[...]

PR #221, opened 2022-07-17
Hera 4.0.0

flaviuvadan:
Adds:
- [ ] #88 - workflow of workflows and patching support
- [ ] #177 - fan out over multiple task implementations through `InputFrom`
- [ ] #218 - pod patch spec via parameterization

Breaking changes:
- [x] #201 - expose namespace configuration on workflow services only rather than workflows
- [ ] #202 215 - support for arbitrary script submissions via `source` rather than `func`
- [ ] #205 - make `Resource` more aligned with the usage and specifications in Kubernetes
- [ ] #220 - dynamic allocation and exposure of task and workflow output fields

Discussion topic:
- [ ] #204 - should we rename some objects so that they are more in line with Argo specifications and allow control over dynamic field reference via `{{...}}` syntax? CC: @trollgeir


flaviuvadan:
Closing in favor of #253 @Trollgeir

PR #230, opened 2022-07-26
Fix undefined task error

akfmdl:
image
Fix undefined variable error


flaviuvadan (approved):
Thank you for the fix @akfmdl! 🚀

PR #232, opened 2022-07-31
poetry files update

reibuildots:
the result of running:
`poetry add "pytz>=2021.3"` 

#227

PR #235, opened 2022-08-10
Remove python <3.11 constraint and unpin transitive dependencies

JacobHayes:
I was adding `hera-workflows` to a package that uses `python = '>=3.9'`, which implicitly supports python 3.11. However, the Hera version spec explicitly excludes 3.11, causing:
```
The current project's Python requirement (>=3.9,<4.0) is not compatible with some of the required packages Python requirement:
- hera-workflows requires Python >=3.7,<3.11, so it will not be satisfied for Python >=3.11,<4.0
```

If this is not strictly needed (eg: no known issues), can we remove that <3.11 constraint?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,7 +27,7 @@ packages = [
 "Bug Tracker" = "https://github.com/argoproj-labs/hera-workflows/issues"
 
 [tool.poetry.dependencies]
-python = ">=3.7, <3.11"
+python = ">=3.7"
 pydantic = "^1.9.0"
 python-dateutil = "^2.8.2"
 urllib3 = "^1.26.8"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -51,13 +51,6 @@ line-length = 119
 target-version = ['py37']
 include = '\.pyi?$'
 
-[tool.coverage.report]
-exclude_lines = [
-    "if TYPE_CHECKING:",
-    "pragma: no cover",
-    "raise NotImplementedError",
-]
-
 [tool.flake8]
 # do not look here for things to lint
 exclude = [".git", "__init__.py"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -51,6 +51,13 @@ line-length = 119
 target-version = ['py37']
 include = '\.pyi?$'
 
+[tool.coverage.report]
+exclude_lines = [
+    "if TYPE_CHECKING:",
+    "pragma: no cover",
+    "raise NotImplementedError",
+]
+
 [tool.flake8]
 # do not look here for things to lint
 exclude = [".git", "__init__.py"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,13 +27,10 @@ packages = [
 "Bug Tracker" = "https://github.com/argoproj-labs/hera-workflows/issues"
 
 [tool.poetry.dependencies]
-python = ">=3.7"
-pydantic = "^1.9.0"
-python-dateutil = "^2.8.2"
-urllib3 = "^1.26.8"
-certifi = "^2021.10.8"
-pytz = ">=2021.3"
 argo-workflows = {version = "6.3.5", allow-prereleases = true}
+pydantic = "^1.9.0"
+python = ">=3.7"
+pytz = "*"
 
 [tool.poetry.dev-dependencies]
 pytest = "*"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "hera-workflows"
-version = "3.6.3"
+version = "3.6.4"
 description="Hera is a Python framework for constructing and submitting Argo Workflows. The main goal of Hera is to make Argo Workflows more accessible by abstracting away some setup that is typically necessary for constructing Argo workflows."
 authors = ["Flaviu Vadan <flaviu.vadan@dynotx.com>"]
 license = "MIT"
--- a/src/hera/config.py
+++ b/src/hera/config.py
@@ -12,7 +12,7 @@ try:
     # __get_config() explicitly disables SSL verification, so urllib3 will throw a warning to the user. Since we have
     # explicitly asked for it to disable SSL, it's safe to ignore the warning.
     urllib3.disable_warnings()
-except ImportError:
+except ImportError:  # pragma: no cover
     pass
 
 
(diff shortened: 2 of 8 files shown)


JacobHayes:
I think a lot of the other deps are unused too (or at least, not by Hera) - I'll push a commit to remove them, but I'm happy to undo if desired. It should reduce the need for/frequency of PRs like https://github.com/argoproj-labs/hera-workflows/pull/232.

src/hera/config.py
```diff
@@ -2,14 +2,18 @@
...
 
-import urllib3
 from argo_workflows.api_client import Configuration as ArgoConfig
 
 from hera.host_config import get_global_host
 
-# __get_config() explicitly disables SSL verification, so urllib3 will throw a warning to the user. Since we have
-# explicitly asked for it to disable SSL, it's safe to ignore the warning.
-urllib3.disable_warnings()
+try:
```
flaviuvadan, line 9: Thank you for adding this as well :)

pyproject.toml
```diff
@@ -27,13 +27,10 @@ packages = [
 "Bug Tracker" = "https://github.com/argoproj-labs/hera-workflows/issues"
 
 [tool.poetry.dependencies]
-python = ">=3.7, <3.11"
-pydantic = "^1.9.0"
-python-dateutil = "^2.8.2"
```
flaviuvadan, line 32: I am trying to recall why this was introduced... I have a hunch it has to do with the way deserialization of cron workflow payloads occurs, which might have a runtime dependency on `python-dateutil` because it assigns a datetime stamp to a field that indicates the last time a cron workflow ran. This happens even when the cron workflow is created at first. However, I could be wrong. Any chance you can submit one of the examples to an Argo cluster and see whether workflows create successfully with this environment configuration?

JacobHayes: Yup, I can test that out in the morning!

I saw we use pytz in the cron module to confirm timezones (I think those are now also in the stdlib, but only for 3.9+ or something), but I didn't see dateutil used directly. It is still brought in as a transitive dependency via `argo-workflows` however, so should still be available. The only dep that got removed completely was `certifi`.

JacobHayes: Looking good, `python-dateutil` still got installed transitively and the scheduled job created alright!

Screen Shot 2022-08-11 at 09 58 27

<details>
<summary>Log</summary>

```sh
[I] ~/hera-test $ python3 -m venv .venv
[I] ~/hera-test $ source .venv/bin/activate.fish
[I] (.venv) ~/hera-test $ pip3 install -e /Users/jacobhayes/src/github.com/argoproj-labs/hera-workflows
Obtaining file:///Users/jacobhayes/src/github.com/argoproj-labs/hera-workflows
  Installing build dependencies ... done
  Checking if build backend supports build_editable ... done
  Getting requirements to build editable ... done
  Preparing editable metadata (pyproject.toml) ... done
Collecting argo-workflows==6.3.5
  Using cached argo_workflows-6.3.5-py3-none-any.whl (1.3 MB)
Collecting pydantic<2.0.0,>=1.9.0
  Downloading pydantic-1.9.2-cp39-cp39-macosx_11_0_arm64.whl (2.5 MB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 2.5/2.5 MB 27.2 MB/s eta 0:00:00
Collecting pytz
  Using cached pytz-2022.1-py2.py3-none-any.whl (503 kB)
Collecting urllib3>=1.25.3
  Using cached urllib3-1.26.11-py2.py3-none-any.whl (139 kB)
Collecting python-dateutil
  Using cached python_dateutil-2.8.2-py2.py3-none-any.whl (247 kB)
Collecting typing-extensions>=3.7.4.3
  Using cached typing_extensions-4.3.0-py3-none-any.whl (25 kB)
Collecting six>=1.5
  Using cached six-1.16.0-py2.py3-none-any.whl (11 kB)
Building wheels for collected packages: hera-workflows
  Building editable for hera-workflows (pyproject.toml) ... done
  Created wheel for hera-workflows: filename=hera_workflows-3.6.3-py3-none-any.whl size=5399 sha256=df3e2f92e2727fb708c0b9a465e479fb7e281d91436cd7ed4c0b8df485886133
  Stored in directory: /private/var/folders/ss/tgmzwbx52y35qpr7rzrtv8ym0000gn/T/pip-ephem-wheel-cache-4nwz6v81/wheels/02/7f/31/bd06123ae4a7affe003677ed2df75772f5694403c190eb6025
Successfully built hera-workflows
Installing collected packages: pytz, urllib3, typing-extensions, six, python-dateutil, pydantic, argo-workflows, hera-workflows
Successfully installed argo-workflows-6.3.5 hera-workflows-3.6.3 pydantic-1.9.2 python-dateutil-2.8.2 pytz-2022.1 six-1.16.0 typing-extensions-4.3.0 urllib3-1.26.11
WARNING: There was an error checking the latest version of pip.
[I] (.venv) ~/hera-test $ cat cron.py
"""This example showcases the cron hello world example of Hera"""

from hera import CronWorkflow, CronWorkflowService, Task


def hello():
    print("Hello, Hera!")


with CronWorkflow(
    "hello-hera-cron",
    "5 4 * * *",
    service=CronWorkflowService(
        host="redacted",
        token="redacted",
    ),
    timezone="UTC",
) as cw:
    Task("t", hello)

cw.create()
[I] (.venv) ~/hera-test $ python3 cron.py
[I] (.venv) ~/hera-test $
```

</details>

flaviuvadan (approved):
Small - bump the PATCH so Arti will get the patched version with new deps 🙂

JacobHayes:
I think codecov is only angry b/c it doesn't respect the pyproject coverage "exclude" and I touched three lines that `coverage` excludes, but it doesn't (even though they were _not_ covered before this PR anyway) 🤷

pyproject.toml
```diff
@@ -54,6 +51,13 @@ line-length = 119
 target-version = ['py37']
 include = '\.pyi?$'
 
+[tool.coverage.report]
+exclude_lines = [
+    "if TYPE_CHECKING:",
+    "pragma: no cover",
+    "raise NotImplementedError",
```
flaviuvadan, line 58: Maybe this should be `raise NotImplementedError()` so that coverage does ignore it, and therefore pass the check? 🤔

JacobHayes: Good call, I added the `()`, but think I also needed to move things to `tox.ini`

PR #244, opened 2022-08-26
Tolerations can now be set via Workflow

jpmnteiro:
Argo supports tolerations being set via the workflow spec, which get propagated to the pods of the workflow.

However, hera, at the moment, doesn't offer the ability to set these tolerations at the workflow level, forcing the caller to set the tolerations at the task level.

The aim of this PR is to add tolerations to the workflow.

I've also taken the opportunity to lightly refactor `hera.Toleration` with a function that converts it to an `argo.Toleration`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,12 @@ The general format is:
 
 ```
 
+# 3.7.0 - DATE (26/08/2022)
+
+### Added
+
+- tolerations can now be set via workflow
+
 # 3.6.4 - DATE (11/08/2022)
 
 ### Removed
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "hera-workflows"
-version = "3.6.4"
+version = "3.7.0"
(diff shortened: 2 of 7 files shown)


flaviuvadan:
This looks awesome @jpmnteiro! Any chance you are open towards bumping the version of Hera to 3.7 [here](https://github.com/argoproj-labs/hera-workflows/pull/244/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R3)? This is technically a new feature, and a MINOR bump seems more fitting for that :) [This PR](https://github.com/argoproj-labs/hera-workflows/pull/241) will likely be merged soon, and it bumps to 3.6.5. Rather than merge and release two version I will make that PR merge into a 3.7.0 release branch and unify it with your PR (3.7.0), so we can release the other PR and your contribution together!

jpmnteiro:
@flaviuvadan of course, happy to do it.

* cue magic incantation *

jpmnteiro:
@flaviuvadan just a heads up, I found a bug with the tolerations, I'll submit a fix shortly.

In a nutshell, while the tolerations are being set in the templates, they (also) need to be set in the [workflow spec](https://github.com/argoproj-labs/hera-workflows/blob/main/src/hera/workflow.py#L156). Otherwise, they won't be applied.

Note, the same bug applies to nodeSelectors and affinities.

PR #246, opened 2022-08-30
Tolerations, nodeSelectors and affinity not being properly set

jpmnteiro:
As I commented on #244, the tolerations, nodeSelectors and affinity are now being properly set in the workflow with the current code:

```python
    workflow = Workflow(
        name="my-wf",
        tolerations=generate_workflow_tolerations(),
        node_selectors=generate_workflow_node_selectors(),
        affinity=generate_workflow_affinity(),
    )
```

results in:

```yaml
spec:
  templates:
  - templates:
    ...
    nodeSelector:
      some_key: value
    tolerations:
    - effect: NoSchedule
      key: some_key
      operator: Exists
    affinity:
      ...
```

These entries in the yaml will be ignored by Argo/k8s and won't have any effect on the workflow, since they are *children* of `templates`.

The solution is to have the entries defined as siblings of `templates`:

```yaml
spec:
  nodeSelector:
    some_key: value
  tolerations:
  - effect: NoSchedule
    key: some_key
    operator: Exists
  affinity:
    ...
  templates:
  - templates:
    ...
```

This matches the API seen in [IoArgoprojWorkflowV1alpha1WorkflowSpec](https://github.com/argoproj/argo-workflows/blob/master/sdks/python/client/argo_workflows/model/io_argoproj_workflow_v1alpha1_workflow_spec.py#L135).

---

Apologies for not spotting this sooner :/

### DIFF
--- a/src/hera/workflow.py
+++ b/src/hera/workflow.py
@@ -143,23 +143,19 @@ class Workflow:
             parallelism=self.parallelism,
         )
 
+        self.spec = IoArgoprojWorkflowV1alpha1WorkflowSpec(
+            templates=[self.template],
+            entrypoint=self.name,
+            volumes=[],
+            volume_claim_templates=[],
+            parallelism=self.parallelism,
+        )
+
         if self.workflow_template_ref:
             self.workflow_template = IoArgoprojWorkflowV1alpha1WorkflowTemplateRef(name=self.workflow_template_ref)
-            self.spec = IoArgoprojWorkflowV1alpha1WorkflowSpec(
-                workflow_template_ref=self.workflow_template,
-                entrypoint=self.workflow_template_ref,
-                volumes=[],
-                volume_claim_templates=[],
-                parallelism=self.parallelism,
-            )
-        else:
-            self.spec = IoArgoprojWorkflowV1alpha1WorkflowSpec(
-                templates=[self.template],
-                entrypoint=self.name,
-                volumes=[],
-                volume_claim_templates=[],
-                parallelism=self.parallelism,
-            )
+            self.spec.workflow_template_ref = self.workflow_template
+            self.spec.entrypoint = self.workflow_template_ref
+
         if self.volumes is not None:
             for volume in self.volumes:
                 if isinstance(volume, Volume):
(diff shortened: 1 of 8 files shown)


src/hera/workflow.py
```diff
@@ -143,23 +143,19 @@ def __init__(
...
 
+        self.spec = IoArgoprojWorkflowV1alpha1WorkflowSpec(
+            templates=[self.template],
+            entrypoint=self.name,
+            volumes=[],
+            volume_claim_templates=[],
+            parallelism=self.parallelism,
+        )
+
         if self.workflow_template_ref:
```
jpmnteiro, line 154: Makes this bit a little bit simpler to read (no code dupe): we set the spec with the template by default and only override with a ref if it's set.

src/hera/workflow.py
```diff
@@ -193,8 +189,9 @@ def __init__(
             setattr(self.spec, "image_pull_secrets", secret_refs)
 
         if self.affinity:
-            setattr(self.exit_template, "affinity", self.affinity.get_spec())
```
jpmnteiro, line 196: nit: follow the same order when setting attributes (spec -> dag -> template -> exit_template).

flaviuvadan (commented):
Super cool @jpmnteiro! Thank you for the fix! 🚀

Any chance this fix can be applied to cron workflow and workflow template?

Also, is it worth adding the tolerations and selectors to the template if they are already part of the spec? Feels like it's redundant for the workflow top level template since Argo will apply the spec to the template and, in this case, the template is no different

jpmnteiro:
@flaviuvadan good point, I'll fix those two as well.

Re: being redundant adding it to the inner templates, I agree with you that it _looks/feels_ redundant, but I'm not 100% certain that removing it won't break something else given that Argo's own API models accept these attributes, which, to me, means they are being used somewhere (their docs mention that it's possible for some of the attributes set in the workflow to be overridden).

Furthermore, there's a non 0% chance that someone out there is relying on these being set. 

With that in mind, I decided to keep them to be on the safe side. Happy to defer to your judgement and decision though!

flaviuvadan:
> Furthermore, there's a non 0% chance that someone out there is relying on these being set

This is a great point, actually, so let's merge it this way :)

flaviuvadan (approved):
Thank you for the contribution @jpmnteiro! 🚀

jpmnteiro:
(for the lack of a better discussion forum)

@flaviuvadan is there anything else you need help with on hera? Perhaps a roadmap you're trying to fulfil? Would like to continue contributing.

Issue #202, opened 2022-07-06
Bash script example?


Trollgeir:
I've been struggling a bit with the API, and I can't quite seem to figure out how to run a generic script with Hera (independent of python). [Here's an example from the argo workflows running a bash script](https://github.com/argoproj/argo-workflows/blob/master/examples/scripts-bash.yaml).

flaviuvadan:
Hi @Trollgeir 🙂 I believe you're looking for [this example](https://github.com/argoproj-labs/hera-workflows/blob/main/examples/container.py)! A similar example that runs 2 bash commands with Hera:
```python
"""This example showcases how to run a container, rather than a Python, function, as the payload of a task in Hera"""

from hera import Task, Workflow, InputParameter, OutputPathParameter, set_global_host, set_global_token

set_global_host(...)
set_global_token(...)

w = Workflow('container')

g = Task(
    'generate',
    image='alpine:latest',
    command=['sh', '-c'],
    args=["echo $RANDOM > result"],
    outputs=[OutputPathParameter('result', '/result')],
)
c = Task(
    'consume',
    image='alpine:latest',
    command=['sh', '-c'],
    args=["echo result was: {{inputs.parameters.message}}"],
    inputs=[InputParameter('message', g.name, 'result')],
)

g >> c
w.add_tasks(g, c)
w.create()
```

Issue #201, opened 2022-07-06
Default namespace for workflows


Trollgeir:
Is there a specific reason that the default namespace for a `Workflow` object doesn't inherit from `WorkflowService`?

flaviuvadan:
Hello again @Trollgeir! By design a workflow service can operate independently of a workflow object. The service takes a namespace and offers specific operations for that namespace, such as resuming a workflow, deleting, etc. Some users have argo installed in multiple namespaces, so they can use a service for performing operations within that namespace, along with submitting workflows to that namespace. However, a workflow service also hosts a token and a domain where operations can be applied, such as workflow submission. Therefore, a workflow service can be used to, technically, submit a workflow to any namespace, provided that a namespace is specified. For instance, I can have a `ws = WorkflowService(...)` and delete a workflow from namespace `A` but then I can also create a `w = Workflow(..., service=ws)` and submit the workflow to namespace `B` via `w.create(namespace='B')`. Currently, the workflow has a condition upon creation to check whether a namespace was provided on the `create(...)` call and fall back on the `Workflow.namespace` field otherwise, which provides additional flexibility. 

One problem that I noticed, as you brought up the question (thank you!), is that the overall namespace on the service, provided through `__init__` is not even used by the service inside independent methods as the methods rely on the `namespace` provided to them. I believe this is where the confusion comes from... you may have passed a `namespace` to the service and expected that workflow to be submitted to that namespace by virtue of submission through the service. What do you think is a better design here? Perhaps we can remove the `namespace` from the workflow service `__init__` and make it a general workflow orchestration service used for submission, listing, etc? Or, should we make the service take in a global `namespace`, as it does now, and make all calls use that global namespace? This prevent the workflow from setting a different namespace upon submission, so all workflows will go to the service namespace. Thoughts?

Trollgeir:
> ... you may have passed a namespace to the service and expected that workflow to be submitted to that namespace by virtue of submission through the service. What do you think is a better design here? 

Yeah that was my expectation. Right now I feel like this is a gotcha-scenario, 

Removing it as an argument for the `WorkflowService` would eliminate the gotcha, which I like better than the way things are now.

[...]

Issue #177, opened 2022-06-02
Creating a nested parallel DAG


vikramsg:
Hi, 

I am trying to understand the parallelism syntax for Hera but right now I am unable to understand if nested parallelism can work. What I mean is the following. Supposed I have a generator task and then multiple consumer tasks that must run one after the other in parallel. For example, the following. 

```
def generate() -> None:
    import json
    import sys

    chunk_number_list = list(range(0, 10))
    json.dump([{"chunk_number": chunk_number} for chunk_number in chunk_number_list], sys.stdout)


def consumeA(chunk_number) -> None:
    print(f"A {chunk_number}")
    
def consumeB(chunk_number) -> None:
    print(f"B {chunk_number}")
    
generate_task = Task(
    name="generate",
    func=generate,
    image=image,
)
consume_taskA = Task(
    name="consumeA",
    func=consumeA,
    input_from=InputFrom(name="generate", parameters=["chunk_number"]),
    image=image,
)
consume_taskB = Task(
    name="consumeB",
    func=consumeB,
    input_from=InputFrom(name="generate", parameters=["chunk_number"]),
    image=image,
)
generate_task.next(consume_taskA)
consume_taskA.next(consume_taskB)

workflow.add_tasks(generate_task, consume_taskA, consume_taskB)
```

What I want is that after doing the generate task, the workflow should fan out and then within each parallel task, it should do A and B serially. However, what is actually happening is that it first fans out for A, waits until completion and then again fans out for B. 

Please tell me if the question is not clear or you need any other information. Thanks for the help.

flaviuvadan:
Hello @vikramsg! That is something that is possible in Argo Workflows since fanning out over `InputFrom` uses a single template, and a single template can point to an independent DAG, which can represent a series of templates/tasks to execute on the fanned out output from a task. However, this is not something that is currently supported by Hera. I think it would require something like a `TaskChain` concept, which would be almost a pseudo-workflow representation. With that users could do something like 
```python
def f1(...): ...
def f2(...): ...
def f3(...): ...

generate = Task(...)
consume = TaskChain([f1, f2, f3], input_from=InputFrom(generate.name, ...))
generate >> consume
```
and that would translate into the structure you would like to use. Does that sound like something that would solve the problem for you?

vikramsg:
Hi @flaviuvadan ,
Thanks for looking at this. Yes, I think that would solve the issue,

flaviuvadan:
[...]

Issue #88, opened 2022-02-08
Workflows of workflows support


flaviuvadan:
https://argoproj.github.io/argo-workflows/workflow-of-workflows/

dixon1e:
This is one of the primary requests from our scientists.

bchalk101:
The requirement for this has re-arisen for us, specifically we now are having to implement stateful sets for daemons, as seen here https://github.com/argoproj/argo-workflows/blob/master/examples/daemoned-stateful-set-with-service.yaml

Happy to jump onto implementing it, if its not being worked on otherwise.

100mi:
Hey, I was wondering if there is a reference example for implementing `workflow-of-workflows` using hera-workflows ?

flaviuvadan:
Going to be released in V4

PR #253, opened 2022-09-09
Hera V4

flaviuvadan:
## TODO

- [x] docstrings
- [x] examples
- [x] unit tests
- [x] functional tests
- [x] dependency check can lead to false positive results, `TODO:task.py:260`
- [x] ~`when` (decided to move ahead with examples, study the ergonomics there, and then revisit, deprioritized)~

Adds:
- [x] #88 - workflow of workflows and patching support
- [x] #177 - fan out over multiple task implementations through `InputFrom`
- [x] #218 - pod patch spec via parameterization

Breaking changes:
- [x] #201 - expose namespace configuration on workflow services only rather than workflows
- [x] #202 215 - support for arbitrary script submissions via `source` rather than `func`
- [x] #205 - make `Resource` more aligned with the usage and specifications in Kubernetes
- [x] #220 - dynamic allocation and exposure of task and workflow output fields

Discussion topic:
- [x] #204 - should we rename some objects so that they are more in line with Argo specifications and allow control over dynamic field reference via `{{...}}` syntax? CC: @trollgeir


src/hera/resources.py
```diff
@@ -70,15 +70,20 @@ def __post_init__(self):
...
+        if self.memory_request is None and self.memory_limit is not None:
+            self.memory_request = self.memory_limit
+
     def build(self) -> ResourceRequirements:
         """Builds the resource requirements of the pod"""
         resources = dict()
 
         if self.cpu_limit is not None:
-            resources = merge_dicts(resources, dict(limit=dict(cpu=self.cpu_limit)))
+            resources = merge_dicts(resources, dict(limit=dict(cpus=str(self.cpu_limit))))
```
JacobHayes, line 83: Should this one be `cpus` plural (but singular below for request)?

JacobHayes:
🎉

src/hera/resources.py
flaviuvadan: Independent of anything else yes, I agree with you, but recall that `ResourceRequirements` is mapped to Kubernetes resource requirements, and those use `cpu` only, inside the `requests` and `limits` fields! See [docs here](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#resource-units-in-kubernetes)!

JacobHayes: In that case, should this be:
```suggestion
            resources = merge_dicts(resources, dict(limit=dict(cpu=str(self.cpu_limit))))
```
?

flaviuvadan: Oh... yes 😬

PR #254, opened 2022-09-12
Add docstrings for V4

flaviuvadan:


src/hera/artifact.py
```diff
@@ -20,10 +20,12 @@ class Artifact:
...
-    path: str
+    path: Optional[str] = None
         The path to the file to be assembled into an artifact. This path is relative to the task environment.
         If clients have a volume where results are stored, it's the path to the created object on the
         respective volume.
+    from_task: Optional[str] = None
+        The name of the task that generates the artifact.
 
     Notes
     -----
```
Trollgeir, line 31: I'm guessing this is obsolete now

src/hera/task.py
```diff
@@ -500,7 +502,7 @@ def deduce_parameters(
                             "`with_params` is empty and there exists non-default arguments "
                             f"which aren't covered by `inputs`: {missing_args}"
                         )
-            elif isinstance(self.with_param, Union[str, Parameter]):
+            elif isinstance(self.with_param, str) or isinstance(self.with_param, Parameter):
```
Trollgeir, line 505: Aren't these functionally equivalent? I feel like the former reads better.

src/hera/task.py
```diff
@@ -664,17 +666,20 @@ def build_volume_mounts(self) -> Optional[List[VolumeMount]]:
...
         return [v.build_claim_spec() for v in self.volumes if isinstance(v, Volume)]
 
-    def build_persistent_volume_claims(self):
+    def build_persistent_volume_claims(self) -> Optional[List[ArgoVolume]]:
+        """Assembles the list of Argo volume specifications"""
         return [
             v.build_claim_spec()
             for v in self.volumes
-            if isinstance(v, Union[ExistingVolume, SecretVolume, EmptyDirVolume])
+            if isinstance(v, ExistingVolume) or isinstance(v, SecretVolume) or isinstance(v, EmptyDirVolume)
```
Trollgeir, line 678: Same as above

src/hera/artifact.py
Trollgeir: Referring to the notes below.

src/hera/task.py
flaviuvadan: Does it work in all versions of Python 3.7+? 😮

flaviuvadan: I tried the following in Hera's virtual environment and it raised an error:
```python
from typing import Union

class A:
    f: int = 42

def foo(x: Union[str, A]):
    print(isinstance(x, Union[str, A]))

foo(a)
```

```
~/.pyenv-x86_64/versions/3.7.12/lib/python3.7/typing.py in __subclasscheck__(self, cls)
    722             if cls._special:
    723                 return issubclass(cls.__origin__, self.__origin__)
--> 724         raise TypeError("Subscripted generics cannot be used with"
    725                         " class and instance checks")
    726 

TypeError: Subscripted generics cannot be used with class and instance checks

```

Trollgeir: These both run fine for me, but I'm on python 3.10. So yeah, let's use your revision :+1:

Trollgeir (approved):

PR #255, opened 2022-09-12
add resource template by @michaelvay

flaviuvadan:
Main contributor, full credit: @michaelvay 

Reopened PR because of V4 conflicts


Trollgeir:
This has a `pydantic` dependency which has been revoked in 4.0.0

flaviuvadan:
> This has a `pydantic` dependency which has been revoked in 4.0.0

Removed the dependency and instead used dataclasses

Trollgeir (approved):
Nice! Like the use of `asdict`, might consider that pattern for a lot of other places.

PR #256, opened 2022-09-13
V4 examples

flaviuvadan:
@Trollgeir the main branch of Hera seems to expose parameters via environment variables quite effectively. This functionality is now gone, and I wonder if we can create a env specification similar to [VariableAsEnv](https://github.com/argoproj-labs/hera-workflows/blob/main/src/hera/variable.py#L31) and have [this example](https://github.com/argoproj-labs/hera-workflows/blob/main/examples/variables.py) work almost as is. At the moment the interface makes me struggle setting up the right parameter + env interaction because `tasks.daemon.ip` is not resolved by the environment variable field. That is the only example that does not work at the moment, everything else is fine! 🚀

### DIFF
--- a/examples/any_success_all_fail.py
+++ b/examples/any_success_all_fail.py
@@ -1,4 +1,4 @@
-from hera import Task, Workflow, WorkflowService
+from hera import Task, Workflow
 
 
 def foo(a):
@@ -17,9 +17,8 @@ def fail(a):
     raise Exception(a)
 
 
-with Workflow(
-    "any-success-all-fail", service=WorkflowService(host="https://my-argo-server.com", token="my-auth-token")
-) as w:
+# assumes you used `hera.set_global_token` and `hera.set_global_host` so that the workflow can be submitted
+with Workflow("any-success-all-fail") as w:
     t1 = Task("t1", random_fail, [{"a": 1}, {"a": 2}, {"a": 3}])
     t2 = Task("t2", fail, [{"a": 1}, {"a": 2}, {"a": 3}])
     t3 = Task("t3", foo, [{"a": 1}, {"a": 2}, {"a": 3}])
--- a/examples/artifact.py
+++ b/examples/artifact.py
@@ -4,7 +4,7 @@ The first task, t1, creates a file located at `/file` containing a message. The 
 task, t2, takes this artifact, places it at its own `/file` path, and print out the content.
 """
 
-from hera import Artifact, Task, Workflow, WorkflowService
+from hera import Artifact, Task, Workflow
 
 
 def writer():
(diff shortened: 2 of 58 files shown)


Trollgeir:
Have you tried this?

```# Enablement of https://argoproj.github.io/argo-workflows/variables/
# This example creates two tasks, one of the Tasks is a deamond task and its IP address is shared with the second task
# The daemoned task operates as server, serving an example payload, with the second task operating as a client, making
# http requests to the server

from hera import EnvSpec, Task, Workflow, WorkflowService


def server():
    from http.server import BaseHTTPRequestHandler, HTTPServer

    class MyServer(BaseHTTPRequestHandler):
        def do_GET(self):
            self.send_response(200)
            self.send_header("Content-type", "application/json")
            self.end_headers()
            self.wfile.write(bytes("{'name':'John'}", "utf-8"))

    webServer = HTTPServer(("0.0.0.0", 8080), MyServer)
    webServer.serve_forever()


def consumer():
    import http.client
    import os

    print(os.environ["SERVER_IP"])
    server_ip = os.environ["SERVER_IP"].replace('"', "")
    connection = http.client.HTTPConnection(f"{server_ip}:8080")
    connection.request("GET", "/")
    response = connection.getresponse()
    print(response.read())


with Workflow("variables", service=WorkflowService(host="my-argo-server.com", token="my-auth-token")) as w:
    d = Task("daemon", server, daemon=True)
    t = Task("consumer", consumer, env=[EnvSpec(name="SERVER_IP", value_from_input=d.ip)])
    d >> t

w.create()
```

Trollgeir:
I like the `on_other_output` method. However, should we consider expanding this to include specific outputs?
E.g:
```python

def on_other_output(self, other: "Task", operator: Operator, value: str, output_name: Optional[str] = None) -> "Task":
    """
    Execute this task based on the `other` result. Will default to get the tasks output, optionally getting a specific output if `output_name` is specified.
    """
    if output_name:
        output = other.get_output(output_name)
        assert isinstance(output, Parameter), "Output is not a parameter"
        output = output.value
    else:
        output = other.get_result()
                
    expression = f"'{output}' {operator} {value}"
    if self.when:
        self.when += f" {Operator.And} {expression}"
    else:
        self.when = expression
    other.next(self)
    return self
```

Warning: Untested!

Trollgeir:
Could we rename `on_other_output` to `on_other_result` and instead introduce something like this for `on_other_output` ?

```python
def on_other_output(self, other: "Task", operator: Operator, value: str, output_name: str) -> "Task":
    """
    Execute this task based on the `other` tasks' output.
    """
    output = other.get_output(output_name)
    assert isinstance(output, Parameter), "Output is not a parameter"
                
    expression = f"'{output.value}' {operator} {value}"
    if self.when:
        self.when += f" {Operator.And} {expression}"
    else:
        self.when = expression
    other.next(self)
    return self
```

Trollgeir (approved):
Looks good!

PR #257, opened 2022-09-14
Fix dependency check

Trollgeir:


src/hera/task.py
```diff
@@ -237,6 +237,17 @@ def ip(self):
...
 
+    def _get_dependency_tasks(self) -> List[str]:
+        """Extract task names from `depends` string"""
+        if self.depends is None:
+            return []
+        # Filter out operators
+        all_operators = [o.value for o in Operator]
+        tasks = [t for t in self.depends.split() if t not in all_operators]
+        # Remove dot suffixes
+        task_names = [t.split(".")[0] for t in tasks]
```
flaviuvadan, line 248: Is it possible for this to yield false positives via `dag.tasks.X` specifications?

flaviuvadan (approved):
Small question for my own understanding - LGTM!

src/hera/task.py
Trollgeir: I _think_ it might if the user sets their own `self.depends` to something custom.

PR #259, opened 2022-09-14
Avoid circular dag_context import

JacobHayes:
I think `dag_context` imports were moved into the functions, but the root import was left behind, causing circular imports. This remove that and fixes the imports!


JacobHayes:
I think these breaks are just the same as those in the base branch, https://github.com/argoproj-labs/hera-workflows/pull/253.

flaviuvadan (approved):
Thank you @JacobHayes ❤️

PR #260, opened 2022-09-14
Fix flake8 issues in v4rc

JacobHayes:
Should fix the flake8 issues in https://github.com/argoproj-labs/hera-workflows/pull/253 and https://github.com/argoproj-labs/hera-workflows/pull/259. I didn't get around to the mypy ones.


JacobHayes:
Just pushing to #259 since that fixes a lint issue too (the last one here)

PR #264, opened 2022-09-16
Hera/4.0.0rc2

flaviuvadan:
This is a follow up PR that collects some of the feedback post-V4 release:
- [x] fix container fanout example
- [x] fix `cpu` resource field that incorrectly uses `cpus` in one location

### DIFF
--- a/examples/dynamic_fanout_container.py
+++ b/examples/dynamic_fanout_container.py
@@ -5,27 +5,22 @@ tasks are also container only.
 More details can be found here: https://github.com/argoproj-labs/hera-workflows/issues/250
 """
 
-from hera import InputFrom, Task, Workflow, WorkflowService
+from hera import Task, Workflow
 
-
-with Workflow(
-    "dynamic-fanout-container", service=WorkflowService(host="https://my-argo-server.com", token="my-auth-token"),
-) as w:
-
-    # this can be anything! e.g fetch from some API, then in parallel process all entities; chunk database records
+# assumes you used `hera.set_global_token` and `hera.set_global_host` so that the workflow can be submitted
+with Workflow("dynamic-fanout-container") as w:
+    # this can be anything! e.g. fetch from some API, then in parallel process all entities; chunk database records
     # and process them in parallel, etc.
     generate_task = Task(
         "generate", image="alpine:latest", command=["echo", '[{"value": "a"}, {"value": "b"}, {"value": "c"}]'],
     )
-
     fanout_task = Task(
         "fanout",
-        input_from=InputFrom("generate", ["value"]),
+        with_param=generate_task.get_result(),  # this make the task fan out over the `with_param`
+        inputs=[generate_task.get_result_as('value')],  # this sets the correct input parameter so the result is used
         image="alpine:latest",
-        command=["echo"],
-        args=["{{inputs.parameters.value}}"],
+        command=["echo", "{{inputs.parameters.value}}"],
     )
-
     generate_task >> fanout_task
 
 w.create()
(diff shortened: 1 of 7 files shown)


examples/dynamic_fanout_container.py
```diff
@@ -5,27 +5,22 @@
...
         "fanout",
-        input_from=InputFrom("generate", ["value"]),
+        with_param=generate_task.get_result(),  # this make the task fan out over the `with_param`
+        inputs=[generate_task.get_result_as('value')],  # this sets the correct input parameter so the result is used
         image="alpine:latest",
-        command=["echo"],
-        args=["{{inputs.parameters.value}}"],
+        command=["echo", "{{inputs.parameters.value}}"],
     )
-
```
Trollgeir, line 28: This will make the `with_param` redundant because every task is getting the entire output of `generate_task`

We need to do this:

```python
    fanout_task = Task(
        "fanout",
        with_param=generate_task.get_result(),  # this make the task fan out over the `with_param`
        inputs=[Parameter(name="value", value="{{item.value}}")],  # this sets the correct input parameter so the result is used
        image="alpine:latest",
        command=["echo", "{{inputs.parameters.value}}"],
    )
```
Since this isn't a python function it's not as easy to deduce the argo-templates (maybe in the future?). However, this could serve as a good example of how people can use the native template system.

Trollgeir (dismissed):

examples/dynamic_fanout_container.py
Trollgeir: An alternative workflow is:

```python
from hera import Task, Workflow

def echo(value: str) -> str:
    return f"echo {value}"

with Workflow(
    "dynamic-fanout-container",
) as w:

    # this can be anything! e.g fetch from some API, then in parallel process all entities; chunk database records
    # and process them in parallel, etc.
    generate_task = Task(
        "generate",
        image="alpine:latest",
        command=["echo", '[{"value": "a"}, {"value": "b"}, {"value": "c"}]'],
    )

    fanout_task = Task(
        "fanout",
        echo,
        with_param=generate_task.get_result(),  # this make the task fan out over the `with_param`
        image="alpine:latest",
        command=["sh"]
    )

    generate_task >> fanout_task
```

However, this will print `{value:x}` for each step. This is because it's impossible to know what the content of the `with_param` will be (could be a list of values, or a list of dicts). Hera will only assume the `with_param` contains dicts if the consuming function has multiple arguments.

If we'd use this example, it'd be better if `generate_task` produced a list instead.

flaviuvadan: Adjusted based on the first suggestion!

src/hera/io.py
```diff
@@ -56,5 +56,11 @@ def _validate_io(self):
...
-        assert len(set([i.name for i in self.inputs])) == len(self.inputs), "input objects must have unique names"
-        assert len(set([o.name for o in self.outputs])) == len(self.outputs), "output objects must have unique names"
+        i_parameters = [obj.as_input() for obj in self.inputs if isinstance(obj, Parameter)]
+        i_artifacts = [obj.as_input() for obj in self.inputs if isinstance(obj, Artifact)]
+        o_parameters = [obj.as_output() for obj in self.outputs if isinstance(obj, Parameter)]
+        o_artifacts = [obj.as_output() for obj in self.outputs if isinstance(obj, Artifact)]
+        assert len(set([i.name for i in i_parameters])) == len(i_parameters), "input parameters must have unique names"
+        assert len(set([i.name for i in i_artifacts])) == len(i_artifacts), "input artifacts must have unique names"
+        assert len(set([o.name for o in o_parameters])) == len(o_parameters), "input parameters must have unique names"
+        assert len(set([o.name for o in o_artifacts])) == len(o_artifacts), "output objects must have unique names"
```
Trollgeir, line 66: I made a typo here, should be "artifacts" not "objects"

Trollgeir (approved):
LGTM!

PR #266, opened 2022-09-17
Split artifacts and parameters

Trollgeir:
* Split `get_output` into `get_artifact` and `get_parameter` for `Task`
* Refactor `get_outputs_as` to `get_parameters_as` (this never applied for artifacts)
* `as_name` argument in `get_output` moved as a separate method for both artifact and parameter. E.g: `get_parameter("param").as_name("value")`
* Bucket artifacts has `archive` argument
* Add new examples

### DIFF
--- a/examples/artifact.py
+++ b/examples/artifact.py
@@ -23,7 +23,7 @@ with Workflow("artifact") as w:
     c_t = Task(
         "consumer",
         consumer,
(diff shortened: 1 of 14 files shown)


flaviuvadan (approved):
🚀 really like the fanout and parallel DAG examples!

PR #274, opened 2022-09-23
Update some examples

Trollgeir:


examples/env.py
```diff
@@ -0,0 +1,29 @@
+"""
+This example showcases how Hera can dynamically set environmental variables
+"""
+
+from hera import ConfigMapEnvFromSpec, EnvSpec, Task, Workflow
+from hera.parameter import Parameter
```
flaviuvadan, line 6: Minor, minor suggestion 
```suggestion
from hera import ConfigMapEnvFromSpec, EnvSpec, Task, Workflow, Parameter
```

flaviuvadan (approved):
The `EnvSpec` rename can occur as part of the 4.0.0 change perhaps? There's not much left! 🥳

PR #275, opened 2022-09-25
Rework deduction and add with_sequence

Trollgeir:


src/hera/task.py
```diff
@@ -89,6 +92,9 @@ class Task(IO):
         function call with the given parameters. Otherwise, if parallel is False and the list of params is a list of
         lists (each list contains a series of values to pass to the function), the task will execute as a task
         group and schedule multiple function calls in parallel.
+    with_sequence: Optional[dict] = None,
```
flaviuvadan, line 95: I like this! It's super flexible. I wonder if there are some problems that some "magic" object, `Sequence`, would solve for the user?

flaviuvadan:
Should we merge this in to `hera/4.0.0rc2` and release soon?

src/hera/task.py
Trollgeir: I could be ok with this, but then I feel like we should also add `value_from` as a dedicated object.

Trollgeir:
> Should we merge this in to `hera/4.0.0rc2` and release soon?

Feel free to merge once approved :)

flaviuvadan (approved):
🚀

src/hera/task.py
flaviuvadan: Sounds reasonable. I think we can always add such magic and support `Union[dict, MagicSequence]` :)

Issue #276, opened 2022-09-25
Is Python 3.11 supported? - possible accidental revert


nikhiljha:
I noticed this commit https://github.com/argoproj-labs/hera-workflows/commit/c93b1a0635177812de5e87cce30b47c67f39e1fb modified `pyproject.toml` to enforce a constraint that `python < 3.11`, which this commit previously removed:https://github.com/argoproj-labs/hera-workflows/commit/f6cf903b8974a76c814d636009b09085dfd3cbf4. It doesn't seem like there's any reason to explicitly disallow Python 3.11. Was that intentional, or were mistakes made while merging branches? 😄

Right now, poetry complains when I try to depend on `hera-workflows`...

> The current project's Python requirement (>=3.10,<4.0) is not compatible with some of the required packages Python requirement:
>  - hera-workflows requires Python >=3.7,<3.11, so it will not be satisfied for Python >=3.11,<4.0

Trollgeir:
Nice find!
I'm fairly sure this was accidental. @flaviuvadan ?

flaviuvadan:
That was likely a bad merge conflict resolution, apologies 😕 we can likely remove it!

PR #278, opened 2022-09-25
Remove Py3.11 limit

flaviuvadan:
Solves: #276 

Note that a limit is still required for otherwise Poetry complains that the `urllib3` dependency requires a limit of <Py4


flaviuvadan:
@Trollgeir we should merge this into `hera/4.0.0rc2`, release `rc3` (yes, I know, I messed up the branch names 😢), and release the official 4.0.0 version after!

Issue #277, opened 2022-09-25
getting typed Kubernetes objects without setting a fake address / port


nikhiljha:
I plan to use `hera-workflows` to generate Argo Workflows templates as typed objects (which I'll render to YAML/JSON), and then apply them to my cluster at a later time. This is possible right now by...

```
$ export ARGO_SERVER_PORT_2746_TCP_PORT=fake
$ export ARGO_SERVER_PORT_2746_TCP_ADDR=fake
$ python
>>> from hera import Operator, Task, WorkflowTemplate, set_global_token
>>> set_global_token("fake")
>>> with WorkflowTemplate("workflow-template") as w:
...    Task("cowsay", image="docker/whalesay", command=["cowsay", "foo"])
...
>>> obj = w.build() # This gives me a typed Kubernetes object, which is what I want!
```

I'd prefer if I didn't have to input fake values in order to do this. I'm happy to submit a PR, but I'm not sure what solution to this the maintainers would prefer:

- Maybe the validation could happen when you try to do `w.create()` instead?
- An optional parameter in the `WorkflowTemplate` initializer that skips that validation?
- Something else?

Thanks! :)

Trollgeir:
Thanks for the issue!

I was looking at this, and I feel like there are two natural paths:
* Delay instantiation of the `WorkflowService` object until it has to exist (on `create()` `update()` etc.)
* Allow unset variables in `WorkflowService`, which will raise when trying to do actions on the service.

The first one leads to a lot of repetitive several places, which I sort of don't like.
```
if self.service is None:
    self.service = WorkflowService()
```

flaviuvadan:
Thank you for the issue @nikhiljha! 

@Trollgeir the second option sounds reasonable to me given the existing usage of the underlying Argo SDK service - instantiated upon a create/update/etc. call. Do you think this will have impact on testing/mocking ability? That is one thing that is a downside of using the Argo SDK services the way Hera does at the moment

nikhiljha:
If you're happy with the second option (allow unset variables in WorkflowService, which will raise when trying to do actions on the service), I can try to make a PR if that'd be helpful! I'll likely be poking around here quite a bit in the future if we (@ocf) end up using it, so it could be nice to get the ball rolling with a relatively simple contribution.

flaviuvadan:
@nikhiljha, out of curiosity, why use `obj = w.build()` vs. `w.create` to submit workflows to Argo via Hera?

[...]

PR #282, opened 2022-09-27
Return argo result on submission

Trollgeir:
If we're using concepts like `generate_name`, we won't know the final name of the workflow until we submit it. The user might want to fetch this via the returned object upon creation, which holds this information. Typically: `result.metadata['name']`

### DIFF
--- a/src/hera/cron_workflow.py
+++ b/src/hera/cron_workflow.py
@@ -100,18 +100,17 @@ class CronWorkflow(Workflow):
             spec=cron_workflow_spec,
         )
 
-    def create(self) -> "CronWorkflow":
+    def create(self) -> IoArgoprojWorkflowV1alpha1CronWorkflow:
         """Creates the cron workflow"""
         if self.in_context:
             raise ValueError("Cannot invoke `create` when using a Hera context")
-        self.service.create_cron_workflow(self.build())
-        return self
+        return self.service.create_cron_workflow(self.build())
 
     def delete(self) -> Tuple[object, int, dict]:
         """Deletes the cron workflow"""
         return self.service.delete_cron_workflow(self.name)
 
-    def update(self) -> "CronWorkflow":
+    def update(self) -> IoArgoprojWorkflowV1alpha1CronWorkflow:
         """Updates the cron workflow in the server"""
 
         # when update cron_workflow, metadata.resourceVersion and metadata.uid should be same as the previous value
@@ -119,8 +118,7 @@ class CronWorkflow(Workflow):
         cron_workflow = self.build()
         cron_workflow.metadata["resourceVersion"] = old_workflow.metadata["resourceVersion"]
         cron_workflow.metadata["uid"] = old_workflow.metadata["uid"]
-        self.service.update_cron_workflow(self.name, cron_workflow)
-        return self
+        return self.service.update_cron_workflow(self.name, cron_workflow)
 
     def suspend(self) -> Tuple[object, int, dict]:
         """Suspends the cron workflow"""
--- a/src/hera/workflow.py
+++ b/src/hera/workflow.py
@@ -258,14 +258,13 @@ class Workflow:
         add_tasks(self, *ts)
         return self
 
-    def create(self) -> "Workflow":
+    def create(self) -> IoArgoprojWorkflowV1alpha1Workflow:
         """Creates the workflow"""
         assert self.dag
         if self.in_context:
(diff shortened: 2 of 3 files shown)


src/hera/cron_workflow.py
```diff
@@ -100,27 +100,25 @@ def build(self) -> IoArgoprojWorkflowV1alpha1CronWorkflow:
             spec=cron_workflow_spec,
         )
 
-    def create(self) -> "CronWorkflow":
+    def create(self) -> IoArgoprojWorkflowV1alpha1CronWorkflow:
         """Creates the cron workflow"""
         if self.in_context:
             raise ValueError("Cannot invoke `create` when using a Hera context")
-        self.service.create_cron_workflow(self.build())
-        return self
```
flaviuvadan, line 108: I think Hera should attempt to deserialize the fields of the returned response - the fields of `IoArgoprojWorkflowV1alpha1CronWorkflow` - into its own `CronWorkflow` representation. This way (1) users get the result enriching the Hera object, which they can continue using for their own purposes, such as switching fields post-submission, then `.create()` the workflow again, and (2) it shields the user from the underlying implementation, which is via the Argo generated SDK. What do you think?

Trollgeir: I'm a bit unsure if I know what you mean. I don't know if there's anything of value returned by creating a cron-workflow besides if it was a success or not (`generate-name` isn't supported for cron workflows). Also the return object is a kubernetes object I believe, so if we would want to parse that back to a cron-workflow we'd have to pick out bits and pieces which are relevant to us. 

Do you have any examples of a use case?

Trollgeir: If generate-name is the only use-case for a return object, we could instead fetch and return only that to the user upon creation if you want to shield the users from the return object.

flaviuvadan: > Also the return object is a kubernetes object I believe, so if we would want to parse that back to a cron-workflow we'd have to pick out bits and pieces which are relevant to us.

This is exactly what I am thinking about! The main advantage I see is that people can do follow up actions such as get the workflow status, delete, etc, without having to recreate the workflow object themselves. For something like generated names they will at least be included in the workflow object now, so people can perform said actions on it. Also, parsing out fields from the returned response is not _that_ bad because Hera can implement a method that parses out these responses and only extracts stuff like the template and the spec. Thoughts?

Trollgeir: Well I'm not going to object doing something like this in the future, but I think that the user should still be able to have the freedom to get everything _raw_ back from the Argo API without us censoring anything.

flaviuvadan: > should still be able to have the freedom to get everything raw back from the Argo API without us censoring anything
The user still has access to raw objects via `build` but I do understand the point is about returned types. My concern is that this might be confusing because the user interacts with Hera via something like `create()` and then the expectation might be to have, further, access to the fields of the workflow post submission. For instance, getting a workflow link from a submitted workflow to issue a notification somewhere. If the user receives back the Argo SDK type then they also have to learn that type's fields, which forces the user to learn the SDK, and that is what Hera attempts to abstract because it's complex. I understand the argument for the freedom to use anything but I think the type of freedom we're discussing here is not very aligned, in my perspective, with Hera's principle of simplicity. The mental toil associated with knowing both Hera and the Argo SDK types will be quite a lot for users, IMO, and knowing Hera should be sufficient to achieve everything. Right now that is not the case, certainly, but with the release of v4, which brings Hera closer to Argo native notation, I now see Hera on an even better path towards making Argo access simpler. What do you think?

> Maybe there are special cluster settings that will be returned that we'd never be able to catch if we filter stuff
Ah, this is interesting. Are there Argo cluster settings to set? I can imagine things like cluster config maps that set the artifact repository (like GCS), the Postgres instance IP, etc, but am not sure whether those are returned as part of some types. Do you know of any?

flaviuvadan: I think https://github.com/argoproj-labs/hera-workflows/issues/277 is a good example of a use case when someone knew they wanted a K8S object and they obtained it via `build`, which is the primary reason `build` was left a public method. I think `build` can be the formal interface that exposes users who are comfortable with the Argo generated SDK to the SDK types but shields other users from the SDK and helps them stay within the boundary of Hera.

Trollgeir:
Closing in favor of #299

PR #283, opened 2022-09-27
Make RetryStrategy more argo native

Trollgeir:
Since the `RetryStrategy` object (refactored from `Retry`) is independent of the rest of the workflow (no hera magic), we can do the build on the object itself. In addition, there were some fields lacking for full argo features which has been added. Should now be a 1:1 with argo.

The only downside is that the user has to create dicts for the nested objects in `RetryStrategy`.


src/hera/retry_strategy.py
```diff
@@ -0,0 +1,53 @@
...
+@dataclass
+class RetryStrategy:
+    """Retry holds the duration values for retrying tasks.
+
+    Attributes
+    ----------
+    affinity: Optional[Dict] = None
+        Affinity prevents running workflow's step on the same host.
+        See: https://argoproj.github.io/argo-workflows/fields/#retryaffinity
+    backoff: Optional[Dict] = None
```
flaviuvadan, line 22: What do you think about a custom `Backoff` object for this? 🤔 Seems like it's pretty simple, with 3 fields only. It would save users from having to reference the Argo docs

Trollgeir: We could, but then I think we should add it everywhere we're currently using dicts as valid inputs.

Trollgeir: Maybe in a future argo version they'll add some more parameters. If we support dict we don't need to update anything.

flaviuvadan (approved):

src/hera/retry_strategy.py
flaviuvadan: I see your perspective. Perhaps we can add it in the future! Also, I want to mention that these things can be introduced gradually. I appreciate that we should add it everywhere, such as to `ValueFrom`, but even that can be introduced one by one! 

Thank you for the discussion 🙂

PR #284, opened 2022-09-28
fix git artifact ssh secrets

JCoder01:
I think this is what was actually intended...

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,7 +114,7 @@ If you plan to submit contributions to Hera you can install Hera in a virtual en
 poetry install
 ```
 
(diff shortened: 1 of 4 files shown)


flaviuvadan (commented):
Thank you for the great catch @JCoder01! 🙂 Any chance the PR can include a small test for this so it's captured in the future?

JCoder01:
@flaviuvadan sure, just pushed an update with a test. It fails on `main` but passes on the branch. I also fixed a very, very minor typo in the readme. If you want that on a separate commit/pr, let me know.

flaviuvadan (approved):
Awesome! This looks great! And thank you for the README adjustment! 🚀

PR #285, opened 2022-09-28
allow unset variables in WorkflowService

### LINKED ISSUE #277: getting typed Kubernetes objects without setting a fake address / port

nikhiljha:
This ended up being really simple, and now works fine for my use case. Let me know what you think!

fixes #277

### DIFF
--- a/src/hera/workflow.py
+++ b/src/hera/workflow.py
@@ -106,7 +106,7 @@ class Workflow:
         active_deadline_seconds: Optional[int] = None,
     ):
         self.name = validate_name(name)
-        self.service = service or WorkflowService()
+        self._service = service
         self.parallelism = parallelism
         self.security_context = security_context
         self.service_account_name = service_account_name
@@ -127,6 +127,16 @@ class Workflow:
         self.active_deadline_seconds = active_deadline_seconds
         self.exit_task: Optional[str] = None
         self.tasks: List["Task"] = []
+
+    @property
+    def service(self) -> WorkflowService:
+        if self._service is None:
+            self._service = WorkflowService()
+        return self._service
+
+    @service.setter
+    def service(self, value: WorkflowService):
+        self._service = value
 
     def get_name(self) -> str:
         """


src/hera/workflow.py
```diff
@@ -264,6 +264,9 @@ def create(self) -> "Workflow":
         if self.in_context:
             raise ValueError("Cannot invoke `create` when using a Hera context")
 
+        if self.service is None:
+            self.service = WorkflowService()
+
         self.service.create_workflow(self.build())
```
flaviuvadan, line 270: Looks like a great start @nikhiljha! Thank you for the PR! I think there might be other places to update as well. Specifically, the `CronWorkflow` and `WorkflowTemplate`. Can the PR include those as well?

nikhiljha: @flaviuvadan Ah okay! Just pushed a new PR that should fix it anywhere (including in future objects), since they're all subclassed on `Workflow`.

flaviuvadan (approved):
This is nice and elegant @nikhiljha! Thank you for the contribution 🚀

PR #292, opened 2022-10-02
Adjust README

flaviuvadan:
@Trollgeir we discussed removing the `Comparison` section. However, I kept it here because I noticed that _most_ open-source projects on GitHub (from the ones I viewed) do not provide a comparison. Providing a comparison helps users understand advantages/disadvantages of their SDK choice, Argo SDK in this case. Thoughts?

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,31 +21,25 @@ and its crew were specially protected by the goddess Hera.
 
 [![License: MIT](https://img.shields.io/badge/License-MIT-yellow.svg)](https://opensource.org/licenses/MIT)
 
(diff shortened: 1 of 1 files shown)


Trollgeir:
> @Trollgeir we discussed removing the `Comparison` section. However, I kept it here because I noticed that _most_ open-source projects on GitHub (from the ones I viewed) do not provide a comparison. Providing a comparison helps users understand advantages/disadvantages of their SDK choice, Argo SDK in this case. Thoughts?

I think the comparison I feel we should sell isn't Hera vs the other SDKs, rather Hera vs YML.

Trollgeir (approved):

PR #299, opened 2022-10-03
Add generated name upon creation

Trollgeir:
If the `generate_name` is set on the workflow, the resulting name generated by the argo server upon creation will be put in `generated_name`.


flaviuvadan (approved):
:rocket:

PR #307, opened 2022-10-05
Fix typo in Resources

Trollgeir:


flaviuvadan (approved):
Why did `ResourceRequirements` not catch it? 🤔

Trollgeir:
The model isn't strict with checking the name of the fields.

PR #309, opened 2022-10-05
Add `DAG` in workflow init 

flaviuvadan:


Trollgeir (approved):

src/hera/dag.py
```diff
@@ -122,15 +121,20 @@ def __exit__(self, exc_type, exc_val, exc_tb) -> None:
 
     def add_task(self, t: Task) -> "DAG":
         """Add a task to the DAG. Note that tasks are added automatically when the DAG context is used"""
-        add_task(self, t)
+        if t is None:
+            return self
```
JacobHayes, line 125: `t` is not marked `Optional`, so this should never happen. Or, does the hint need to be `: Optional[Task]`? I don't think one should be allowed to pass in a `None`.

Similar in `add_tasks` (also, could that one just call `self.add_task` inside the loop?)

src/hera/workflow.py
```diff
@@ -283,12 +278,14 @@ def __exit__(self, exc_type, exc_val, exc_tb) -> None:
 
     def add_task(self, t: Task) -> "Workflow":
         """Add a task to the workflow"""
-        add_task(self, t)
+        assert self.dag is not None, "A `DAG` must be defined when adding a task to a workflow"
```
JacobHayes, line 281: We may not need these now that we always set a value in `__init__`.

If this was to make `mypy` happy (ie: it thought `self.dag` was an `Optional[DAG]`), I think we can add a type hint to the original `self.dag = ` to be `self.dag: DAG = ...` (I'd expect mypy to infer that, but I think I've also seen issues with it's type narrowing when using ternaries).

JacobHayes (commented):
Thanks for the fixes! A couple quick comments

tests/test_dag.py
```diff
@@ -154,3 +154,10 @@ def test_get_parameter_raises_on_param_not_found(self):
         with pytest.raises(KeyError) as e:
             dag.get_parameter("a")
         assert str(e.value) == "'Could not assemble a parameter as `a` is not a DAG output'"
+
+    def test_adds(self):
+        dag = DAG("test").add_tasks(Task('t1'), Task('t2'))  # type: ignore
```
JacobHayes, line 159:
```suggestion
        dag = DAG("test").add_tasks(Task('t1'), Task('t2'))
```

PR #311, opened 2022-10-05
Lower coverage precision to reduce jitter

JacobHayes:
Codecov was mad in https://github.com/argoproj-labs/hera-workflows/pull/309 because some source code was removed (thus "no longer covered"), causing some slight (negative) differences in the division. That should not matter (ideally, we'd be incentivised to remove code :), so this PR reduces the % diff decimal precision from 2 places to 1, which would have resulted in that PR passing, even without the additional `asserts`.

### DIFF
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -1,0 +1,2 @@
+coverage:
+  precision: 1  # Don't bother with very small changes, which are often caused by subtle division differences when SLOC are removed.
--- a/tox.ini
+++ b/tox.ini
@@ -14,6 +14,7 @@ exclude_lines =
     if TYPE_CHECKING:
     pragma: no cover
     raise NotImplementedError()
+precision = 1
 skip_covered = True
 show_missing = True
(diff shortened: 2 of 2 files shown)


flaviuvadan (approved):
🚀

JacobHayes:
Hmm, the comment still seems to show 2 decimal places - maybe it didn't pick up the config (or that doesn't follow it)? I'll make another PR branching off this that should "reduce" coverage.

JacobHayes:
The precision [didn't work](https://github.com/argoproj-labs/hera-workflows/pull/312), will poke around a bit more.

JacobHayes:
I wonder if codecov uses the config from the default branch, not the current branch (similar to how GH Actions run from the base branch). If you're comfortable, do you mind merging and I can test/iterate in https://github.com/argoproj-labs/hera-workflows/pull/312 if needed (regardless of whether we close that, will be useful to test)?

PR #312, opened 2022-10-05
Cleanup Workflow.dag checks

JacobHayes:
https://github.com/argoproj-labs/hera-workflows/pull/309 added some extra `assert self.dag is not None` to `Workflow`, even though `self.dag` is now _always_ set/required. The added `test_workflow_catches_none_dag_upon_task_insert` test checks for this, but this is a degenerate case anyway, as `mypy` shows:

```
$ mypy -p hera -p tests
... # lots of other errors in tests, these aren't normally checked
tests/test_workflow.py:373: error: Incompatible types in assignment (expression has type "None", variable has type "DAG")
...
```

With https://github.com/argoproj-labs/hera-workflows/pull/311, we don't have to worry as much about codecov/coverage jitter, so the extra checks+test can be removed without making codecov mad. I figured making the code a bit more clear (ie: not make you second guess your mental model of `self.dag`) might be good, but feel free to close this if you still prefer the checks!

### DIFF
--- a/src/hera/workflow.py
+++ b/src/hera/workflow.py
@@ -181,7 +181,6 @@ class Workflow:
         """Assembles the spec of the workflow"""
         # Main difference between workflow and workflow template spec is that WT
         # (generally) doesn't have an entrypoint
-        assert self.dag is not None
         spec = IoArgoprojWorkflowV1alpha1WorkflowSpec()
         templates = self.dag._build_templates()
 
@@ -278,19 +277,16 @@ class Workflow:
 
     def add_task(self, t: Task) -> "Workflow":
         """Add a task to the workflow"""
-        assert self.dag is not None, "A `DAG` must be defined when adding a task to a workflow"
         self.dag.add_task(t)
         return self
 
     def add_tasks(self, *ts: Task) -> "Workflow":
         """Add a collection of tasks to the workflow"""
-        assert self.dag is not None, "A `DAG` must be defined when adding tasks to a workflow"
         self.dag.add_tasks(*ts)
         return self
 
     def create(self) -> "Workflow":
         """Creates the workflow"""
-        assert self.dag
         if self.in_context:
             raise ValueError("Cannot invoke `create` when using a Hera context")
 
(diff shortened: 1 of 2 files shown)


JacobHayes:
I couldn't set the base branch to be https://github.com/argoproj-labs/hera-workflows/pull/311/files (since that branch is from a fork), but can rebase or something after/if that is merged.

Trollgeir:
If the `dag` member should never be directly altered after construction, should we then make it private? As in, `_dag`.

JacobHayes:
I tend to prefer public members, particularly for things (like `DAG`) that are objects a user may construct and pass in, but would happily defer. 😄

I guess `self._service` does follow a private precedent, but also provides a `@property` getter and setter (so, could still be modified). Does `WorkflowService()` make network connections or why do we defer that to the `@property` rather than doing `self.service = WorkflowService() if service is None else service` in `Workflow.__init__`?

Trollgeir:
> I guess `self._service` does follow a private precedent, but also provides a `@property` getter and setter (so, could still be modified). Does `WorkflowService()` make network connections or why do we defer that to the `@property` access rather than doing `self.service = WorkflowService() if service is None else service` in `Workflow.__init__`?

It's related to https://github.com/argoproj-labs/hera-workflows/pull/285

flaviuvadan (approved):
LGTM! Thank you for addressing this and some of the other mypy related problems :)

JacobHayes:
Thanks for the fix in https://github.com/argoproj-labs/hera-workflows/pull/317 @flaviuvadan (and 'doh haha), I rebased this so coverage should pass now!

JacobHayes:
Haha, and now the good old codecov upload flake:
> There was an error running the uploader: Error uploading to [https://codecov.io:](https://codecov.io/) Error: There was an error fetching the storage URL during POST: 404 - {'detail': ErrorDetail(string='Unable to locate build via Github Actions API. Please upload with the Codecov repository upload token to resolve issue.', code='not_found')}

Maybe retry [that step](https://github.com/argoproj-labs/hera-workflows/actions/runs/3283413251/jobs/5408138985)?

PR #313, opened 2022-10-06
Bugfix: env value_from_input was ignored

tmi:
In case of `Env` being specified via `value_from_input`, the actual value of the `value_from_input` is ignored and just its not-`None`ness is checked. We fix it here to use it. Current buggy state has the unfortunate consequence of having to name the env var the same as the parameter

### DIFF
--- a/src/hera/env.py
+++ b/src/hera/env.py
@@ -61,10 +61,10 @@ class Env:
     @staticmethod
     def _sanitise_param_for_argo(v: str) -> str:
         """Argo has some strict parameter validation. To satisfy, we replace all ._ with a dash,
-        take only first 32 characters from a-zA-Z-, and append md5 digest of the original string."""
+        take only first 32 characters from a-zA-Z0-9-, and append md5 digest of the original string."""
         # NOTE move this to some general purpose utils?
-        replaced_dashes = v.translate(str.maketrans({e: '-' for e in "_."})) # type: ignore
-        legit_set = string.ascii_letters + '-'
+        replaced_dashes = v.translate(str.maketrans({e: '-' for e in "_."}))  # type: ignore
+        legit_set = string.ascii_letters + string.digits + '-'
         legit_prefix = "".join(islice((c for c in replaced_dashes if c in legit_set), 32))
         hash_suffix = hashlib.md5(v.encode("utf-8")).hexdigest()
         return f"{legit_prefix}-{hash_suffix}"
--- a/src/hera/env.py
+++ b/src/hera/env.py
@@ -1,5 +1,8 @@
+import hashlib
 import json
+import string
 from dataclasses import dataclass
+from itertools import islice
 from typing import Any, Optional, Union
 
 from argo_workflows.models import (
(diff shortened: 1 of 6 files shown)


tmi:
additionally, would you mind if I add `value_from_param`, which would make the var resolve to `f"{{{{workflow.parameters.{param_name}}}}}"` instead? That's something I use quite often as well

tmi:
@Trollgeir thank you for your message, and apologies for my late reply, and what appears to be a hasty and misguided PR O:)

my original reason for opening the PR was that the Docker images I'm scheduling are expecting some Environment variables with dots in their name -- which is troublesome here, since the current code seems to use the same name for env var and for the param name, and thus a dot in the name gets refused by Argo param name validation

I can't objectively call it a bug, and I can do a workaround via `Env(name=dot_param_name), value=f"{{{{workflow.parameters.{sanitize(dot_param_name)}}}}}")`.

However, I'd be happier if I could specify it in a less `{{{{}}}}`-manner, but not sure if you guys will think a fix systemic. I can just add some `_sanitize_for_argo` function that replaces dots with underscore and call it in `env.py:L61` and `task.py:L710`. Does that look legit?

If I get what you say and what the code suggests, I can just do this for global params, right?:
```
global_param = Param(name=sanitize(dot_param_name), value="<something>")
env = Env(name=dot_param_name, value_from_input=global_param)
create_worklflow(..., global_param, ..., create_task(..., env, ...))
```
I then agree it makes no sense to add `value_from_workflow_param`. However, since I'll be touching it, would you mind if I add a dataclass postinit auto validation that complaints if both `value` and `value_from_input` is specified? That would seem wrong use to me, right?

Trollgeir:
> my original reason for opening the PR was that the Docker images I'm scheduling are expecting some Environment variables with dots in their name -- which is troublesome here, since the current code seems to use the same name for env var and for the param name, and thus a dot in the name gets refused by Argo param name validation

Ah, nice catch! Yeah this is something I hadn't thought of, and obviously we need to fix. 

> I can't objectively call it a bug, and I can do a workaround via `Env(name=dot_param_name), value=f"{{{{workflow.parameters.{sanitize(dot_param_name)}}}}}")`.
> 
Great, I hope this will suffice until we get a fix out!

> However, I'd be happier if I could specify it in a less `{{{{}}}}`-manner, but not sure if you guys will think a fix systemic. I can just add some `_sanitize_for_argo` function that replaces dots with underscore and call it in `env.py:L61` and `task.py:L710`. Does that look legit?

I think that there's an easier workaround than sanitizing the fields. The generated parameters which are used to feed-forward the envs can be called whatever we want. In my initial implementation I just defaulted them to have the same name as the env-variable. However, I _think_ we could possibly just generate some argo-friendly hash values for names, like `env-variable-<generated-hash>`. Hash could also be a checksum based on the env-name, which guarantees uniqueness. What do you think?

tmi:
hmmm, the hash thing may be inconvenient for any sort of debugging/troubleshooting -- eg, when you are looking at some workflow that has failed overnight, you often check out the parameters. I always log all the input configuration at the start of the container, so I would not be affected; but I guess some users may be inspecting the params, and this makes it a bit inconvenient

but from implementation PoV/reliability, its definitely a good idea and I like it :) I'd lean toward `hash(env_name)` than toward say `uuid4`, to have at least stability in the param names.

maybe a good compromise would be to take first ~32 characters of env, remove all that Argo doesnt like, and append the hash of the original value. This way, you get around the total length limitation and bad characters, you retain some readability, but you dont risk introducing a collision by sanitisation

Trollgeir:
> hmmm, the hash thing may be inconvenient for any sort of debugging/troubleshooting -- eg, when you are looking at some workflow that has failed overnight, you often check out the parameters. I always log all the input configuration at the start of the container, so I would not be affected; but I guess some users may be inspecting the params, and this makes it a bit inconvenient. (The output parameters / global parameters would still have ok names -- but still, it could be nuissance a bit)

Wouldn't it be fairly obvious what ENV it's referencing in the value though?

> maybe a good compromise would be to take first ~32 characters of env, remove all that Argo doesnt like, and append the hash of the original value. This way, you get around the total length limitation and bad characters, you retain some readability, but you dont risk introducing a collision by sanitisation

We could, we just need to implement the sanitation function :)

tmi:
I've put together a simple sanitisation function and used it in this case, let me know if you'd rather see it differently :)

Trollgeir (approved):
This looks really great!

I'm sure we will fragment this code later (the md5 for example) and organize differently, but no need to do so prematurely.

src/hera/env.py
```diff
@@ -55,10 +58,33 @@ class Env:
     value: Optional[Any] = None
     value_from_input: Optional[Union[str, Parameter]] = None
 
+    @staticmethod
+    def _sanitise_param_for_argo(v: str) -> str:
+        """Argo has some strict parameter validation. To satisfy, we replace all ._ with a dash,
+        take only first 32 characters from a-zA-Z-, and append md5 digest of the original string."""
+        # NOTE move this to some general purpose utils?
+        replaced_dashes = v.translate(str.maketrans({e: '-' for e in "_."})) # type: ignore
+        legit_set = string.ascii_letters + '-'
```
flaviuvadan, line 67: IIRC, the naming convention followed by Argo follows the one of K8S, and K8S uses the specification of RFC1123. That said, do you happen to know whether the parameter name follows the same convention? If so, perhaps `0-9` can be part of the legit set?

vojtechtuma-imi: good point, you are right, numbers are indeed tolerated (even at the start of the name). So I've just extended the legit set and one of the test cases.

flaviuvadan (approved):
Thank you for the contribution @tmi! 🚀 and kudos for your first one! I would love to hear your thoughts on other Hera issues 🙂 

P.S: I will merge without the codecov check because we recently adjusted on `main` so that it uses single digit precision, so the `codecov/project` would accept a `0.07%` change

PR #317, opened 2022-10-19
add CodeCov round down to 1 precision

flaviuvadan:
CC: @JacobHayes


JacobHayes (approved):
Aaah, thank you! I missed that there was a root `codecov.yml` when working on https://github.com/argoproj-labs/hera-workflows/pull/311/files, I think we can remove `.github/codecov.yml` now! 🚀

PR #320, opened 2022-10-21
Fix dag template not added for WFTemplates

Trollgeir:
A bug caused the DAG template to be omitted from `WorkflowTemplate`

In addition, I've made it possible to specify what the name of the dag template should be, which before always defaulted to the name of the workflow.

### DIFF
--- a/src/hera/workflow.py
+++ b/src/hera/workflow.py
@@ -113,7 +113,7 @@ class Workflow:
         metrics: Optional[Union[Metric, List[Metric], Metrics]] = None,
     ):
         self.name = validate_name(name)
-        dag_name = dag_name if dag_name is not None else self.name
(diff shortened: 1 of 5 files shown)


src/hera/workflow.py
```diff
@@ -110,6 +113,8 @@ def __init__(
         metrics: Optional[Union[Metric, List[Metric], Metrics]] = None,
     ):
         self.name = validate_name(name)
+        dag_name = dag_name if dag_name is not None else self.name
```
flaviuvadan, line 116: This might be easier to understand
```suggestion
        dag_name = self.name is dag_name is None else dag_name
```

flaviuvadan (approved):

PR #324, opened 2022-10-25
Use >= for version constraints

JacobHayes:
When adding hera-workflows v4 to a project and relocking, I noticed that the version of `certifi` was dropped from "2022.9.24" to "2021.10.8". This was caused by the `pyproject.toml` using `~` for version constraints (the poetry default...) rather than `>=`, causing calver packages to be kept quite outdated (similar for pytz). I changed the deps to use `>=` instead.

Here's a [decent blog post](https://iscinumpy.dev/post/bound-version-constraints/#tldr) about preferring to avoid upper bound caps, particularly in libraries.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -19,18 +19,18 @@ optional = false
 python-versions = ">=3.5"
 
 [package.extras]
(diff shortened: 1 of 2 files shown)


flaviuvadan (approved):
Ty @JacobHayes 🚀

Issue #316, opened 2022-10-14
Better support for GitOps & generating YAML


valorl:
Hi, 

For e.g. a CronWorkflow, it's possible, and quite useful to manage it with GitOps. However, it doesn't seem like Hera's API supports this very well. The API assumes imperative submission of resources with the `create()` method. It would be great to be able to simply generate a fully valid Kubernetes manifest instead of directly talking to the API.

I've tried using `build()` instead of `create()` and serializing it myself with pyyaml, however that payload is not fully valid, e.g.:
- props use snake_case instead of camelCase
- The `apiVersion` and `kind` fields are missing

Other than "patching" this payload myself, are there any alternative APIs in Hera and/or the Argo Python SDK that could help me get a valid Kubernetes manifest out of a `hera.cron_workflow.CronWorkflow` instance ? If not, consider this a feature request.

Trollgeir:
We've considered making this a feature (we've had similar request). However, I was also a bit worried that implementing it would make consumers stray a way from the intended path of applying these CronWorkflows/Templates directly with hera -> argo API. 

I'm also curious if you'd be willing to share how you've set up your CI/CD. Do you intend of resolving the YML-file locally before committing it to a repository, where the CI/CD picks up on it with kubernetes? Or do one of your steps run the python code generating the yml file, and you want to forward it to kubernetes? If it's the latter of these, what holds you back from doing `.create()` at the end?

valorl:
We use GitOps and FluxCD heavily to be able to spin clusters up and down without having to re-run CI and re-apply things.
> Do you intend of resolving the YML-file locally before committing it to a repository, where the CI/CD picks up on it with kubernetes? Or do one of your steps run the python code generating the yml file, and you want to forward it to kubernetes?

It could be either of these approaches. For our more standard K8S resources, we actually generate them in CI and then ship them to a Git repo. Then whenever we spin up a cluster, it auto-bootstraps, installing Flux pointed at that repo, and applying (continuously) all the manifests in it. 

> If it's the latter of these, what holds you back from doing `.create()` at the end?

Doing a `.create()` is basically the equivalent of doing a `kubectl apply` in your CI. The main issue with this is that it's hard to guarantee that the state of the manifest remains as it was at the time of the create/apply. If you instead put the manifest in a Git repo and use a GitOps operator like Flux or ArgoCD, it will always be reconciled back to the state stored in Git. And as I mentioned, it also allows us to bootstrap new clusters quickly and robustly.

[...]

PR #327, opened 2022-10-26
Add YAML and dict support for workflows

flaviuvadan:
Fixes: #316 

CC: @valorl

### DIFF
--- a/examples/memoize.py
+++ b/examples/memoize.py
@@ -12,8 +12,8 @@ def consume(value):
 
 # assumes you used `hera.set_global_token` and `hera.set_global_host` so that the workflow can be submitted
 with Workflow("memoize") as w:
-    g = Task("g", generate, outputs=[Parameter("out", value_from=ValueFrom(path="/out"))])
-    c = Task("c", consume, inputs=[g.get_parameter("out")], memoize=Memoize("value", "memoize", "c"))
+    g = Task("g", generate, outputs=[Parameter("value", value_from=ValueFrom(path="/out"))])
+    c = Task("c", consume, inputs=[g.get_parameter("value")], memoize=Memoize("value", "memoize", "c"))
     g >> c
 
 w.create()
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,6 +1,6 @@
 [mypy]
 check_untyped_defs = True
-mypy_path=src
+mypy_path = src
 no_implicit_optional = True
 strict_optional = True
 
@@ -11,7 +11,4 @@ ignore_missing_imports = True
 ignore_missing_imports = True
 
 [mypy-pytz.*]
-ignore_missing_imports = True
-
-[mypy-yaml.*]
 ignore_missing_imports = True
--- a/mypy.ini
+++ b/mypy.ini
@@ -12,3 +12,6 @@ ignore_missing_imports = True
 
 [mypy-pytz.*]
 ignore_missing_imports = True
+
+[mypy-yaml.*]
+ignore_missing_imports = True
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -32,6 +32,7 @@ python-dateutil = ">=2.8.2"
 urllib3 = ">=1.26.8"
 certifi = ">=2021.10.8"
 pytz = ">=2021.3"
+PyYAML = ">=6.0"
 argo-workflows = { version = "6.3.5", allow-prereleases = true }
 
 [tool.poetry.dev-dependencies]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -32,11 +32,11 @@ python-dateutil = ">=2.8.2"
 urllib3 = ">=1.26.8"
 certifi = ">=2021.10.8"
 pytz = ">=2021.3"
-PyYAML = ">=6.0"
+pyyaml = { version = ">=6.0", optional = true }
 argo-workflows = { version = "6.3.5", allow-prereleases = true }
 
 [tool.poetry.extras]
-yaml = ["PyYAML>=6.0"]
+yaml = ["PyYAML"]
 
 [tool.poetry.dev-dependencies]
 pytest = "*"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,6 +35,9 @@ pytz = ">=2021.3"
 PyYAML = ">=6.0"
 argo-workflows = { version = "6.3.5", allow-prereleases = true }
 
+[tool.poetry.extras]
+yaml = ["PyYAML>=6.0"]
+
 [tool.poetry.dev-dependencies]
 pytest = "*"
 pytest-cov = "*"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,10 +28,10 @@ packages = [
 
 [tool.poetry.dependencies]
 python = ">=3.7,<4"
-python-dateutil = "^2.8.2"
-urllib3 = "^1.26.8"
-certifi = "^2021.10.8"
-pytz = "^2021.3"
+python-dateutil = ">=2.8.2"
+urllib3 = ">=1.26.8"
+certifi = ">=2021.10.8"
+pytz = ">=2021.3"
 argo-workflows = { version = "6.3.5", allow-prereleases = true }
 
 [tool.poetry.dev-dependencies]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,6 +44,7 @@ mypy = "*"
 build = "*"
 isort = "*"
 tox = "<4.0"
+types-PyYAML = "*"
 
 [build-system]
 requires = ["poetry-core>=1.0.0"]
--- a/src/hera/cron_workflow.py
+++ b/src/hera/cron_workflow.py
@@ -8,6 +8,7 @@ from argo_workflows.models import (
     IoArgoprojWorkflowV1alpha1CronWorkflowSpec,
 )
 
+from hera.host_config import get_global_api_version
 from hera.workflow import Workflow
 
 
@@ -96,8 +97,8 @@ class CronWorkflow(Workflow):
             setattr(cron_workflow_spec, "timezone", self.timezone)
 
         return IoArgoprojWorkflowV1alpha1CronWorkflow(
-            api_version="argoproj.io/v1alpha1",
-            kind="CronWorkflow",
+            api_version=get_global_api_version(),
+            kind=self.__class__.__name__,
             metadata=workflow.metadata,
             spec=cron_workflow_spec,
         )
--- a/src/hera/cron_workflow.py
+++ b/src/hera/cron_workflow.py
@@ -96,6 +96,8 @@ class CronWorkflow(Workflow):
             setattr(cron_workflow_spec, "timezone", self.timezone)
 
         return IoArgoprojWorkflowV1alpha1CronWorkflow(
+            api_version="argoproj.io/v1alpha1",
+            kind="CronWorkflow",
             metadata=workflow.metadata,
             spec=cron_workflow_spec,
         )
--- a/src/hera/workflow.py
+++ b/src/hera/workflow.py
@@ -3,6 +3,7 @@ import re
 from typing import Callable, Dict, List, Optional, Tuple, Union
 
 import yaml
+from argo_workflows.model_utils import model_to_dict
 from argo_workflows.models import (
     IoArgoprojWorkflowV1alpha1Arguments,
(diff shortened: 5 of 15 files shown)


examples/memoize.py
```diff
@@ -12,8 +12,8 @@ def consume(value):
 
 # assumes you used `hera.set_global_token` and `hera.set_global_host` so that the workflow can be submitted
 with Workflow("memoize") as w:
-    g = Task("g", generate, outputs=[Parameter("out", value_from=ValueFrom(path="/out"))])
```
flaviuvadan, line 15: @Trollgeir I used this as a test example for YAML export and noticed it was broken because the parameter names did not match, so adjusted it here

src/hera/workflow_template.py
```diff
@@ -19,7 +19,9 @@ class WorkflowTemplate(Workflow):
     def build(self) -> IoArgoprojWorkflowV1alpha1WorkflowTemplate:
         """Builds the workflow"""
         spec = super()._build_spec()
-        return IoArgoprojWorkflowV1alpha1WorkflowTemplate(metadata=self._build_metadata(), spec=spec)
+        return IoArgoprojWorkflowV1alpha1WorkflowTemplate(
+            api_version="argoproj.io/v1alpha1", kind="Workflow", metadata=self._build_metadata(), spec=spec
```
valorl, line 23:
```suggestion
            api_version="argoproj.io/v1alpha1", kind="WorkflowTemplate", metadata=self._build_metadata(), spec=spec
```

src/hera/workflow.py
```diff
@@ -320,3 +327,40 @@ def get_parameter(self, name: str) -> Parameter:
...
+
+    def to_dict(self) -> dict:
+        """Returns the JSON/dictionary representation of the workflow"""
+        return self.build().to_dict()
+
+    def to_yaml(self) -> str:
+        """Returns a YAML representation of the workflow"""
+        dict_repr = self.to_dict()
+        new_dict_repr: dict = {}
+        _traverse(dict_repr, new_dict_repr, _camel_case)
```
valorl, line 339: This will unfortunately also camel-case keys that should be preserved. For example keys in `metadata.labels`, `metadata.annotations`, `spec.nodeSelector` and many more. In general terms, I think we need to preserve keys on generic dictionaries and only camel-case keys that are actual object fields. I'm not much of a Python developer, so not sure about the implementation.

Trollgeir:
Thanks a lot for going through this @valorl , you put my review to shame! :)

src/hera/workflow.py
flaviuvadan: That is a very good point, thank you for bringing it up. I will adjust the implementation!

src/hera/cron_workflow.py
```diff
@@ -96,6 +96,8 @@ def build(self) -> IoArgoprojWorkflowV1alpha1CronWorkflow:
             setattr(cron_workflow_spec, "timezone", self.timezone)
 
         return IoArgoprojWorkflowV1alpha1CronWorkflow(
+            api_version="argoproj.io/v1alpha1",
+            kind="CronWorkflow",
```
elliotgunton, line 100: I would suggest a package level constant `API_VERSION = "argoproj.io/v1alpha1"`.  And for `kind` downstream we considered using `workflow.__class__.__name__` but this might be a bit brittle, so a `self.kind` class member variable would be great.

flaviuvadan: Thank you for the input @elliotgunton! Perhaps we can set the API version the [same way we set the global host and token](https://github.com/argoproj-labs/hera-workflows/blob/main/src/hera/host_config.py)? I think the thought of using `workflow.__class__.__name__` is great as long as Hera will exclusively use class names that are conceptually equivalent to Argo Workflows. So in the context of `Workflow`, `WorkflowTemplate`, and `CronWorkflow` this holds. Perhaps we can start with your suggestion of using the class name and then introduce a `kind` concept later if need be?

src/hera/workflow.py
samj1912, line 339: There is actually a function in the Argo sdk that already contains this logic. See https://github.com/argoproj/argo-workflows/blob/6b086368f6480a2de5e2d43eec73514de0ad01ac/sdks/python/client/argo_workflows/model_utils.py#L1648

when serialize it set to true, it correctly uses the model attribute map to convert python attributes to json/yaml ones.

flaviuvadan:
@elliotgunton, @valorl, @samj1912 I incorporated a combination of your collective feedback. Let me know what you think!

src/hera/workflow.py
flaviuvadan: Ended up going with @samj1912 suggestion and now fields like labels, annotations, and selectors are correctly ignored while everything else is parsed

src/hera/workflow.py
```diff
@@ -320,3 +328,12 @@ def get_parameter(self, name: str) -> Parameter:
         if self.parameters is None or next((p for p in self.parameters if p.name == name), None) is None:
             raise KeyError(f"`{name}` is not a valid workflow parameter")
         return Parameter(name, value=f"{{{{workflow.parameters.{name}}}}}")
+
+    def to_dict(self) -> dict:
```
samj1912, line 332: Should this take an optional `serialize` kwarg?

mypy.ini
```diff
@@ -12,3 +12,6 @@ ignore_missing_imports = True
 
 [mypy-pytz.*]
 ignore_missing_imports = True
+
+[mypy-yaml.*]
+ignore_missing_imports = True
```
samj1912, line 17: pyyaml has a types package at https://pypi.org/project/types-PyYAML/

We can potentially include it in dev requirements?

src/hera/workflow.py
```diff
@@ -320,3 +328,12 @@ def get_parameter(self, name: str) -> Parameter:
...
+        """Returns the JSON/dictionary representation of the workflow"""
+        return self.build().to_dict()
+
+    def to_yaml(self) -> str:
+        """Returns a YAML representation of the workflow"""
+        dict_repr = model_to_dict(self.build())
+        return yaml.dump(dict_repr)
```
samj1912, line 339: Should we instead return json? That way we won't need a dependency on pyyaml and all json documents can be submitted to k8s directly anyway (since json is a strict subset of yaml?)

samj1912: If users need it they can dump the output json to yaml on their own?

pyproject.toml
```diff
@@ -32,6 +32,7 @@ python-dateutil = ">=2.8.2"
 urllib3 = ">=1.26.8"
 certifi = ">=2021.10.8"
 pytz = ">=2021.3"
+PyYAML = ">=6.0"
```
samj1912, line 35: Should we make this an optional dependency? If pyyaml is available we can enable the to_yaml method otherwise we can throw an exception asking users to install `hera-workflows[yaml]`. Additionally maybe we can add a to_json method?

src/hera/workflow.py
flaviuvadan: The `to_dict` method on the built object does not accept the serialize, but the internal call to the `argo_workflows` package serializer does, so we can expose that as an optional parameter through that call if any users would like to use that

flaviuvadan: I noticed the suggestion to add a `to_json`, which I quite like, so will skip adding the serialize kwarg and instead expose the `to_json` method

mypy.ini
flaviuvadan: Thanks for sharing! Did not know about the package :) Will evaluate in the context of the optional dependency and add it maybe

src/hera/workflow.py
flaviuvadan: Personally, I am a fan of the idea of not having the YAML dependency (I avoided adding YAML support to Hera for a while now) but issues like [this one](https://github.com/argoproj-labs/hera-workflows/issues/316) have surfaced before as well. I think there _is_ community appetite to use Hera in a more GitOps fashion, which is the primary factor that influenced the decision to accept the introduction of the YAML dependency. However, I certainly think there's immense value to optionality and access to "control knobs" that dictate whether one installs a dependency or not, so I will update the dependencies to make YAML optional. I have a hunch that the part of our community that does use GitOps + Hera likely _already_ has the YAML dependency 

@valorl would love to hear what you think of this!

pyproject.toml
flaviuvadan: See comment on optionality [here](https://github.com/argoproj-labs/hera-workflows/pull/327#discussion_r1007532840) and [this one](https://github.com/argoproj-labs/hera-workflows/pull/327#discussion_r1007529489) for `to_json`!

src/hera/workflow.py
```diff
@@ -320,3 +338,20 @@ def get_parameter(self, name: str) -> Parameter:
...
             raise KeyError(f"`{name}` is not a valid workflow parameter")
         return Parameter(name, value=f"{{{{workflow.parameters.{name}}}}}")
+
+    def to_dict(self) -> dict:
+        """Returns the dictionary representation of the workflow"""
+        return model_to_dict(self.build(), serialize=False)
+
+    def to_json(self) -> str:
+        """Returns the JSON representation of the workflow"""
+        return json.dumps(model_to_dict(self.build(), serialize=False))
```
samj1912, line 348:
```suggestion
        return json.dumps(model_to_dict(self.build(), serialize=True))
```

Should this be true?

src/hera/workflow.py
```diff
@@ -320,3 +338,20 @@ def get_parameter(self, name: str) -> Parameter:
...
+
+    def to_yaml(self) -> str:
+        """Returns a YAML representation of the workflow"""
+        if _yaml is None:
+            raise ImportError(
+                "Attempted to use `to_yaml` but PyYAML is not available. "
+                "Install `hera-workflow[yaml]` to install the extra dependency"
```
samj1912, line 355:
```suggestion
                "Install `hera-workflows[yaml]` to install the extra dependency"
```

src/hera/workflow.py
flaviuvadan: Yes, added that and adjusted the test!

src/hera/workflow.py
valorl: For our usage, I don't think it's very likely that pyyaml would already be installed for other purposes. So from that perspective, it would be certainly convenient. But I can understand not wanting the dependency. I would always want YAML in my GitOps, but that's my preference so I'm fine with having the dependency on my end. Not sure how generally popular that preference is.

valorl:
LGTM :)

src/hera/workflow.py
```diff
@@ -320,3 +338,20 @@ def get_parameter(self, name: str) -> Parameter:
         if self.parameters is None or next((p for p in self.parameters if p.name == name), None) is None:
             raise KeyError(f"`{name}` is not a valid workflow parameter")
         return Parameter(name, value=f"{{{{workflow.parameters.{name}}}}}")
+
+    def to_dict(self) -> dict:
+        """Returns the dictionary representation of the workflow"""
+        return model_to_dict(self.build())
```
samj1912, line 344:
```suggestion
    def to_dict(self, serialize=True) -> dict:
        """Returns the dictionary representation of the workflow"""
        return model_to_dict(self.build(), serialize=serialize)
```

src/hera/workflow.py
```diff
@@ -320,3 +338,20 @@ def get_parameter(self, name: str) -> Parameter:
...
+
+    def to_json(self) -> str:
+        """Returns the JSON representation of the workflow"""
+        return json.dumps(model_to_dict(self.build()))
```
samj1912, line 348:
```suggestion
        return json.dumps(self.to_dict())
```

src/hera/workflow.py
```diff
@@ -320,3 +338,20 @@ def get_parameter(self, name: str) -> Parameter:
...
+
+    def to_yaml(self) -> str:
+        """Returns a YAML representation of the workflow"""
+        if _yaml is None:
+            raise ImportError(
+                "Attempted to use `to_yaml` but PyYAML is not available. "
+                "Install `hera-workflows[yaml]` to install the extra dependency"
+            )
+        return _yaml.dump(model_to_dict(self.build()))
```
samj1912, line 357:
```suggestion
        return _yaml.dump(self.to_dict())
```

samj1912 (approved):

samj1912:
LGTM - Thanks for this PR. We also needed this! :)

PR #328, opened 2022-10-27
Add global verify SSL and K8S SA

flaviuvadan:
@Trollgeir you mentioned the global service account at some point and the verify SSL came up in conversations with @JacobHayes!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -75,6 +75,8 @@ instance/
 
 # Sphinx documentation
 docs/_build/
+docs/_static/
+docs/_templates/
 
 # PyBuilder
 .pybuilder/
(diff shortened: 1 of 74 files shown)


Trollgeir (approved):
LGTM

For future versions, perhaps we should just bundle everything into a global config object holding the dict to all these fields?

JacobHayes (approved):
I only reviewed https://github.com/argoproj-labs/hera-workflows/pull/328/commits/3479a38253c917e6af878f3f6cf1f380533d387b, but that looks good to me!

+1 to having a global config holding more of these (perhaps even a `Config` instance, though that doesn't currently track the service account).

Issue #335, opened 2022-10-28
Add a lint option


samj1912:
Argo exposes a lint end point. For eg. the lint end point is available at https://github.com/argoproj/argo-workflows/blob/6b086368f6480a2de5e2d43eec73514de0ad01ac/sdks/python/client/argo_workflows/api/workflow_template_service_api.py#L720

This is very useful when verifying that the output template/parameter is syntactically valid rather than just being structurally valid.

We should try and provide a `workflow.lint()` method to allow users to easily test their workflows/templates before dumping/creating them.

PR #337, opened 2022-10-28
Add `lint` to workflow, cron workflow, and workflow template

flaviuvadan:
Addresses: #335 

CC: @samj1912


flaviuvadan:
Coverage is expected to drop with this because of the way the service is tested, going to overwrite the coverage percentage check

PR #340, opened 2022-10-29
Add full implementation for the expr transpilation module

### LINKED ISSUE #336: Investigate python -> expr/argo expression language converter
Currently one of the biggest pain points WRT argo yamls is the expression syntax and manually typing that.

If we could somehow provide a nice wrapper that transpiles python code -> argo expression syntax that would be amazing and would help a lot.

Could be something like

```python
import hera.expr as f

# f.C converts the input to a constant literal
out = f.C(42) + f.C(32)
out.expression() == "42 + 32"

out2 = f.C([1, 2, 3]).map(F.mul(2))

out2.expression = "map([1, 2, 3], { # * 2 })"

# alternatively, not sure if this is possible but

out3 = f.C([1, 2, 3]).map(lambda x: x * 2)

out3.expression = "map([1, 2, 3], { # * 2 })"


```

Not sure how much of this is possible but this would be a huge improvement if this was available.

samj1912:

Fixes #336 

Examples at https://github.com/samj1912/hera-workflows/blob/expr/tests/test_expr.py 

Rendered Docs - https://hera-workflows--340.org.readthedocs.build/en/340/expr.html


samj1912:
@flaviuvadan started with something simple, we can definitely expand on this, but it looks like this is possible.

samj1912:
@flaviuvadan I have added full transpilation support for expr.

Complete docs are available at https://github.com/samj1912/hera-workflows/blob/expr/docs/expr.md

pyproject.toml
```diff
@@ -48,6 +48,8 @@ build = "*"
 isort = "*"
 tox = "<4.0"
 types-PyYAML = "*"
+jsonpath-ng = "^1.5.3"
```
samj1912, line 51: This was added as a dev dependency to verify jsonpath-ng transpilation.

docs/index.rst
```diff
@@ -58,6 +58,10 @@ Welcome to Hera's documentation!
    examples/workflow_with_custom_image
    examples/workflow_with_template_ref
 
+.. toctree::
+   :caption: Expr Transpilation
+
+   expr
```
flaviuvadan, line 64: This is a positive feedback comment :) 

At first I thought of suggesting we [place these docs with the other ones](https://github.com/argoproj-labs/hera-workflows/blob/main/docs/index.rst) but I think adding `expr` as its own section in the docs allows a fairly complicated topic to be isolated into its own logical unit of explanation, which users will likely appreciate

src/hera/expr/_node.py
```diff
@@ -0,0 +1,224 @@
...
+    def length(self):
+        return Callable("len", self)
+
+    def as_float(self):
+        return Callable("asFloat", self)
+
+    def as_int(self):
+        return Callable("asInt", self)
+
+    def check(self, truthy_value: 'Node', falsey_value: 'Node') -> 'Check':
```
flaviuvadan, line 69: Nit - I think this could be `falsy` by majority rule (it is, still, grammatically correct)
```suggestion
    def check(self, truthy_value: 'Node', falsy_value: 'Node') -> 'Check':
```

src/hera/expr/_node.py
```diff
@@ -0,0 +1,224 @@
...
+
+class Identifier(Node):
+    def __init__(self, value: str = ""):
+        self.value = value
+
+    def __repr__(self) -> str:
+        return self.value
+
+
+class Parentheses(Node):
```
flaviuvadan, line 146: I find `Constant` and `Identifier` conceptually sound in the sense that users are likely to have a 1:1 mapping between the _concept_ of a `Constant` and the _purpose_ of a `Constant`, which should facilitate usage. I have the same impression of `Identifier`. However, I am not so sure of `Parentheses`. I wonder if `Group` would be better conceptually? As a user if I want to group a statement I definitely _add_/_use_ parentheses. However, parentheses are an expression of a group rather than a group on its own. The group is a semantic element while the parentheses are a syntactic element. What do you think about the idea of using `Group`? If Argo uses the notion of `Parentheses` somewhere please point me to it! If that is the case then a 1:1 mapping with Argo is very reasonable and aligned with Hera post v4

src/hera/expr/_node.py
```diff
@@ -0,0 +1,224 @@
+from typing import Any, Dict, List, Optional, Union
```
flaviuvadan, line 1: Any chance you are open to adding more inline documentation/docstrings to the elements that are implemented in this module? For long term maintenance I think it would be very helpful to people who will read this code. I think, primarily, class docstrings would be most helpful as independent methods, such as `__add__`, _should_ be understandable at first sight given they are part of the Python ecosystem; plus, the implementation is short and only gives new semantic meaning to the operators, which should be fairly accessible

src/hera/expr/_sprig.py
```diff
@@ -0,0 +1,16 @@
+from typing import Any
```
flaviuvadan, line 1: Perhaps documentation with a link to http://masterminds.github.io/sprig/ would be helpful

flaviuvadan:
Super, super excited for this contribution @samj1912! 🚀 It looks amazing. I left a question about the concept of `Parentheses` and some suggestions for a bit more documentation

src/hera/expr/_node.py
samj1912: @flaviuvadan added some comments and docstrings to explain what is going on. Hopefully that helps.

src/hera/expr/_node.py
samj1912: @flaviuvadan happy to rename this class. How would we want to expose it publicly? Currently it is exposed publicly via the class `P` in the hera.expr module.

flaviuvadan: A global alias of `G`, perhaps?

samj1912: @flaviuvadan that might cause some confusion with the global context variable `g`. Are we okay with that?

samj1912:
@flaviuvadan also added a convenience method to easily surround output expressions with `{{}}` in format string.

Trollgeir:
I really love this @samj1912 ! Like @flaviuvadan said we've been thinking about a system like this for a while, but haven't gotten around to it. Your contribution is massive!

I think we should also add at least one example workflow to show the users some practical applications of these new expressions.

samj1912:
@Trollgeir added https://github.com/samj1912/hera-workflows/blob/expr/examples/complex_expr.py which is a python version of https://github.com/argoproj/argo-workflows/blob/master/examples/expression-reusing-verbose-snippets.yaml 

This truly shows how powerful/easy it is to write this in python as opposed to yaml.

examples/complex_expr.py
```diff
@@ -0,0 +1,53 @@
...
+            {
+                "avg": sprig.add(*iterators) / 5,
+                "min": sprig.min(*iterators),
+                "max": sprig.max(*iterators),
+            }
+        ).to_json()
+    )[0]
+
+
+with Workflow("expression-reusing-verbose-snippets", generate_name=True) as w:
```
Trollgeir, line 36:
```suggestion
with Workflow("expression-reusing-verbose-snippets-", generate_name=True) as w:
```

src/hera/expr/_node.py
flaviuvadan: I originally thought the risk of confusion would not be too high but seeing [this line](https://github.com/argoproj-labs/hera-workflows/pull/340/files#diff-179cd0b6feab44a19c4a993d24c10669a890248331dfd25637bebdfbcf60ca9aR35) suggested, to me, that something like `g[...]` and `G[...]` _could_ both be syntactically correct in some instances but mean something very different, which might cause confusion and challenges debugging. I think we should leave `P` for `Parentheses` as is for now, and perhaps address this later if it _does_ turn out to be confusing in any way!

examples/complex_expr.py
```diff
@@ -0,0 +1,53 @@
+"""This example showcases how to string together complex expressions.
```
flaviuvadan, line 1: Any chance you can make this an example in [here](https://github.com/argoproj-labs/hera-workflows/tree/main/docs/examples) and add it to [this list](https://github.com/argoproj-labs/hera-workflows/blob/main/docs/index.rst)? I think it would complement the documentation written for expr

flaviuvadan (approved):
This looks amazing @samj1912! Thank you again for the contribution and the discussions on all the questions :) Left a last suggestion for making the `expr.py` example into a markdown example as well. I think we can merge after that

src/hera/__init__.py
```diff
@@ -31,15 +31,15 @@
     get_global_api_version,
```
flaviuvadan, line 31: Any chance `expr` can be exposed here as well? `from hera.expr import C, P, ...`

samj1912: We can expose them here as well. Since they were fairly short named variables and come from a well scoped part of hera, having it scoped to `expr` makes it easy to namespace them/know where they came from.

For eg.

```py
import hera.expr as expr

expr.it
expr.g
expr.C
```

vs

```py
import hera

hera.it
hera.g
hera.C
```

LMK what you think.

flaviuvadan: Perhaps this is another instance in which something is a bit unclear and we should leave as is in v0 and welcome feedback from the community. If someone would love to have this exposed at the hera package init level, a PR will be welcome :)

samj1912:
@flaviuvadan I believe all the comments should be addressed. Are we planning a new release anytime soon?

flaviuvadan (approved):

PR #231, opened 2022-07-29
Fix WorkflowService.get_workflow_link when using set_global_host

JacobHayes:
The WorkflowService class was setting `self._host` to the value passed in, but when using `set_global_host`, we often don't pass that parameter. Instead, the instantiated `Config` object calls `get_global_config` to set a fallback for `host=None`, however this wasn't visible to `WorkflowService._host`. Instead of "duplicating" the value across the two classes, I just changed `WorkflowService` to reference the `config` object directly.

PR #350, opened 2022-11-08
Fix WorkflowService.get_workflow_link when using set_global_host

JacobHayes:
Reapplies https://github.com/argoproj-labs/hera-workflows/pull/231, which got accidentally removed in https://github.com/argoproj-labs/hera-workflows/pull/253.


JacobHayes:
Codecov was an upload flake, a retry should do 🤞 (I wonder if this has to do with me / an external contributor submitting the PR and the GITHUB_TOKEN not being set, so codecov can't find 🤔)

flaviuvadan (approved):

Issue #258, opened 2022-09-14
Support for Task/Workflow hooks or modifiers


JacobHayes:
Hi, I'd like to be able to set some "global" hooks that will be called during Task or Workflow creation that can modify them. For example, my company has:
- custom GKE selectors and tolerations required to use high mem instances that must be set in any code that _may_ provision high mem jobs
- custom GKE volume storage classes that set `WaitForFirstConsumer` to prevent volumes from being provisioned in 1 zone, but the pod in another (leading to unschedulable jobs).
- labels that we apply to most jobs for cost and other attribution; we'd like to either add defaults if not set OR error to catch missing ones

For all tasks/workflows submitted within our projects, we'd like these to be handled automatically, rather than every use needing the extra logic. Additionally, I've been thinking about writing some code that wraps `hera` and may not expose Tasks/Workflows directly (ie: it will create those for the user) - this layer would make it a bit more challenging to apply these customizations, but hooks might be an easy way to bypass.

--

A couple design points / questions:
- I think it'd be useful for these to run on Task/Workflow _creation_ (ie: at the end of `__init__`), but another option might be to call them on `.build` (just before generating the spec). I think on `__init__` would be beneficial because:
    - it would allow folks to inspect the generated object and tweak further if desired
    - any hook errors, intentionally or not, will show tracebacks right at the source, not later where `.build` is finally is called
- the hooks could probably be registered to a specific `WorkflowService` instance, rather than be "true" globals. However, setting them on a `WorkflowService` instance would force us to use the `.build` approach from the item above, otherwise the `Task` won't have the service available to identify the hooks. If they were true global hooks, then `Task.__init__` could just find and call them at the end.
- What is the right name for this concept, "hook", "modifier", etc?
    - I somewhat like "hook" because nothing _forces_ the implementation to modify (though I don't know what else they'd do 😅) and it hints at the mechanism (that things are being "intercepted")

--

Spitballing the updates for `WorkflowService`, maybe something like:
```python
TaskHook = Callable[[Task], None]
WorkflowHook = Callable[[Workflow], None]

# Add instance attributes tracking the hooks (pretending this is a pydantic model, so setting fields this way). dicts instead of lists may be useful in case we want to de-register or override.
class WorkflowService(...):
    task_hooks = list[TaskHook] # An ordered set of hooks that will be called on Task creation in FIFO order.
[...]

PR #351, opened 2022-11-09
Add support for `Task`/`Workflow` hooks and create a `GlobalConfig`

flaviuvadan:
Fixes: #258 

CC: @JacobHayes (OP author)
CC: @elliotgunton (supporter of global config, PR discussion)

### DIFF
--- a/docs/examples/gpu.md
+++ b/docs/examples/gpu.md
@@ -39,7 +39,7 @@ def do():
 
 
 with Workflow("gpu") as w:
-    gke_k80_gpu = {"cloud.google.com/gke-accelerator": "nvidia-tesla-k80"}
+    gke_k80_gpu = {"cloud.google.com/gke-accelerator": "nvidia-tesla-t4"}
     d = Task(
         "do",
         do,
--- a/tests/test_global_config.py
+++ b/tests/test_global_config.py
@@ -68,8 +68,11 @@ class TestGlobalConfig:
         c.task_post_init_hooks = [hook1, hook2]
         assert len(c.task_post_init_hooks) == 2
 
-        c.task_post_init_hooks = [hook1, hook2]
+        c.task_post_init_hooks = (hook1, hook2)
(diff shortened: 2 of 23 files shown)


tests/test_global_config.py
```diff
@@ -0,0 +1,84 @@
...
+        assert c.workflow_post_init_hooks == ()
+
+        def hook1(w: Workflow) -> None:
+            w.name = '123'
+
+        def hook2(w: Workflow) -> None:
+            w.service_account_name = "abc"
+
+        c.workflow_post_init_hooks = [hook1, hook2]
+        assert len(c.workflow_post_init_hooks) == 2
```
Trollgeir, line 83: This test (and the same above) seems a bit redundant because it's impossible that the length isn't 2

Trollgeir (approved):
LGTM!

The only thing I don't really like is that we're going to squash all this. It should either have been seperate PRs, or we should support rebasing :)

tests/test_global_config.py
flaviuvadan: Expanded test for the condition in which someone adds hooks on top of existing ones, which is why the PR contained this check for 2 hooks

flaviuvadan: Should have added the check in the first version of the PR.. that would have reduced confusion. Apologies :(

docs/examples/gpu.md
```diff
@@ -39,7 +39,7 @@ def do():
 
 
 with Workflow("gpu") as w:
-    gke_k80_gpu = {"cloud.google.com/gke-accelerator": "nvidia-tesla-k80"}
+    gke_k80_gpu = {"cloud.google.com/gke-accelerator": "nvidia-tesla-t4"}
```
flaviuvadan, line 42: K80s were silently deprecated by Google Cloud so switching examples to T4s

JacobHayes (commented):
Super excited for these!

Issue #346, opened 2022-11-02
Support unset parameter values for workflows


flaviuvadan:
Argo Workflow YAML files can be created with parameters with unset values. These values are set upon workflow creation via the Argo CLI, which sets the parameter values. Users would like to create such YAML files via Hera and then use the Argo CLI to parameterize the workflow. Hera workflow parameters currently require values to be set but they should not given the way Argo Workflows support parameters without values

elliotgunton:
+1 I found this problem today when trying to re-create yaml WorkflowTemplates in Hera. Will also add a note to https://github.com/argoproj-labs/hera-workflows/issues/345 but wanted to raise here as the other use case from 
>create such YAML files via Hera and then use the Argo CLI to parameterize the workflow

would be using a `WorkflowTemplate` with parameters that don't have defaults, in which case using the template in a `TemplateRef` without the parameter present in the `Task`'s `inputs` fails the workflow when linting/creating.

dbragdon1:
+1 also ran into this issue today. I have a lot of templates that I reuse and inherit in downstream workflows, where each workflow uses the same template but with different parameters. Would really appreciate this being implemented.

PR #352, opened 2022-11-10
Support unset workflow parameters to enable GitOps style submissions of workflow + usage of parameterized templates

flaviuvadan:
Fixes: #346 

This PR removes the Hera validation of `value` + `value_from` in the `Parameter` implementation. This allows setting a `Parameter` at the global workflow level via the `Workflow(..., parameters=[Parameter(...)])` field. This will define a workflow containing a global parameter with some name but _without_ a value. With that, a `Task` can obtain a workflow global parameter reference via the `Workflow.get_parameter(...)` API, which exposes the `'{{workflow.parameters.msg}}'` to the `Task` parameter. The value of this parameter can be set via an Argo CLI parameter specification. For testing purposes I created a `WorkflowTemplate`:

```python
from hera import Task, Parameter, WorkflowTemplate


def say(msg: str) -> None:
    print(msg)


with WorkflowTemplate(
    "hera-workflow-templates",
    dag_name="dag",
    parameters=[Parameter("msg")]
) as wt:
    Task('say', inputs=[wt.get_parameter("msg")])

wt.create()
```

The above results in the creation of a workflow template called `hera-workflow-templates`:
Screen Shot 2022-11-10 at 3 52 08 PM

In order to use this workflow template and submit a workflow I issued the following command:
```shell
argo submit --from workflowtemplate/hera-workflow-templates -p msg="hello"
```

This resulted in the successful creation and execution of the following workflow (highlighted parameter value):
Screen Shot 2022-11-10 at 3 53 10 PM

In addition to a workflow template I also functionally tested that after generating a `workflow.yaml` file from the `hello_hera.py` example, parameterized by a global `msg` parameter, I can submit the workflow YAML file via 
```shell
argo submit hello.yaml -p msg="hello"
```

I skipped those screenshots because they are very similar to the ones above.

CC: @elliotgunton 
CC: @samj1912 
CC: @dbragdon1

### DIFF
--- a/docs/examples/gitops.md
+++ b/docs/examples/gitops.md
@@ -1,0 +1,31 @@
+# Sample GitOps pattern
+
+This example showcases the hello world example of Hera in a GitOps style. The workflow gets a global parameter set
+called `msg`. This parameter has neither a `value` nor a `value_from` set. This supports the generation of a workflow
+YAML file that has the full definition and supports submission via `argo submit hello.yaml -p msg="hello"`. This
+parameterizes the workflow global parameter called `msg` to have a `value` of `hello`. This `hello` is then passed to
+the `Task` via the global parameter `w.get_parameter('msg')`, which sets the `'{{workflow.parameters.msg}}'` on the
+task parameter definition.
+
+## Workflow definition (Python)
+
+```python
+from hera import Task, Workflow, Parameter
+
+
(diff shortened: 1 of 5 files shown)


Trollgeir (approved):
LGTM, nice work!

PR #355, opened 2022-11-12
Misc follow ups from #351

JacobHayes:
Saw a couple things we could cleanup when reviewing #351, so just PRing since I got to the review late! :smile:

- Update more k80->t4 refs
- Remove old private host_config vars (if we left these for softer backwards compatibility, even though private, they wouldn't work the same anyway since they're set only on import)
- Hide hera.host_config deprecations in pytest
- Simplify GlobalConfig.reset
- Remove unused config reset and imports in tests


JacobHayes:
I can't help but reduce coverage... aiming for net negative lines contributed I guess lol 😅

flaviuvadan (approved):
Looks awesome 🚀 thank you for the follow up work!

flaviuvadan (approved):

Issue #361, opened 2022-11-17
Question: How to run a DAG via on_exit of a another DAG?


ipl31:
Hello,
I have a situation where I have multiple DAGs, for example d1 and d2. I would like to run d2 via d1.on_exit(). I want d2 to be executed regardless of the outcome (success, failure) of d1. I suspect there is a naive mistake or misunderstanding in my implementation. This example is contrived, the real world use case is a WF with multiple DAGs that each have their own post processing dag to run via on_exit().

I tried the following:

```
    with Workflow(name=workflow_name, service=workflow_service, service_account_name="argo-workflow") as w:
        def hello():
            print("hello")
            with open("bar", "w") as f:
                f.write("bar")
        def fails():
            raise Exception

        with DAG("main") as main_dag:
            m1 = Task("fails", fails)

        with DAG("exit") as exit_dag:
            e1 = Task("hello", hello)
            e2 = Task("hello2", hello)
            e3 = Task("hello3", hello)
            e1 >> [e2, e3]

        d1 = Task("d1", dag=main_dag)
        d2 = Task("d2", dag=exit_dag)
        d1.on_exit(d2)
    w.create()
```

However when the workflow runs I get an error indicating that the d2 template could not be found:
```

template d2 not found
```
Yaml from the argo server:
```
metadata:
  name: ken-test-e06deab1-1668645770
  namespace: argo-workflows
  uid: e25243fe-2a19-45f4-9772-a69a82f2cce6
  resourceVersion: '60758604'
  generation: 3
  creationTimestamp: '2022-11-17T00:42:51Z'
  labels:
    workflows.argoproj.io/completed: 'true'
    workflows.argoproj.io/creator: system-serviceaccount-argo-workflows-admin-user
    workflows.argoproj.io/phase: Error
  annotations:
    workflows.argoproj.io/pod-name-format: v2
  managedFields:
    - manager: argo
      operation: Update
      apiVersion: argoproj.io/v1alpha1
      time: '2022-11-17T00:42:51Z'
      fieldsType: FieldsV1
      fieldsV1:
        'f:metadata':
          'f:labels':
            .: {}
            'f:workflows.argoproj.io/creator': {}
        'f:spec': {}
    - manager: workflow-controller
      operation: Update
      apiVersion: argoproj.io/v1alpha1
      time: '2022-11-17T00:43:01Z'
      fieldsType: FieldsV1
      fieldsV1:
        'f:metadata':
          'f:annotations':
[...]

PR #362, opened 2022-11-21
[WIP] Adjust template creation for exit DAGs

flaviuvadan:
#361 brings up a scenario in which Hera does not create a template that is required. This template is supposed to reference an `onExit` task. Hera is currently missing the logic that would allow it to create the template with a DAG reference that is required in order to run a DAG as an `exitTask` on any Hera created task. This PR is work in progress - it might not be the ideal solution to the problem. See #361 for a discussion on the issue, example generated YAML with missing and added template

### DIFF
--- a/src/hera/dag.py
+++ b/src/hera/dag.py
@@ -51,7 +51,18 @@ class DAG(IO):
         # Assemble the templates from sub-dags
         sub_templates = [t.dag._build_templates() for t in self.tasks if t.dag]
         sub_templates = [t for sublist in sub_templates for t in sublist]  # flatten
-        return templates + sub_templates
+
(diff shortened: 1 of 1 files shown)


ipl31:
Thanks for looking into this @flaviuvadan . I am on vacation this week. When I get back next week I can test this change out in my environment.

flaviuvadan:
> Thanks for looking into this @flaviuvadan . I am on vacation this week. When I get back next week I can test this change out in my environment.

Sounds great! Enjoy the time off 🌴

flaviuvadan:
@ipl31, @Trollgeir and I paired on the solution for this and will open a different PR!

Issue #344, opened 2022-10-31
Make `inputs` design consistent with `with_param`


flaviuvadan:
Making `Parameter`s for a simple list of argument can be cumbersome, especially if users only want a mapping of a key to a value, such as the case for fanouts + parameters. Hera could support both `Parameter`, `Artifact`, and `dict` mappings. Specifically, a dictionary mapping can be interpreted as a list of `Parameters` to set as inputs

See https://github.com/argoproj-labs/hera-workflows/issues/318#issuecomment-1289351225 and https://github.com/argoproj-labs/hera-workflows/issues/318#issuecomment-1289472631

CC: @JacobHayes

flaviuvadan:
We can definitely do that! @Trollgeir and I discussed the option of generating Pydantic models from the OpenAPI specification of Argo Workflows ([Slack post here](https://cloud-native.slack.com/archives/C03NRMD9KPY/p1666821255944699)), we can actually introduce pydantic now, make `Parameter(BaseModel)`, add the validation on top of pydantic, and use the field types you linked above to enrich type accessibility

PR #368, opened 2022-11-27
Support mapping specifications for parameters on `inputs` fields

flaviuvadan:
Fixes: #344

### DIFF
--- a/docs/examples/parameters.md
+++ b/docs/examples/parameters.md
@@ -1,0 +1,20 @@
+# Parameters
+
+This example showcases how one can set parameters on Hera tasks
+
+```python
+from hera import Task, Workflow, set_global_host, set_global_token
+
+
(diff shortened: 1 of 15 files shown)


Trollgeir:
```python
def hello(a: str, b: int, c: dict):
    print(f"a = {a}, type(a) = {type(a)}")
    print(f"b = {b}, type(b) = {type(b)}")
    print(f"c = {c}, type(c) = {type(c)}")
```
Before reading the implementation, I was under the impression that the type annotations for the function arguments would be functional. In other words, that we would attempt to cast whatever the input was into that type. We're still just doing `json.loads` and get the types we want, but it's not a cast-mapping from the signature. 

This being said, I'm not entirely sure what the best course of action is. I just wanted to raise the topic in case someone has an opinion on the matter!

Trollgeir (approved):

PR #376, opened 2022-12-03
Add raw artifact

dubeau:
Adding the raw artifact as it is needed for some workflows.


flaviuvadan (approved):
Thank you so much for the contribution @dubeau! 🚀  Very excited to see this one. We're going to release it as part of 4.3 soon 😃

PR #387, opened 2022-12-16
Add passing yaml dump options to to_yaml

### LINKED ISSUE #378: Ordering keys in the yaml dump
When using the `to_yaml` function of Workflows, it results in unnatural ordering, in particular relating to `Task` objects

```py
with Workflow(
   ...
) as w:
    a = Task(
        name="entrypoint",
        command=["python", "-c"],
        args=["print('Hello, a!')"],
    )

w.to_yaml()
```

Gets yaml for the `entrypoint` task template:
```yaml
  templates:
  - container:
      args:
      - print('Hello, a!')
      command:
      - python
      - -c
      image: python:3.7
    name: entrypoint
```
Where `name` comes after the `container` definition. Is there a way or could we add some way to manually order the yaml spec?

samj1912:
This should allow users to pass in various yaml dump options to customize their yaml output per their wishes.

Fixes #378 

pyyaml exposes a sort_keys=false param which can be used. See https://github.com/yaml/pyyaml/issues/110#issuecomment-500921155

### DIFF
--- a/src/hera/workflow.py
+++ b/src/hera/workflow.py
@@ -1,7 +1,7 @@
 """The implementation of a Hera workflow for Argo-based workflows"""
 import json
 from types import ModuleType
-from typing import Dict, List, Optional, Tuple, TypeVar, Union
+from typing import Any, Dict, List, Optional, Tuple, TypeVar, Union
 
 from argo_workflows.model_utils import model_to_dict
 from argo_workflows.models import (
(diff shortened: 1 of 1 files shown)


flaviuvadan (approved):
Thank @samj1912 I will take a look at the coverage issue in another PR as it does not have to do with the contribution here

PR #388, opened 2022-12-16
Add supplied to value from and expand param

flaviuvadan:
CC: @samj1912

### DIFF
--- a/src/hera/value_from.py
+++ b/src/hera/value_from.py
@@ -32,7 +32,7 @@ class ValueFrom:
         `dag.get_parameter(...)`, and `workflow.get_parameter(...)`.
     path: Optional[str] = None
         Path in the container to retrieve an output parameter value from in container templates.
-    supplied: Optional[bool] = None
+    supplied: bool = False
         Whether the value will be filled via CLI, API, UI, etc.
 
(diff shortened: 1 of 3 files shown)


src/hera/value_from.py
```diff
@@ -46,6 +48,7 @@ class ValueFrom:
     json_path: Optional[str] = None
     parameter: Optional[str] = None
     path: Optional[str] = None
+    supplied: Optional[bool] = None
```
samj1912, line 51: Should this be a bool only instead? It seems like it doesn't matter if the value is set to True or False or not.

samj1912, line 51:
```suggestion
    supplied: bool = False
```

src/hera/value_from.py
```diff
@@ -70,4 +73,6 @@ def build(self) -> IoArgoprojWorkflowV1alpha1ValueFrom:
             setattr(value_from, "parameter", self.parameter)
         if self.path is not None:
             setattr(value_from, "path", self.path)
+        if self.supplied is not None:
```
samj1912, line 76:
```suggestion
        if self.supplied:
```

PR #389, opened 2022-12-16
Allow poetry in tox allowed list

flaviuvadan:
CC: @samj1912

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -8,6 +8,7 @@ addopts = --durations=5 -vv
 [coverage:run]
 branch = true
 parallel = true
+allowlist_externals = poetry
 
 [coverage:report]
 exclude_lines =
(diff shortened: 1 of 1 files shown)


tox.ini
```diff
@@ -1,6 +1,7 @@
 [tox]
 envlist = py{37,38,39,310},coverage,lint,typecheck,format
 isolated_build = true
+allowlist_externals = *
```
samj1912, line 4:
```suggestion
allowlist_externals = poetry
```
should this just be poetry?

flaviuvadan: Tried in https://github.com/argoproj-labs/hera-workflows/pull/389/commits/f6f80aef9a5fe0061291f683c21b8f834a8f4bd8 and it did not seem to work, which is why I wanted to give "*" a go 🤔

samj1912:
@flaviuvadan it seems to me that the issue is the release of tox 4.0 recently. If we pin back to 3.x the old config should work.

samj1912:
See https://github.com/argoproj-labs/hera-workflows/pull/391 should probably work

PR #391, opened 2022-12-16
Fix tox coverage issue

samj1912:


flaviuvadan:
I see... https://github.com/argoproj-labs/hera-workflows/blob/main/pyproject.toml#L49 already has tox<4. Thank you for figuring this out @samj1912!

samj1912:
@flaviuvadan looks like we are good to go

flaviuvadan (approved):

PR #356, opened 2022-11-14
Remove "." in name validation

elliotgunton:
* Argo allows dots in Workflow names

Fixes #353 

image

Issue #353, opened 2022-11-10
hera name validation is too strict 


samj1912:
Argo allows workflow template names to contain `.`.

Hera however prohibits this. Hera validation logic should be updated to match the restrictions at https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names

Trollgeir:
We decided to postpone this and instead keep the same validation logic for the workflow and template names for the initial implementation, and eventually branch it off into separate validations.

We're also looking into shifting all the models over to pydantic at some point which could utilize the native pydantic validation system and autogenerate all the models based on Argo's OpenAPI spec. What would be amazing would be if Argo's OpenAPI spec with constrained name fields could also be decorated with a [pattern](https://swagger.io/docs/specification/data-models/data-types/#pattern) (regex). This should then also be propagated to our autogenerated pydantic models.

 CC @alexec

PR #393, opened 2022-12-19
Fix name validation, add support for generate_name

tachylatus:
While browsing through the code, I spotted an issue in `validate_name` in that it uses [`re.match`](https://docs.python.org/3/library/re.html#re.match) (match the beginning of the string) rather than [`re.fullmatch`](https://docs.python.org/3/library/re.html#re.fullmatch) (match the whole string). One thing led to another, and I found and fixed a couple more bugs, and added a new parameter for proper validation of `metadata.generateName`.

Names in k8s cannot end in hyphen "-" or period ".", however the same is a valid choice for `metadata.generateName`. I have extended `validate_name` with an optional `generate_name` parameter to support this, synchronized with the identical parameter of `workflow.__init__`.

### Changes
- Validate entire name in `validate_name`, not only the beginning of the string
  - Without this, the invalid name `"test#"` would unexpectedly pass validation.
- Fix quoting of "." in pattern used in `validate_name`
  - Without this, the invalid name `r"test\@"` would unexpectedly pass validation
- Add a new `generate_name` boolean parameter, default to `False`; if `True`, a `name` ending in a single period "." or any number of hyphens "-" will pass validation, as alphanumeric characters are expected to be appended in the final name
  - Without this, the valid generate name `"test-"` would unexpectedly fail validation
- Strip trailing "-" and "." characters when workflow name is reused for an unnamed dag
  - Without this, the valid generate name `"test-"` would result in a validation error if the dag is not named explicitly

### DIFF
--- a/src/hera/validators.py
+++ b/src/hera/validators.py
@@ -10,11 +10,22 @@ def validate_name(name: str, max_length: Optional[int] = None, generate_name: bo
     ----------
     name: str
         The name which should be validated.
+    max_length: Optional[int] = None
+        Specify a maximum length of the name.
+        Example: Kubernetes labels have a maximum length of 63 characters.
+    generate_name: bool = False
+        Whether the provided name is to be used as a prefix for name generation.
+        If set, name is allowed to end in a single dot (.) or any number of hyphens (-).
 
     Raises
     ------
     ValueError
         When the name is invalid according to specifications.
+
+    Notes
+    -----
+    Official doc on object names in Kubernetes:
+    https://kubernetes.io/docs/concepts/overview/working-with-objects/names/
     """
     if max_length and len(name) > max_length:
         raise ValueError(f"Name is too long. Max length: {max_length}, found: {len(name)}")
--- a/src/hera/validators.py
+++ b/src/hera/validators.py
@@ -31,6 +31,8 @@ def validate_name(name: str, max_length: Optional[int] = None, generate_name: bo
         raise ValueError(f"Name is too long. Max length: {max_length}, found: {len(name)}")
     if "_" in name:
         raise ValueError("Name cannot include an underscore")
+    if not generate_name and name.endswith((".", "-")):
+        raise ValueError("Name cannot end with '.' nor '-', unless it is used as a prefix for name generation")
 
     pattern = r"[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*"
     if generate_name:
--- a/src/hera/validators.py
+++ b/src/hera/validators.py
@@ -22,7 +22,7 @@ def validate_name(name: str, max_length: Optional[int] = None) -> str:
         raise ValueError("Name cannot include an underscore")
 
     pattern = r"[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*"
-    match_obj = re.match(pattern, name)
+    match_obj = re.fullmatch(pattern, name)
     if not match_obj:
         raise ValueError(f"Name is invalid: '{name}'. Regex used for validation is {pattern}")
     return name
--- a/src/hera/validators.py
+++ b/src/hera/validators.py
@@ -3,7 +3,7 @@ import re
 from typing import Optional
 
 
-def validate_name(name: str, max_length: Optional[int] = None) -> str:
+def validate_name(name: str, max_length: Optional[int] = None, generate_name: bool = False) -> str:
     """Validates a name according to standard argo/kubernetes limitations
 
     Parameters
@@ -22,6 +22,8 @@ def validate_name(name: str, max_length: Optional[int] = None) -> str:
         raise ValueError("Name cannot include an underscore")
 
     pattern = r"[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*"
+    if generate_name:
+        pattern += r"(\.?|-*)"
     match_obj = re.fullmatch(pattern, name)
     if not match_obj:
         raise ValueError(f"Name is invalid: '{name}'. Regex used for validation is {pattern}")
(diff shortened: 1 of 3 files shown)


tachylatus:
Just had to fix the missing sign-offs + small change in commit message

samj1912:
@tachylatus JFYI, https://github.com/argoproj-labs/hera-workflows/issues/353 and https://github.com/argoproj-labs/hera-workflows/pull/356

src/hera/validators.py
```diff
@@ -3,7 +3,7 @@
 from typing import Optional
 
 
-def validate_name(name: str, max_length: Optional[int] = None) -> str:
+def validate_name(name: str, max_length: Optional[int] = None, generate_name: bool = False) -> str:
```
flaviuvadan, line 6: Small: any chance `generate_name: bool = False` can be part of the docstring?

src/hera/validators.py
```diff
@@ -21,8 +21,10 @@ def validate_name(name: str, max_length: Optional[int] = None) -> str:
     if "_" in name:
         raise ValueError("Name cannot include an underscore")
 
-    pattern = r"[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*"
-    match_obj = re.match(pattern, name)
+    pattern = r"[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*"
```
flaviuvadan, line 24: Small: any chance [this link](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/) can be included as a comment or a `Notes` in the docstring? I _think_ that documents the motivation for this specific validation and it would be nice to have a record in code that points contributors to it

src/hera/validators.py
```diff
@@ -21,8 +21,10 @@ def validate_name(name: str, max_length: Optional[int] = None) -> str:
...
+    if generate_name:
+        pattern += r"(\.?|-*)"
+    match_obj = re.fullmatch(pattern, name)
```
flaviuvadan, line 27: Good catch 🚀

src/hera/validators.py
```diff
@@ -21,8 +21,10 @@ def validate_name(name: str, max_length: Optional[int] = None) -> str:
     if "_" in name:
         raise ValueError("Name cannot include an underscore")
 
-    pattern = r"[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*"
-    match_obj = re.match(pattern, name)
+    pattern = r"[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*"
+    if generate_name:
+        pattern += r"(\.?|-*)"
```
flaviuvadan, line 26: Does this need a quantifier at the end?
```suggestion
        pattern += r"(\.?|-*)*"
```

src/hera/workflow.py
```diff
@@ -127,8 +127,8 @@ def __init__(
         active_deadline_seconds: Optional[int] = None,
         metrics: Optional[Union[Metric, List[Metric], Metrics]] = None,
     ):
-        self.name = validate_name(name)
-        dag_name = self.name if dag_name is None else dag_name
+        self.name = validate_name(name, generate_name=generate_name)
+        dag_name = self.name.rstrip("-.") if dag_name is None else dag_name
```
flaviuvadan, line 131: What do you think about making this part of the validation process as well? :) It might be useful to tell the user that `-.` or `.-` is not accepted the same way `_` is not accepted, which is reported by the validator

src/hera/validators.py
tachylatus: No, that would allow the name to end in multiple consecutive dots, which is invalid. There should always be at least one character between two dots (DNS naming restriction).

src/hera/validators.py
tachylatus: Added docstring for both generate_name and max_length

src/hera/validators.py
tachylatus: Added `Notes` in the docstring.

src/hera/workflow.py
tachylatus: Added explicit validation and custom error message for trailing `.` and `-` in `validators.py`.

Just to be clear, the addition of `rstrip("-.")` in `workflow.py` was to prevent unexpected validation error when an otherwise valid `name` would be invalid for use in `dag_name`.

src/hera/validators.py
tachylatus: Thanks 😄

tachylatus:
The "Read the Docs" build fails due to Sphinx <= 4.0.1 trying to import `environmentfilter` from `jinja2` ([code](https://github.com/sphinx-doc/sphinx/blob/v4.0.1/sphinx/util/rst.py#L21)), which has been deprecated in Jinja 3.0, removed in Jinja 3.1 ([Jinja2 documentation)](https://jinja.palletsprojects.com/en/3.0.x/api/#jinja2.environmentfilter)). Sphinx >= 4.0.2 should not have this issue ([code](https://github.com/sphinx-doc/sphinx/blob/v4.0.2/sphinx/util/rst.py#L21)),

flaviuvadan:
> The "Read the Docs" build fails due to Sphinx <= 4.0.1 trying to import `environmentfilter` from `jinja2` ([code](https://github.com/sphinx-doc/sphinx/blob/v4.0.1/sphinx/util/rst.py#L21)), which has been deprecated in Jinja 3.0, removed in Jinja 3.1 ([Jinja2 documentation)](https://jinja.palletsprojects.com/en/3.0.x/api/#jinja2.environmentfilter)). Sphinx >= 4.0.2 should not have this issue ([code](https://github.com/sphinx-doc/sphinx/blob/v4.0.2/sphinx/util/rst.py#L21)),

Fixed the Jinja issue on main, so it's all good! :)

flaviuvadan (approved):
Thank you for the contribution @tachylatus! This is awesome 🚀

PR #410, opened 2023-01-27
Ephemeral storage

lancour:
Kubernetes now supports an 'ephemeral-storage' spec on resources, which can allow pods some amount of specified memory for logging and the such. This adds the same named parameter to Hera.

### DIFF
--- a/tests/test_resources.py
+++ b/tests/test_resources.py
@@ -22,6 +22,12 @@ class TestResources:
         with pytest.raises(ValueError):
             Resources(memory_limit="4")
 
+    def test_init_raises_on_invalid_eph(self):
+        with pytest.raises(ValueError):
+            Resources(ephemeral_request="3m")
+        with pytest.raises(ValueError):
+            Resources(ephemeral_limit="2m")
+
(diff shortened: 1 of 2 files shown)


tests/test_resources.py
```diff
@@ -22,6 +22,12 @@ def test_init_raises_on_invalid_mem(self):
         with pytest.raises(ValueError):
             Resources(memory_limit="4")
 
+    def test_init_raises_on_invalid_eph(self):
+        with pytest.raises(ValueError):
+            Resources(ephemeral_request="3m")
+        with pytest.raises(ValueError):
+            Resources(ephemeral_limit="2m")
```
JacobHayes, line 29: @lancour looks like hera uses `assert`s, so these need to be `AssertionError`:
```suggestion
        with pytest.raises(AssertionError):
            Resources(ephemeral_request="3m")
        with pytest.raises(AssertionError):
            Resources(ephemeral_limit="2m")
```

JacobHayes (approved):
All seems sane to me - have you tried submitting a job and confirming the pod has the desired constraints?

flaviuvadan (approved):

PR #411, opened 2023-01-30
add container port support to Task by Hera

dpeer6:
We wanted to be able and scrape long running container metrics, and that requires a port to be exposed.
I compared the scenario with Hera and Argo Yaml, and indeed that was the missing part to be able and scrape the metrics.
see [here](https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/user-guides/getting-started.md#include-podmonitors)


flaviuvadan (commented):
Thank you for the catch and contribution @dpeer6! LGTM. Going to merge once the build is fine

dpeer6:
I hope now it is ok.

src/hera/task.py
```diff
@@ -272,6 +276,7 @@ def __init__(
         self.is_exit_task: bool = False
         self.depends: Optional[str] = None
         self.when: Optional[str] = None
+        self.ports = ports or []
```
flaviuvadan, line 279: I think this should stay as `ports` so that it takes a `None` value when one is not provided. That way when line 935 builds the ports it would not create an empty list. I believe that, currently, the empty list of ports would show up in the YAML manifest submitted to Argo. Then line 935 becomes: 
```python
ports=None if self.ports is None else [p.build() for p in self.ports]
```

dpeer6: Aha, good catch, so just to make sure:
line 279: self.port = ports
line 935: ports=None if self.ports is None else [p.build() for p in self.ports]

flaviuvadan (approved):
Awesome work @dpeer6! Thank you for the contribution

dpeer6:
@flaviuvadan Thank you!
When can I expect the next release that includes it, as an official package? like pip and/or conda?

PR #412, opened 2023-01-31
Fix parsing of task parameter defaults

JacobHayes:
There are a few edge cases I've seen with the parsing of a task function's parameter default values:
- `None` defaults are mistaken as required
    - eg: `ValueError: param in `with_params` is missing non-default argument: {'b'}`
- defaults are serialized with `str`, but deserialization in the task first tries `json.loads` before embedding as a string literal
    - eg: `x=False` is converted to a default of `"False"`, which fails `json.loads` (json requires `false`) and ends up as `x="False"` (which is truthy, so quietly wrong :grin:)

The first issue comes from mixing up explicit `None`s with an omitted argument to `Parameter`, which is fixed by using a `MISSING = object()` sentinel and serializing the explicit `None`s. The second issue comes from the `Parameter.default` value being serialized with `str` instead of `json.dumps`d.


JacobHayes:
> codecov/project — 95.3% (-0.2%) compared to 73699fa

I lower coverage every commit lmao

flaviuvadan (approved):

PR #413, opened 2023-01-31
Simplify Task and Workflow script comparisons

JacobHayes:
The tests comparing to static generated script output are a bit tricky to read and update due to the string continuations and special characters. Before making other changes to the generated scripts, I wanted to tidy this up to keep the PR tidy.

This changes the script comparisons to use multiline strings with `textwrap.dedent`.

### DIFF
--- a/tests/test_task.py
+++ b/tests/test_task.py
@@ -1,4 +1,5 @@
 import json
+from textwrap import dedent
 from unittest import mock
 
 import pytest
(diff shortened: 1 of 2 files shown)


JacobHayes:
This probably should have run tests (perhaps CI only runs them on changes to `src`?), but they passed locally for me:
```sh
$ git rev-parse HEAD
980a8e1e2fc1317dfd9b1bc3af2456f2733d9998
$ pytest
............................................................................................................................................. [ 39%]
............................................................................................................................................. [ 79%]
.........................................................................                                                                     [100%]
355 passed in 0.65s
```

flaviuvadan:
Tests did not run because of https://github.com/argoproj-labs/hera-workflows/blob/main/.github/workflows/cicd.yaml#L7

PR #414, opened 2023-01-31
Support multiprocessing scripts

JacobHayes:
NOTE: This PR is based off of https://github.com/argoproj-labs/hera-workflows/pull/413 to keep the diff cleaner (... I guess PRs from forks can't use base branches also from the fork, so just look at the last commit I guess haha and I can try to rebase once/if that is merged).

When using `multiprocessing` directly or indirectly (eg: with `dask.distributed.Client()`) in a task, forked subprocesses attempt to import the script again, causing recursive errors. A user can manually add an `if __name__ == "__main__":` check to their code, however this is hard to intuit ahead of time (the issue only arises *once run in Argo*, since when called normally as a function, imports won't cause issues) and can make tasks harder to test programmatically (again, calling the task func in a test won't pass the `__main__` comparison).

The fact that Hera embeds the function's *source code* directly as a module is an implementation detail, so it seems right to have Hera automatically add the `__main__` checks.

---

For example, when run directly as a script, this code:

```
from multiprocessing import Pool

chars = [72, 101, 114, 97]
with Pool(5) as p:
    print("".join(p.map(chr, chars)))
```

will repeatedly (infinitely?) error with:

```
RuntimeError:
        An attempt has been made to start a new process before the
        current process has finished its bootstrapping phase.

        This probably means that you are not using fork to start your
        child processes and you have forgotten to use the proper idiom
        in the main module:

            if __name__ == '__main__':
                freeze_support()
                ...

        The "freeze_support()" line can be omitted if the program
        is not going to be frozen to produce an executable.
```

### DIFF
--- a/src/hera/task.py
+++ b/src/hera/task.py
@@ -877,7 +877,13 @@ class Task(IO):
 
                 s = "".join(content[token_index:])
                 script += textwrap.dedent(s)
-                return textwrap.dedent(script)
+                script = textwrap.dedent(script)
+                # Guard everything under a main check to avoid recursive issues with
+                # tasks that import the defining module (eg: multiprocessing).
+                res = "if __name__ == '__main__':"
+                for line in script.split("\n"):
+                    res += f"\n    {line}" if line else "\n"
+                return res
         else:
             assert isinstance(self.source, str)
             return self.source
--- a/tests/test_task.py
+++ b/tests/test_task.py
@@ -1,4 +1,5 @@
 import json
+from textwrap import dedent
 from unittest import mock
 
 import pytest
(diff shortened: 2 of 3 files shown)


JacobHayes:
Ah, the local multiprocessing example runs differently on mac and linux - seems `spawn` is the default on mac and `fork` on linux. `spawn` seems to attempt an import while `fork` does not.

So, if you use `set_start_method("spawn")` on linux (as `dask.distributed` seems to [by default](https://docs.dask.org/en/stable/configuration.html#distributed.worker.multiprocessing-method)), that reproduces the issue.

However... the `spawn` method also reveals another issue: values referenced by the submitted function are pickled (and thus attempted to be imported, since pickle only serializes refs to eg: functions). This means lambdas (which can't be pickled) and any name not available at the module level _outside_ the `if __name__ == "__main__"` check will not be available to the subprocesses.

Given this change puts the `__main__` check over all of the script, this could actually cause a breaking change in the (rare?) case someone is using multiprocessing with a well-placed `__main__` check (only over the final pool creation). **However**, any code written with `multiprocessing` 

---

I can't think of any way to avoid these conflicting issues without either:
- explicitly separating a "module" part and a "main" part of the script (so Hera can place the `__main__` check in a good spot)
- doing that thing where we always embed the entire defining module (imports, other functions, and all) as one file and change the script to import that and call the function as a normal function

The latter change is a lot larger, but I think would make Hera "just work" in a lot more situations (you wouldn't have to ensure your imports are _inside_ the script function definition; you could reference other functions within the module like "normal" python, etc).

PR #407, opened 2023-01-23
Get Hera feature parity with Argo Workflows via Pydantic models

flaviuvadan:
TODO: 
- [ ] enrich PR description
- [ ] tests
- [ ] typechecks
- [x] validate examples
- [ ] document scripts
- [ ] enrich README

PR #419, opened 2023-02-07
[1/N] Add `workflows` module in preparation for V5 introduction of `events`

flaviuvadan:
Hera v5 will introduce a method that generates all the Argo Workflows models, which will help Hera with Argo feature parity. This means Hera will introduce events as well. However, there are some generated objects that have naming clashes, such as `Event`. A solution that has the potential to provide clarity between modules while at the same time saving users from having to perform import aliasing is separating the `workflows` and `events` module. This PR introduces the `workflows` module and migrates everything that currently exists over into the new module in prep for the new models. This work is a derivative of #407. I decided to break apart the PR to increase release safety

CC: @JacobHayes 
CC: @elliotgunton


samj1912 (approved):

elliotgunton (approved):
LGTM

PR #428, opened 2023-02-16
Parameter bug fixes

### LINKED ISSUE #425: Cannot submit Workflow that uses ValueFrom "supplied" field
When trying to create a Task which uses the "supplied" field of ValueFrom (see https://argoproj.github.io/argo-workflows/fields/#valuefrom), the submission fails as the field in Hera is a bool, so the output yaml contains `supplied: true` instead of the expected `supplied: {}` which is shown here https://github.com/argoproj/argo-workflows/blob/master/examples/suspend-template-outputs.yaml#L33

The error itself:
```
cannot unmarshal bool into Go struct field ValueFrom.template.spec.templates.outputs.parameters.valueFrom.supplied of type v1alpha1.SuppliedValueFrom
```

I believe this will be fixed by #420 as https://github.com/argoproj-labs/hera-workflows/pull/420/files#diff-b1d6ef48f102cc95d204c1e886127c7e057e3dca0622f020b0f61c8ea3d62ab1R1109 as `SuppliedValueFrom` will be an empty class, not a boolean, so the new usage code will look like

```py
Parameter(name="supplied-var", value_from=ValueFrom(supplied=SuppliedValueFrom()))
```

Just filing the issue for visibility and to ask for the PRs to merge and release asap! 🙏 Thanks!

### LINKED ISSUE #426: Enum input parameters ignored
The intermediate parameters example in Argo at https://argoproj.github.io/argo-workflows/intermediate-inputs/ is not actually possible as the input parameter enum is not output in the `as_input` function at https://github.com/argoproj-labs/hera-workflows/blob/main/src/hera/parameter.py#L92

The fix is to simply copy the `self.enum` usage from `as_output` at https://github.com/argoproj-labs/hera-workflows/blob/7aa297bc2d020e9ceef7853cabc2b89f2094233f/src/hera/parameter.py#L116-L117

### LINKED ISSUE #427: Suspend template task not implemented
cc: @samj1912

There is an example at https://hera-workflows.readthedocs.io/examples/suspend.html but the Task `__init__` doesn't contain a suspend kwarg.

https://github.com/argoproj-labs/hera-workflows/blob/main/src/hera/task.py#L184

elliotgunton:
Fixes #425
Fixes #426
Fixes #427

I think there are other issues with parameters as inputs to DAG templates which I'm still investigating, will document in new issues

### DIFF
--- a/tests/test_parameter.py
+++ b/tests/test_parameter.py
@@ -37,9 +37,10 @@ class TestParameter:
         assert arg.value_from.path == "abc"
 
     def test_as_input_returns_expected_parameter(self):
-        param = Parameter("a", default="42").as_input()
+        param = Parameter("a", default="42", enum=["42", "100"]).as_input()
         assert isinstance(param, IoArgoprojWorkflowV1alpha1Parameter)
         assert hasattr(param, "default")
+        assert hasattr(param, "enum") and param.enum == ["42", "100"]
         assert param.default == "42"
 
     def test_as_output_returns_expected_parameter(self):
--- a/tests/test_parameter.py
+++ b/tests/test_parameter.py
@@ -9,10 +9,20 @@ from hera.value_from import ValueFrom
 
 
 class TestParameter:
-    def test_init_raises_value_error(self):
+    def test_init_raises_value_error_using_value_and_value_from(self):
         with pytest.raises(ValueError) as e:
             Parameter("a", value_from=ValueFrom(default="42"), value="42")
         assert str(e.value) == "Cannot specify both `value` and `value_from` when instantiating `Parameter`"
+
+    def test_init_raises_value_error_using_invalid_value_with_enum(self):
+        with pytest.raises(ValueError) as e:
+            Parameter("a", value="42", enum=["41"])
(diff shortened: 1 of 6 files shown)


elliotgunton:
@flaviuvadan Could we cut a release with these fixes as a patch to 4.4? We'd like to use these fixes asap. Also @samj1912 suggested if a 4.4 version could be maintained for future fixes for some time alongside v5?

tests/test_parameter.py
```diff
@@ -37,9 +37,10 @@ def test_as_argument_returns_expected_parameter(self):
         assert arg.value_from.path == "abc"
 
     def test_as_input_returns_expected_parameter(self):
-        param = Parameter("a", default="42").as_input()
+        param = Parameter("a", default="42", enum=["42", "100"]).as_input()
```
elliotgunton, line 40: Might be helpful to add `if default not in enum: raise(Error)` in the `__init__` of Parameter?

flaviuvadan (approved):
Thank you again @elliotgunton! 🚀

PR #432, opened 2023-02-16
Remove restrictions from enums/defaults

samj1912:
enums/default values may be sourced from an expression/variable. In that case, the strict validation that hera performs currently to check if they are present in the enum may be invalid.

Remove this validation.


flaviuvadan (approved):
I assume this is the case but double checking that @elliotgunton is aware of this!

elliotgunton (approved):

elliotgunton:
Yep LGTM, I didn't consider values passed through expressions 🤦‍♂️

PR #434, opened 2023-02-17
[WIP] Consolidate parameter funcs

### LINKED ISSUE #430: Parameter `as_*` functions inconsistency -- "value_from" missing in as_input
The three functions in [parameter.py](https://github.com/argoproj-labs/hera-workflows/blob/main/src/hera/parameter.py):
* `as_argument`
* `as_input`
* `as_output`

treat the `Parameter`'s member variables differently but seem inconsistent (e.g. just fixed "enum" missing in `as_input` in #428). Is there a reference for how the member variables should be treated based on whether the `Parameter` is being used as an argument/input/output? The [field reference for parameter](https://argoproj.github.io/argo-workflows/fields/#parameter) doesn't mention explicitly which fields are allowed, e.g.
>Default is the default value to use for an input parameter

does this mean `default` isn't allowed for outputs? 

If there's a reference I will gladly PR fixes to go along with a `value_from` fix, see below.

---

It's not currently possible to set a `value_from` on an input parameter, seen in the argo examples here: https://github.com/argoproj/argo-workflows/blob/3f70162f95c9df6dc885a788164780f87cbd6e4d/examples/arguments-parameters-from-configmap.yaml#L20-L27

```yaml
    inputs:
      parameters:
      # Parameters can also be passed via configmap reference.
      - name: message
        valueFrom:
          configMapKeyRef:
            name: simple-parameters
            key: msg
```

---

We also found that the configmap key ref does not work as the `config_map_key_ref` member is a str, where it actually needs a "name" for the configmap, and a "key" -- this type is added in #420 here: https://github.com/argoproj-labs/hera-workflows/blob/f75597bd4f68bddbf7d2e39a146ee26615af55b9/src/hera/workflows/models/io/k8s/api/core/v1.py#L111

elliotgunton:
* Consolidating/removing the individual `as_X` functions. (WIP)
* Will fix #430 (needs reopening)

### DIFF
--- a/src/hera/parameter.py
+++ b/src/hera/parameter.py
@@ -75,10 +75,10 @@ class Parameter:
     def _build(self) -> IoArgoprojWorkflowV1alpha1Parameter:
         parameter = IoArgoprojWorkflowV1alpha1Parameter(name=self.name)
 
-        if self.value is not MISSING:
+        if self.value is not None:
             setattr(parameter, "value", self.value)
 
-        if self.default is not MISSING:
+        if self.default is not None:
             setattr(parameter, "default", self.default)
 
         if self.value_from is not None:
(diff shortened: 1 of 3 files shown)


elliotgunton:
Closing for #437 - these issues (and more!) will be fixed by the new incoming V5 types

PR #449, opened 2023-03-04
Add DAG support to V5

flaviuvadan:
CC: 
- @elliotgunton 
- @samj1912


src/hera/workflows/v5/dag.py
```diff
@@ -0,0 +1,70 @@
...
+        from hera.workflows.v5._context import _context
+
+        _context.enter(self)
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb) -> None:
+        from hera.workflows.v5._context import _context
+
+        for task in self._task_queue:
+            self.tasks.append(task._build_dag_task())
```
samj1912, line 29: I believe we should be able to simplify this if we inherit from _SubNodeMixin and get rid of these two methods?

src/hera/workflows/v5/dag.py
```diff
@@ -0,0 +1,70 @@
...
+
+        _context.exit()
+
+    def _add_sub(self, node: Any):
+        self.add_task(node)
+
+    def add_task(self, task: _DAGTaskMixin) -> None:
+        self._task_queue.append(task)
```
samj1912, line 37:
```suggestion
    def add_task(self, task: Union[_DAGTaskMixin, DAGTask]) -> None:
		self._task_queue.append(task)

File: src/hera/workflows/v5/dag.py [hunk 3]
```diff
@@ -0,0 +1,70 @@
...
+
+    def _build_template(self) -> _ModelTemplate:
```
samj1912, line 39: Can we resolve the task queue here instead of when we exit the context?

That way explicit tasks added outside of the context handler will also be accounted for.

```py
tasks = []
for task in self._task_queue:
	if isinstance(task, DAGTask):
		tasks.append(task)
	elif isinstance(task, _DagTaskMixin):
		tasks.append(task._build_dag_task())
```

src/hera/workflows/v5/_context.py
```diff
@@ -17,11 +17,15 @@ def enter(self, p: Subbable) -> None:
...
     def exit(self) -> None:
-        self._pieces.pop()
+        if len(self._pieces) == 1:
+            return  # only the workflow is in the context
 
-    def add_sub_node(self, node: _SubNodeMixin) -> None:
-        if self._pieces:
-            self._pieces[-1]._add_sub(node)
+        popped = self._pieces.pop()
+        main = self._pieces[0]
```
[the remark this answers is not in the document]
samj1912: I think we also need to make sure we do the _build_* calls for each sub node from the parent container when the build call is made to the parent container, otherwise we might miss changes.

PR #451, opened 2023-03-05
Add automated test-harness for examples and fix DAG tests

samj1912:
Signed-off-by: Sambhav Kothari <sambhavs.email@gmail.com>

The last commit adds a test harness to hera to allow it to compare yaml outputs from workflow module files in 

`hera.workflows.v5.examples` and compare them against the checked-in yaml files. The yaml files can be auto-generated by setting `HERA_REGENERATE_EXAMPLES` environment variable or they are generated once a new python example is checked in without a corresponding yaml output.

It also adds a `hera.workflows.v5.examples.upstream` folder, that does all of the above + fetches the corresponding example from argo workflow examples and compares them against the workflow module as well as the generated yaml file. 

It also fixes the DAG example to match the upstream example exactly.

It also fixes the script template to generate an appropriate output.


src/hera/workflows/v5/examples/coinflip.yaml
```diff
@@ -0,0 +1,26 @@
+apiVersion: argoproj.io/v1alpha1
+kind: Workflow
+metadata:
+  annotations:
+    workflows.argoproj.io/description: This is an example of coin flip defined as
+      a sequence of conditional steps.
+  generateName: coinflip-
+  namespace: default
```
elliotgunton, line 8: `namespace` isn't used in upstream examples, I see an empty namespace is equivalent to "default" but I don't think we need to include it here - I see it's because in the `GlobalConfig`, `namespace` is set to `"default"`, which is then set in the `Workflow` class as a default value [here](https://github.com/samj1912/hera-workflows/blob/91385cdda1c75fe4bfaeb889caf2306a7c386835/src/hera/workflows/v5/workflow.py#L72).

src/hera/workflows/v5/examples/upstream/dag_conditional_parameters.py
```diff
@@ -0,0 +1,64 @@
...
+    entrypoint="main",
+    namespace=None,
+    labels={"workflows.argoproj.io/test": "true"},
+    annotations={
+        "workflows.argoproj.io/description": (
+            "Conditional parameters provide a way to choose the output parameters "
+            "based on expression.\n\nIn this example DAG template has two task which"
+            " will run conditionally based on `when`.\n\nBased on this condition one"
+            " of task may not execute."
+            " The template's output parameter will be set to the\nexecuted taks's output result.\n"
```
elliotgunton, line 48: The typos 😭 But they're from the upstream example - are we syncing files down to this repo or copied once to compare python<>yaml?

If they're not resynced:
```suggestion
            "Conditional parameters provide a way to choose the output parameters "
            "based on an expression.\n\nIn this example the DAG template has two tasks which"
            " will run conditionally based on `when`.\n\nBased on this condition, one"
            " of the tasks may not execute."
            " The template's output parameter will be set to the\nexecuted task's output result.\n"
```

src/hera/workflows/v5/examples/upstream/dag_diamond.py
```diff
@@ -0,0 +1,32 @@
...
+echo = Container(
+    name="echo",
+    image="alpine:3.7",
+    command=["echo", "{{inputs.parameters.message}}"],
+    inputs=[Parameter(name="message")],
+)
+with Workflow(
+    generate_name="dag-diamond-",
+    entrypoint="diamond",
+    namespace=None,
```
elliotgunton, line 23: As in the yaml comment, I also think this adds an extra line for new users to understand without really being needed.

src/hera/workflows/v5/task.py
```diff
@@ -0,0 +1,166 @@
...
+from hera.workflows.v5.workflow_status import WorkflowStatus
+
+
+class Task(_SubNodeMixin):
+    name: str
+    arguments: Optional[Arguments] = None
+    continue_on: Optional[ContinueOn] = None
+    dependencies: Optional[List[str]] = None
+    depends: Optional[str] = None
+    hooks: Optional[Dict[str, LifecycleHook]]
```
elliotgunton, line 27: Missing a `None`?

src/hera/workflows/v5/tests/test_workflow.py
```diff
@@ -1,58 +1,47 @@
...
-                    "container": {"args": ["echo 'it was tails'"], "command": ["sh", "-c"], "image": "alpine:3.6"},
-                    "name": "tails",
-                },
-            ]
-        },
-    }
+    if not yaml_path.exists() or os.environ.get("HERA_REGENERATE_EXAMPLES"):
+        yaml_path.write_text(yaml.dump(output))
+    if not upstream_yaml_path.exists() or os.environ.get("HERA_REGENERATE_EXAMPLES"):
+        upstream_yaml_path.write_text(requests.get(f"{ARGO_EXAMPLES_URL}/{module_name.replace('_', '-')}.yaml").text)
```
elliotgunton, line 45: I see we're pulling fresh files during the test

src/hera/workflows/v5/workflow.py
```diff
@@ -123,7 +125,22 @@ def _set_workflows_service(cls, v):
...
 
+    @validator('kind', pre=True, always=True)
+    def _set_kind(cls, v):
+        if v is None:
+            return cls.__name__  # type: ignore
+        return v
+
     def build(self) -> _ModelWorkflow:
+        templates = []
+        for template in self.templates or []:
```
elliotgunton, line 136: `self.templates` is already defaulted to `[]` - is the `or []` necessary?

With the `templates=templates or None` used below in the `_ModelWorkflow` constructor, I'm not sure this code is the clearest as it's hiding away the `or None` below. IMO showing that templates can be `[]` or `None` up front makes the rest a bit clearer

```py
def build(self) -> _ModelWorkflow:
     templates = [] if self.templates else None
     for template in self.templates:
     # ...
```

tests/test_artifact.py
```diff
@@ -53,7 +53,11 @@ def test_output_artifact_contains_expected_fields(self):
         path = "/input/path"
         from_task = "produce-artifact"
 
-        expected = Artifact(name=name, path=path, from_task=f"{{{{tasks.{from_task}.outputs.artifacts.{name}}}}}")
+        expected = Artifact(
+            name=name,
+            path=path,
+            from_task=f"{{{{tasks.{from_task}.outputs.artifacts.{name}}}}}",
```
elliotgunton, line 59: Not directly relevant to PR: is there a way to make these f-strings less horrible? 😅 Bracket overload 🤯

elliotgunton (commented):
Couple of small comments to address

tests/test_artifact.py
samj1912: These are not a part of the PR. The brackets can be avoided using hera.expr + fstrings. https://hera-workflows.readthedocs.io/expr.html#usage

src/hera/workflows/v5/task.py
samj1912: Copied from the existing task files, will add.

src/hera/workflows/v5/examples/upstream/dag_diamond.py
samj1912: This is the default behaviour from hera right now, happy to change it.

src/hera/workflows/v5/examples/upstream/dag_conditional_parameters.py
samj1912: They are synced unfortunately.

src/hera/workflows/v5/examples/coinflip.yaml
samj1912: the yaml is auto-generated from the python code, this happens because of the default global config.

elliotgunton (commented):
CI/CD broken from tests relating to the default namespace change - LGTM when that's fixed! 🚀

elliotgunton (approved):
LGTM 💯

src/hera/workflows/v5/_context.py
```diff
@@ -17,15 +17,11 @@ def enter(self, p: Subbable) -> None:
...
 
-        popped = self._pieces.pop()
-        main = self._pieces[0]
-        main._add_sub(popped)
-
-    def add_sub_node(self, node: Union[_SubNodeMixin, TTemplate, _DAGTaskMixin]) -> None:
-        self._pieces[-1]._add_sub(node)
+    def add_sub_node(self, node: Union[_SubNodeMixin, TTemplate]) -> None:
+        if self._pieces:
+            self._pieces[-1]._add_sub(node)
```
samj1912, line 24: Related to https://github.com/argoproj-labs/hera-workflows/pull/449#discussion_r1125715224

src/hera/workflows/v5/_mixins.py
```diff
@@ -45,23 +37,42 @@
...
+
+        This supports using `with Workflow(...)`.
+        """
+        from hera.workflows.v5._context import _context
+
+        _context.exit()
+
+    def _add_sub(self, node: Any) -> Any:
+        raise NotImplementedError()
+
```
samj1912, line 74: Related to https://github.com/argoproj-labs/hera-workflows/pull/449#discussion_r1125716579

src/hera/workflows/v5/_mixins.py
```diff
@@ -232,155 +244,3 @@ def _build_volumes(self) -> Optional[List[Volume]]:
         if self.volumes is None:
             return None
         return [v._build_volume() for v in self.volumes]
-
-
-class _DAGTaskMixin(_BaseMixin):
```
samj1912, line 237: This was moved to a Task class since Task/Step have different fields and the callable syntactic sugar doesn't account for both. Deferring that for now and we can always implement syntactic sugar later.

It doesn't change much. Things go from

```py
task = echo(...)
```

to

```py
task = Task(echo)
# or
step = Step(echo)
```

src/hera/workflows/v5/dag.py
```diff
@@ -1,51 +1,38 @@
...
-        self._task_queue.append(task)
+        self.tasks.append(node)
 
     def _build_template(self) -> _ModelTemplate:
+        tasks = []
+        for task in self.tasks:
+            if isinstance(task, Task):
+                tasks.append(task._build_dag_task())
+            else:
+                tasks.append(task)
```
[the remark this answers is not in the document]
samj1912: See https://github.com/argoproj-labs/hera-workflows/pull/449#discussion_r1125717724

src/hera/workflows/v5/examples/upstream/upstream.dag-diamond.yaml
```diff
@@ -0,0 +1,44 @@
+# The following workflow executes a diamond workflow
```
samj1912, line 1: This is synced automatically from upstream.

Convention is examples/upstream.<example>.yaml <> <example>.py

src/hera/workflows/v5/script.py
```diff
@@ -23,21 +23,21 @@
...
 ):
-    name: str
+    container_name: Optional[str] = None
     args: Optional[List[str]] = None
     command: Optional[List[str]] = None
     lifecycle: Optional[Lifecycle] = None
     security_context: Optional[SecurityContext] = None
     source: Optional[Union[Callable, str]] = None
     working_dir: Optional[str] = None
+    add_cwd_to_sys_path: bool = True
```
samj1912, line 40: Added this option so that users have control over whether they want to modify syspath or not. This leads to cleaner output scripts.

src/hera/workflows/v5/script.py
```diff
@@ -74,6 +74,7 @@ def _build_source(self) -> str:
             Final formatted script.
         """
         if callable(self.source):
+            self.command = self.command or ["python"]
```
samj1912, line 77: If the source is defined and command isn't, we now default to python if the source is a callable.

src/hera/workflows/v5/script.py
```diff
@@ -116,7 +119,51 @@ def _build_source(self) -> str:
             assert isinstance(self.source, str)
             return self.source
 
-    def _build_template(self) -> _ModelScriptTemplate:
+    def _build_template(self) -> _ModelTemplate:
```
samj1912, line 122: This fixes a bug with the Script template where it was returning a sub-object instead of the actual template.

src/hera/workflows/v5/workflow.py
```diff
@@ -125,7 +125,22 @@ def _set_workflows_service(cls, v):
...
         return v
 
+    @validator('kind', pre=True, always=True)
+    def _set_kind(cls, v):
+        if v is None:
+            return cls.__name__  # type: ignore
+        return v
+
     def build(self) -> _ModelWorkflow:
+        templates = []
```
samj1912, line 135: Deferred template resultion all the way to the end as well to ensure that we capture all the updates to added objects.

src/hera/workflows/v5/workflow.py
```diff
@@ -125,7 +125,22 @@ def _set_workflows_service(cls, v):
             return WorkflowsService()
         return v
 
+    @validator('kind', pre=True, always=True)
+    def _set_kind(cls, v):
+        if v is None:
+            return cls.__name__  # type: ignore
```
samj1912, line 131: We were missing the output `kind`.

src/hera/workflows/v5/workflow.py
samj1912, line 135: Deferred template resultion all the way to the end as well to ensure that we capture all the updates to added objects.

src/hera/workflows/v5/script.py
flaviuvadan: Maybe we can make this command a global param? 🤔 It would go well with `image`

src/hera/workflows/v5/script.py
```diff
@@ -126,12 +173,12 @@ def _build_template(self) -> _ModelScriptTemplate:
             image_pull_policy=self._build_image_pull_policy(),
             lifecycle=self.lifecycle,
             liveness_probe=self.liveness_probe,
-            name=self.name,
```
flaviuvadan, line 129: Is this necessary because of the clash with the template name field, which is required?

samj1912: Precisely!

flaviuvadan: Small: what do you think about setting `template_name` on the template mixin rather than `container_name` on the subclass? I _think_ that might be easier for users since there's a single name field to set on the template. By comparison, having `container_name` does not fully align with the concept of a task/step 🤔

samj1912: gotcha - I don’t think the script name is important/actually used in any of the examples. The reason I called it container_name was because of 
> name	string	Name of the container specified as a DNS_LABEL. Each container in a pod must have a unique name (DNS_LABEL). Cannot be updated.

from https://argoproj.github.io/argo-workflows/fields/#scripttemplate

Given that the `name` field users generally want to set is the template name, I wanted to keep that short/simple.

flaviuvadan: Got it! What do you think about making the name the template name if the `container_name` is not specified? 🤔

flaviuvadan (approved):

PR #452, opened 2023-03-06
Add suspend

flaviuvadan:


samj1912 (commented):
@flaviuvadan could we add some examples to the v5 folder so that this can be tested? Maybe some of the suspend + intermediate input examples?

src/hera/workflows/v5/suspend.py
```diff
@@ -0,0 +1,65 @@
...
+    SuspendTemplate as _ModelSuspendTemplate,
+    Template as _ModelTemplate,
+)
+from hera.workflows.v5._mixins import _SubNodeMixin, _TemplateMixin
+from hera.workflows.v5.parameter import Parameter
+
+
+class Suspend(_TemplateMixin, _SubNodeMixin):
+    duration: Optional[Union[int, str]] = None
+    intermediate_parameters: List[Parameter] = []
```
elliotgunton, line 15: These are in effect mutually exclusive (using a default in the output parameter along with a duration) due to an issue upstream - see https://github.com/argoproj/argo-workflows/discussions/7905 - is it worth making a note or enforcing that in this class?

src/hera/workflows/v5/suspend.py
```diff
@@ -0,0 +1,65 @@
...
+        return Outputs(parameters=outputs) if outputs else None
+
+    def _build_inputs(self) -> Optional[Inputs]:
+        inputs = []
+        for param in self.intermediate_parameters:
+            inputs.append(param.as_input())
+        return Inputs(parameters=inputs) if inputs else None
+
+    def _build_template(self) -> _ModelTemplate:
+        print(self._build_inputs())
```
elliotgunton, line 37: leftover print

src/hera/workflows/v5/suspend.py
samj1912: I think just noting is fine for now since it is a bug. The two things should ideally work in conjunction.

samj1912 (approved):

src/hera/workflows/v5/suspend.py
elliotgunton: NB a Hera user can't give an output valueFrom a default value right now as we build it for them, see `_build_outputs` [below](https://github.com/argoproj-labs/hera-workflows/pull/452/files#diff-fab7239714c9572b508d61fc88bb31547561268c4960db565a13e7f4450b8d49R26), so we (hopefully) won't get complaints from users. It does mean we'll need to revisit this class if/when it gets fixed upstream (the issue doesn't seem very known for now).

elliotgunton (approved):
LGTM

PR #453, opened 2023-03-06
Add HTTP

flaviuvadan:


src/hera/workflows/v5/_mixins.py
```diff
@@ -170,6 +170,7 @@ def _build_env_from(self) -> Optional[EnvFromSource]:
 
 
 class _TemplateMixin(_BaseMixin):
+    name: str
```
elliotgunton, line 173: Mentioned on Sam's PR - [inline templates don't need a name](https://github.com/argoproj-labs/hera-workflows/pull/451#discussion_r1126197475) - up to you if you want to enforce naming within Hera code? The `Template` model class itself also has `name: Optional[str]`.

elliotgunton (approved):
LGTM

src/hera/workflows/v5/http.py
```diff
@@ -0,0 +1,60 @@
...
+
+class HTTP(_TemplateMixin, _SubNodeMixin):
+    url: str
+    body: Optional[str] = None
+    body_from: Optional[HTTPBodySource] = None
+    headers: Optional[List[HTTPHeader]] = None
+    insecure_skip_verify: Optional[bool] = None
+    method: Optional[str] = None
+    success_condition: Optional[str] = None
+    timeout_seconds: Optional[int] = None
```
elliotgunton, line 20: This class is a good candidate for a discussion topic: are we increasing the maintenance burden on this repo (i.e. Hera devs) by copy/pasting a subset of member variables from the `Template` model class?

Could we instead subclass it (for example), which gives us the Pydantic `Field` info, and reduces the need for us to keep things in sync. If we could subclass `Template`, somehow bring forward the relevant fields for code completion when creating an `HTTP` object, and ensure mutually exclusive fields from the parent `Template` class aren't available, this might be a cleaner solution? Then the `_build` methods could return a copy of `self` with certain fields build (here it would be `http` and `metadata`.

Obviously I haven't fully fleshed out the alternatives - just an idea to bounce off for a discussion (slack or an issue?) 🚀

samj1912:
Closing in favor of #454

PR #455, opened 2023-03-06
Add resource template

flaviuvadan:


src/hera/workflows/v5/dag.py
```diff
@@ -44,15 +51,20 @@ def _build_template(self) -> _ModelTemplate:
             name=self.name,
             node_selector=self.node_selector,
             outputs=self._build_outputs(),
+            parallelism=self.paralellism,
```
flaviuvadan, line 54:
```suggestion
            paralellism=self.paralellism,
```

elliotgunton:
From this PR do we only want `resource.py`? The rest of the changes seem to conflict with the existing code on the v5 branch or have already been added (data, http and suspend). cc @samj1912

PR #456, opened 2023-03-08
Add Steps and ParallelSteps

elliotgunton:
* Adds the `Steps` and `ParallelSteps` implementations as context managers
* A single `Step` must appear within one of these context managers, and `ParallelSteps` itself
  must appear under a `Steps` context
* A `Step`'s `arguments` are declared in a single list i.e. a union of `Artifacts` and `Parameters`


src/hera/workflows/v5/steps.py
```diff
@@ -0,0 +1,145 @@
...
+            host_aliases=self.host_aliases,
+            init_containers=self.init_containers,
+            inputs=self._build_inputs(),
+            memoize=self.memoize,
+            metadata=self._build_metadata(),
+            metrics=self.metrics,
+            name=self.name,
+            node_selector=self.node_selector,
+            outputs=self._build_outputs(),
+            parallelism=self.paralellism,
```
samj1912, line 128:
```suggestion
            parallelism=self.parallelism,
```

src/hera/workflows/v5/steps.py
```diff
@@ -0,0 +1,146 @@
...
+        return [step._build_as_workflow_step() for step in self.parallel_steps]
+
+
+class Steps(
+    _ContextMixin,
+    _IOMixin,
+    _SubNodeMixin,
+    _TemplateMixin,
+):
+    workflow_steps: List[Union[Step, ParallelSteps]] = []
```
samj1912, line 90: Can we also add the step model here?

samj1912, line 90: Can we also allow a list of list of steps?

elliotgunton: Is it actually worth adding the model? It would be unintuitive to users that they have to call `_add_sub` as the model WorkflowStep can't be a `_SubNodeMixin` - so maybe we should add an accompanying manual `add_step` function (which just forwards to `_add_sub`)?

elliotgunton: It might not be clear that the list of steps has to be declared outside of the context manager
```py
my_steps = [Step(...), Step(...), Step(...)]  # treat as parallel steps

    s._add_sub(my_steps)
    s._add_sub(_ModelWorkflowStep(...))  # backup option to use WorkflowStep
    s._add_sub([_ModelWorkflowStep(...), _ModelWorkflowStep(...)])  # also allow List[WorkflowStep]?
```

elliotgunton: Updated comment - users can use `workflow_steps.append`

elliotgunton:
Need to rename member vars and add an `add_step` function

samj1912 (approved):

src/hera/workflows/v5/examples/upstream/steps.py
```diff
@@ -15,14 +15,14 @@
...
-    with Steps(name="hello-hello-hello"):
+    with Steps(name="hello-hello-hello") as s:
         Step(
             name="hello1",
             template="whalesay",
             arguments=[Parameter(name="message", value="hello1")],
         )
 
-        with ParallelSteps():
+        with s.parallel():
```
elliotgunton, line 25: Nice addition 💯 Makes it (more) obvious it's part of `s`.

Extending on this change - should we make `Parallel` a private class so it's not directly instantiable? i.e. what if I have `with Parallel` directly under a `with Workflow`?

PR #461, opened 2023-03-11
add resource, fix steps

flaviuvadan:


examples/workflows/k8s-resource-log-selector.yaml
```diff
@@ -0,0 +1,48 @@
+# This example demonstrates how to include and obtain logs from pods created by a
```
samj1912, line 1: @flaviuvadan since this is an upstream example, could you move the python file to `examples/workflow/upstream`? Then if you run `make regenerate-test-data` it should generate all required yaml files.

samj1912:
@flaviuvadan  btw there might be something weird with your DCO, your commits are authored using 

```
Flaviu Vadan <flaviuvadan@gmail.com>
```

but signed off using 
```
# Notice the different name
flaviuvadan <flaviuvadan@gmail.com>
```
 which causes DCO to fail

src/hera/workflows/steps.py
```diff
@@ -41,13 +41,17 @@ class Step(
...
+        arguments = (
+            None if model_arguments.artifacts is None and model_arguments.parameters is None else model_arguments
         )
-
         return _ModelWorkflowStep(
-            arguments=model_arguments,
+            arguments=arguments,
             continue_on=self.continue_on,
             hooks=self.hooks,
             inline=self.inline,
```
samj1912, line 57: Could we make step here accept a template object similar to task? That should simplify the examples.

examples/workflows/resource_delete_with_flags.py
```diff
@@ -0,0 +1,41 @@
...
+delete_resource = Resource(
+    name="delete-resource",
+    action="delete",
+    flags=["configmap", "--selector", "{{inputs.parameters.selector}}"],
+    inputs=Inputs(
+        parameters=[Parameter(name="selector")],
+    ),
+)
+
+with Workflow(
```
samj1912, line 28: Given that the templates are in order, should we just use the context handler to define all the templates inside the workflow instead of having to add them?

samj1912 (approved):

samj1912 (approved):

src/hera/workflows/steps.py
```diff
@@ -41,19 +41,23 @@ class Step(
...
             name=self.name,
             on_exit=self.on_exit,
-            template=self.template,
+            template=self.template if isinstance(self.template, str) else self.template.name,
```
elliotgunton, line 60: Thank you, was meaning to add this! 🙏

PR #464, opened 2023-03-12
Add a script to generate init files

samj1912:
Adds an auto-gen script to generate the top level `__init__` files with all the relevant imports.


flaviuvadan (commented):
@samj1912 the original v5 branch has an init generation script that adds imports from the `models` directory to the Hera workflows init as well. The (potentially wrong) steps that were implemented previously are:
- find the Hera objects
- find the models objects
- find their intersection and only keep the Hera objects in the init
- include the diff in objects from Hera objects and models objects, respectively

What do you think about that approach? Do you think that makes it easier for users to access Hera autogen models + Hera provided objects?

samj1912:
I would personally prefer if we kept the auto-generated models under the models directory and only kept the hand written ones at the top level, mostly because they behave slightly differently and this signals the user on what's auto generated vs what's hand-written code from hera. Otherwise users might confuse for eg. why one container class works with context managers and why the other doesn't. 

In the examples, I have updated them with importing hera.workflows.models as `m`. This allows users to use auto-generated models using `m.<ModelName>`.

If we see some common patterns emerge we can always add back the imports to init, but taking them away (in case we see confusion with usage) would become a b/w incompatible change.

samj1912:
The other reason for keeping them separate is let's say we add a hera specific version of a class in the future because we discovered some syntactic sugar that helps, we wouldn't be able to replace the auto-generated model class as it may have different behavior/parameters.

flaviuvadan (approved):

PR #465, opened 2023-03-12
Add container set for V5

flaviuvadan:


src/hera/workflows/container_set.py
```diff
@@ -0,0 +1,75 @@
...
+class ContainerSet(
+    IOMixin,
+    ContainerMixin,
+    EnvMixin,
+    TemplateMixin,
+    ResourceMixin,
+    VolumeMountMixin,
+):
+    containers: List[Union[Container, ContainerNode]]
+    container_set_retry_strategy: Optional[ContainerSetRetryStrategy] = None
```
flaviuvadan, line 32: @samj1912 and @elliotgunton this is required because of the conflict with `retry_strategy` on the `TemplateMixin`. I figured adding a prefix here is better than adjusting the template mixing since that will push a `template_retry_strategy` downstream to all inheriting classes. This way the "specialized" template class sets the field it needs

examples/workflows/workspace_workflow.py
```diff
@@ -0,0 +1,35 @@
...
+    annotations={
+        "workflows.argoproj.io/description": "This workflow demonstrates using a workspace to share files between containers. "
+        'This also allows containers not called "main" to create output artifacts.',
+        "workflows.argoproj.io/version": ">= 3.1.0",
+    },
+) as w:
+    ContainerSet(
+        name="main",
+        containers=[
+            ContainerNode(
```
samj1912, line 19: Could we have container node be added to the set similar to how we add tasks to Dag or steps to Steps? (with the context manager)

examples/workflows/workspace_workflow.py
```diff
@@ -0,0 +1,35 @@
+from hera.workflows.container_set import ContainerNode, ContainerSet
```
samj1912, line 1: @flaviuvadan could you run `make examples` to ensure that the markdown gets created? I think something broke in the CI and it's not being caught for some reason.

samj1912 (approved):

src/hera/workflows/container_set.py
elliotgunton: Sensible! 💯

PR #472, opened 2023-03-13
Add dictionary support to Parameter init

flaviuvadan:


src/hera/workflows/_mixins.py
```diff
@@ -101,7 +101,11 @@ def _build_inputs(self) -> Optional[ModelInputs]:
 
         result = ModelInputs()
         for value in self.inputs:
-            if isinstance(value, Parameter):
+            if isinstance(value, dict):
```
samj1912, line 104: @flaviuvadan could we add a test for this?

samj1912: Can we also add similar support for arguments?

PR #475, opened 2023-03-14
Add WorkflowTemplate and CronWorkflow

elliotgunton:
* upstream example is reproducible with minimal WorkflowTemplate implementation (currently is just a named subclass of Workflow)
* the `inline` and `template` fields are mutually exclusive as described in the model DAGTask's inline field description

- [x] WorkflowTemplate
- [x] CronWorkflow


src/hera/workflows/cron_workflow.py
```diff
@@ -1 +1,86 @@
...
+
+class CronWorkflow(Workflow):
+    concurrency_policy: Optional[str] = None
+    failed_jobs_history_limit: Optional[int] = None
+    schedule: str
+    starting_deadline_seconds: Optional[int] = None
+    successful_jobs_history_limit: Optional[int] = None
+    cron_suspend: Optional[bool] = None
+    timezone: Optional[str] = None
+    cron_status: Optional[CronWorkflowStatus] = None
```
elliotgunton, line 33: Note the `cron_` prefix here when the actual field value is without (as it would shadow the workflow spec's suspend and status)

src/hera/workflows/cron_workflow.py
```diff
@@ -1 +1,87 @@
...
+                failed_jobs_history_limit=self.failed_jobs_history_limit,
+                schedule=self.schedule,
+                starting_deadline_seconds=self.starting_deadline_seconds,
+                successful_jobs_history_limit=self.successful_jobs_history_limit,
+                suspend=self.cron_suspend,
+                timezone=self.timezone,
+                workflow_metadata=None,
+                workflow_spec=super().build().spec,
+            ),
+            status=self.status,
```
elliotgunton, line 69:
```suggestion
            status=self.cron_status,
```

flaviuvadan (approved):
This looks really great!

PR #476, opened 2023-03-14
Isolate arguments in mixin

flaviuvadan:


src/hera/workflows/container.py
```diff
@@ -95,20 +92,5 @@ def _build_template(self) -> _ModelTemplate:
...
         )
 
-    def _build_arguments(self) -> Optional[Arguments]:
-        # parameters = [obj.as_argument() for obj in self.inputs if isinstance(obj, Parameter)]
-        # parameters = [p for p in parameters if p is not None]  # Some parameters might not resolve
-        if self.inputs is None:
-            return None
-
-        parameters = [p for p in self.inputs if isinstance(p, Parameter)]
-        artifacts = [a for a in self.inputs if isinstance(a, Artifact)]
```
elliotgunton, line 105: This is using `self.inputs`, not `self.arguments`

flaviuvadan: But the function was not used at all 🤔 that's why it's deleted here

elliotgunton: Lol didn't realise! Lgtm then 🤣

elliotgunton (approved):

PR #477, opened 2023-03-14
Workflow spec

elliotgunton:


elliotgunton:
Closing as I think we've solved this for now

PR #478, opened 2023-03-15
Make templates callable

elliotgunton:
- [x] Container
- [ ] Script

### DIFF
--- a/docs/examples/workflows/dag_with_callable_templates.md
+++ b/docs/examples/workflows/dag_with_callable_templates.md
@@ -1,0 +1,70 @@
+# Dag With Callable Templates
+
+
+
+
+
+## Hera
+
+```python
+from hera.workflows import (
+    DAG,
+    Container,
+    Parameter,
+    Workflow,
+)
+
+with Workflow(
+    generate_name="callable-templates-",
+    entrypoint="hello-world",
+) as w:
+    echo = Container(
+        name="echo",
+        image="alpine:3.7",
+        command=["echo", "{{inputs.parameters.message}}"],
+        inputs=[Parameter(name="message")],
+    )
+
+    with DAG(name="hello-world") as d:
+        (
+            echo(name="hello1", arguments=[Parameter(name="message", value="hello1")])
+            >> echo(name="hello2", arguments=[Parameter(name="message", value="hello2")])
+        )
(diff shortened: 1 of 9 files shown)


examples/workflows/upstream/dag_diamond.py
```diff
@@ -18,8 +17,8 @@
...
     )
     with DAG(name="diamond"):
-        A = Task(name="A", template=echo, arguments=m.Arguments(parameters=[Parameter(name="message", value="A")]))
-        B = Task(name="B", template=echo, arguments=m.Arguments(parameters=[Parameter(name="message", value="B")]))
-        C = Task(name="C", template=echo, arguments=m.Arguments(parameters=[Parameter(name="message", value="C")]))
-        D = Task(name="D", template=echo, arguments=m.Arguments(parameters=[Parameter(name="message", value="D")]))
+        A = Task(name="A", template=echo, arguments=[Parameter(name="message", value="A")])
+        B = Task(name="B", template=echo, arguments=[Parameter(name="message", value="B")])
+        C = Task(name="C", template=echo, arguments=[Parameter(name="message", value="C")])
+        D = Task(name="D", template=echo, arguments=[Parameter(name="message", value="D")])
```
samj1912, line 23: Can we replace this with the callable stuff as well?

src/hera/workflows/_mixins.py
```diff
@@ -284,3 +285,22 @@ def _build_arguments(self) -> Optional[ModelArguments]:
             parameters=parameters or None,
         )
         return model_arguments
+
+
+class CallableTemplateMixin:
+    def __call__(self, *args, **kwargs) -> SubNodeMixin:
```
flaviuvadan, line 291: What do you think about exposing the common Step/Task arguments here? That way users get auto-completion and we can further develop features for the callable, such as template parameter inference

elliotgunton: Based on Sam's comment in slack:
> We should additionally look into stub generation to better support auto complete

I'd prefer to do that in a new PR whether for 5.0.0 or 5.0.1 so that we can merge this PR as it is 🚀 

(I definitely think getting auto-complete working for callable templates would be a great QOL improvement!)

PR #437, opened 2023-02-20
[WIP] Hera V5

flaviuvadan:
Checklist:
- [x] Separate `workflows` and `events` into their own modules
- [x] Generate all `workflows` and `events` models
- [x] Add script to auto-generate the service for Workflows and Events
- [ ] Create release candidate based on new models
- [x] Adjust CodeCov to ignore auto-generated code
- [ ] Isolate template objects such as `Container`, `Script`, `Data`, etc, into their own objects
- [ ] Make more examples that showcase feature parity
- [ ] Documentation
- [ ] Tests
- [ ] Official release

CC:
- @elliotgunton 
- @samj1912

PR #479, opened 2023-03-16
Add Workflow Volume example

elliotgunton:
* We must pass a model Volume and model EmptyDirVolumeSource, possiblity for a Volume builder mixin?


examples/workflows/upstream/volumes_emptydir.py
```diff
@@ -0,0 +1,23 @@
+from hera.workflows import (
+    Container,
+    Workflow,
+    models as m,
+)
+
+with Workflow(
+    generate_name="volumes-emptydir-",
+    entrypoint="volumes-emptydir-example",
+    volumes=[m.Volume(name="workdir", empty_dir=m.EmptyDirVolumeSource())],
```
flaviuvadan, line 10: @elliotgunton we [already have a volume mixin](https://github.com/argoproj-labs/hera-workflows/pull/437/files#diff-079dae561586ca661f24348fd6560e265e3c6eba6650abd05491201598ae9288R233)! [Here's a usage pattern on container](https://github.com/argoproj-labs/hera-workflows/pull/437/files#diff-8ee20f78bd3bb4a719ab88b4388313de94453566413efd6a7d1d5927a8b2b770R28) and [how it's built](https://github.com/argoproj-labs/hera-workflows/pull/437/files#diff-8ee20f78bd3bb4a719ab88b4388313de94453566413efd6a7d1d5927a8b2b770R94). I think one thing that's missing at the moment, on workflow, is the automatic creation of PVCs for any volumes that are created i.e. if I set a `Volume(size=5Gi, mount_path=/mnt)` on a Step/Task template, workflow should perform a `.to_pvc` on the volume, on the user behalf

elliotgunton:
Merging as we can update volume mixins later (after v5 rc release)

PR #480, opened 2023-03-16
Added upstream examples for volume

Degiorgio:
- Adds two upstream examples for volume

### DIFF
--- a/examples/workflows/upstream/volumes-existing.py
+++ b/examples/workflows/upstream/volumes-existing.py
@@ -12,7 +12,7 @@ def get_container(name: str, args: List[str]) -> Container:
         args=args,
         volume_mounts=[
             m.VolumeMount(name="workdir", mount_path="/mnt/vol"),
-        ]
+        ],
     )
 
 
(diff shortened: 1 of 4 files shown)


examples/workflows/upstream/volumes-existing.py
```diff
@@ -0,0 +1,29 @@
...
+    )
+    Container(
+        name="print-accesslog",
+        image="alpine:latest",
+        command=["sh", "-c"],
+        args=["echo 'Volume access log:'; cat /mnt/vol/accesslog"],
+        volume_mounts=[
+            m.VolumeMount(name="workdir", mount_path="/mnt/vol"),
+        ],
+    )
```
samj1912, line 29: Can we create a function that wraps the common arguments here and returns the container object?

examples/workflows/upstream/volumes-existing.py
```diff
@@ -0,0 +1,29 @@
...
+from hera.workflows import models as m
+
+with Workflow(
+    generate_name="volumes-existing-",
+    entrypoint="volumes-existing-example",
+    volumes=[m.Volume(name="workdir", persistent_volume_claim={"claim_name": "my-existing-volume"})],
+) as w:
+    with Steps(name="volumes-existing-example") as s:
+        Step(name="generate", template="append-to-accesslog")
+        Step(name="print", template="print-accesslog")
```
samj1912, line 11: Can we use the new callable syntax here?

The template order does not matter so we can define our template first and then reference them there.

examples/workflows/upstream/volumes-existing.py
```diff
@@ -0,0 +1,29 @@
...
+
+    get_container("append-to-accesslog", ["echo accessed at: $(date) | tee -a /mnt/vol/accesslog"])
+    get_container("print-accesslog", ["echo 'Volume access log:'; cat /mnt/vol/accesslog"])
```
elliotgunton, line 29: We should encourage the container definitions to be before the steps, and here I added the callable syntax
```suggestion
    append_to_log = get_container("append-to-accesslog", ["echo accessed at: $(date) | tee -a /mnt/vol/accesslog"])
    print_log = get_container("print-accesslog", ["echo 'Volume access log:'; cat /mnt/vol/accesslog"])

    with Steps(name="volumes-existing-example") as s:
        append_to_log (name="generate")
        print_log (name="print")

    ```

File: examples/workflows/upstream/volumes-pvc.py [hunk 4]
```diff
@@ -0,0 +1,49 @@
...
+    get_container(
+        "whalesay",
+        "docker/whalesay:latest",
+        ["echo generating message in volume; cowsay hello world | tee /mnt/vol/hello_world.txt"],
+    )
+    get_container(
+        "print-message",
+        "alpine:latest",
+        ["echo getting message from volume; find /mnt/vol; cat /mnt/vol/hello_world.txt"],
+    )
```
elliotgunton, line 49: Containers can be first here too

elliotgunton (approved):
LGTM, looking into using a `VolumeMixin` in https://github.com/argoproj-labs/hera-workflows/pull/479 so may revisit these examples just for the volume/volume_mount assignments

PR #482, opened 2023-03-16
Add PVC inference 

flaviuvadan:
When users want to provision volumes dynamically, they have to create both a persistent volume and a persistent volume claim. The claim is set at the Workflow level in order to tell K8S to provision specific volumes for the workflow. It can be cumbersome to set both a volume claim and a volume. This PR allows users to create a Volume() specification on a templatable object. Then, the workflow will automatically create the PVC that is required for the volume that is added by the user. Note that PVCs need to be unique. Hence, the workflow only adds PVCs that are not already added by the user, based on the name of the PVC, since PVC names have to be unique

### DIFF
--- a/docs/examples/workflows/upstream/volumes_pvc.md
+++ b/docs/examples/workflows/upstream/volumes_pvc.md
@@ -46,43 +46,49 @@ spec:
 ## Hera
 
 ```python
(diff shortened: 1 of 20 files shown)


samj1912:
@flaviuvadan could you add some details in the PR description on what this change does? I'm not able to figure it out looking at the code.

PR #483, opened 2023-03-16
Add hooks for V5

flaviuvadan:


samj1912:
@flaviuvadan maybe I'm missing something but I don't see the updates related to post init where hera hooks are called?

src/hera/workflows/_base_model.py
```diff
@@ -24,3 +24,6 @@ def _init_private_attributes(self):
 
     def __hera_init__(self):
         ...
+
+    def __hera_hooks__(self):
+        ...
```
samj1912, line 29: @flaviuvadan this needs to be added after L23

src/hera/workflows/container.py
```diff
@@ -34,6 +35,9 @@ class Container(
     security_context: Optional[SecurityContext] = None
     working_dir: Optional[str] = None
 
+    def __hera_hooks__(self):
+        GlobalConfig.dispatch_hooks(self)
```
samj1912, line 39: @flaviuvadan we might have a circular import issue if we import global config here and import these objects back in global config. Can we break this dependency by introducing a protocol in a different third module and using that in appropriate places.

flaviuvadan: I will add some tests. I do not think this is the case because we import the objects in the config file only when performing type checking

flaviuvadan: Ah, no, I was wrong.. the circular import is a problem because we have `isinstance` calls

flaviuvadan:
@samj1912 added tests and added a protocol as you suggested. The tests demonstrate how hooks can be set up, and how they are automatically applied to a workflow and container, for instance :) going to merge this and we can apply any desirable modifications prior to rc1

PR #484, opened 2023-03-17
Support dynamic image lookup

JacobHayes:
I have images that are tagged by git sha. When launching argo jobs from a long running process (eg: a notebook, dev server, etc) and committing code, the default image needs to be pulled dynamically.

This updates `GlobalConfig.image` to support setting a function that will be called on access, similar to `GlobalConfig.token`.

### DIFF
--- a/src/hera/global_config.py
+++ b/src/hera/global_config.py
@@ -37,13 +37,13 @@ class _GlobalConfig:
 
     # protected attributes are ones that are computed/go through some light processing upon setting or
     # are processed upon accessing. The rest, which use primitive types, such as `str`, can remain public
+    _image: Union[str, Callable[[], str]] = "python:3.7"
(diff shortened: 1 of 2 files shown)


JacobHayes:
I always break coverage, but [this time](https://github.com/argoproj-labs/hera-workflows/actions/runs/4449488151/jobs/7813807068) it's not some rounding issue haha: `Invalid PEP 440 version: '0.0.0+'`

samj1912:
@JacobHayes yes, I think we pushed a fix to our v5 branch but not our main branch. We plan on releasing v5 soon. Would you like to submit this PR against v5 instead? I will also send a PR to fix v4 but if you don't care about this being back ported to v4, I would recommend v5 instead as that is going to be much better than v4.

flaviuvadan (approved):

PR #489, opened 2023-03-20
Add decorators for pre-build hooks

samj1912:
Adds pre-build hooks for hera to allow users to specify typed callables they can use to modify the respective objects pre-build.

Also moves a few files/variables around before release to make them easily accessible.

- All models now share a common BaseModel defined in `hera.shared._base_model`
- global config was brought up to `hera.shared` and is now called `global_config` to signify that it is a variable and not a class in itself.


src/hera/workflows/_mixins.py
```diff
@@ -74,6 +74,15 @@
...
 EnvFromT = Optional[Union[Union[_BaseEnvFrom, EnvFromSource], List[Union[_BaseEnvFrom, EnvFromSource]]]]
 TContext = TypeVar("TContext", bound="ContextMixin")
+THookable = TypeVar("THookable", bound="HookMixin")
+
+
+class HookMixin(BaseMixin):
+    def _dispatch_hooks(self: THookable) -> THookable:
+        output = self
+        for hook in global_config._get_pre_build_hooks(output):
+            output = hook(output)
```
flaviuvadan, line 84: Is it necessary for the hook to return the output? I wonder if there are instances in which users will forget to actually return the object, and will result in a `None` type _somewhere_. I think it might be easier for users to design a hook that takes an object and modifies the reference/applies changes in place

src/hera/shared/_global_config.py
```diff
@@ -0,0 +1,66 @@
...
+        self._pre_build_hooks.setdefault(return_type, []).append(hook)
+        return hook
+
+    def _get_pre_build_hooks(self, instance: TBase) -> List[Hook]:
+        """Registers a hook to be called before building a model."""
+        self._pre_build_hooks = self._pre_build_hooks or {}
+        return self._pre_build_hooks.get(type(instance)) or []
+
+
+global_config = _GlobalConfig()
```
flaviuvadan, line 65: The upper case is nice for a global config IMO 🤔 ```suggestion
global_config = _GlobalConfig()
GlobalConfig = global_config
```

flaviuvadan (approved):
Overall, LGTM! I left a question and a suggestion:
- I think it would be nice to have the `GlobalConfig` accessible
- it might be better to not force users to return the output of a hook for otherwise we risk returning `None` from the hooks

PR #491, opened 2023-03-20
Add dynamic image setter

flaviuvadan:
CC: @JacobHayes 
See: https://github.com/argoproj-labs/hera-workflows/pull/484 for V4


src/hera/shared/global_config.py
```diff
@@ -33,7 +33,7 @@ class _GlobalConfig:
     verify_ssl: bool = True
```
samj1912, line 33: @flaviuvadan does it make sense to allow other properties to have a similar behavior? For eg. `host` or `service_account_name`?

flaviuvadan: Yes, I think so! However, I have not heard a request for that _yet_ so I am inclined to introduce it when someone submits an issue for it

PR #493, opened 2023-03-20
Bring back the global config object and raise exceptions in case of nones from hooks

samj1912:
Addresses comments from https://github.com/argoproj-labs/hera-workflows/pull/489#pullrequestreview-1349469654


flaviuvadan (approved):
Thank you <3

PR #494, opened 2023-03-20
Add a pre-release workflow

samj1912:
With this our typical workflow should be -

- Set the version number/release tag
- Mark as pre-release so that it gets uploaded to test-pypi.
- Check if everything looks good
- Mark as release so that the same version gets uploaded to pypi

### DIFF
--- a/.github/workflows/pre-release.yml
+++ b/.github/workflows/pre-release.yml
@@ -1,0 +1,26 @@
+---
+name: Test PyPI publish
+
+on:
+  release:
+    types: [prereleased]
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Python 3.8
+        uses: actions/setup-python@v4
+        with:
+          python-version: 3.8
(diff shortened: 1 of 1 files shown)


.github/workflows/pre-release.yml
```diff
@@ -0,0 +1,26 @@
...
+        run: |
+          python -m pip install --upgrade pip
+          python -m pip install poetry
+      - name: Bump version number
+        run: poetry version ${{ github.event.release.tag_name }}
+      - name: Build and publish
+        run: |
+          poetry build
+          poetry config repositories.testpypi https://test.pypi.org/legacy/
+          poetry publish -r testpypi -u __token__ -p ${{ secrets.TEST_PYPI_PASSWORD }}
```
samj1912, line 26: @flaviuvadan you will need to add this secret for test pypi.

flaviuvadan, line 26:
```suggestion
          poetry publish -r testpypi -u __token__ -p ${{ secrets. PYPI_TEST_API_TOKEN }}
```

PR #495, opened 2023-03-21
Update README and examples

elliotgunton:
* Reword comparison to use past tense for Couler/Argo DSL
* Move code comparison into an expandable section
* Update coinflip example to be complete


elliotgunton:
Does it make sense to force the `semver:patch` label if we're just updating docs? 🤔

samj1912 (approved):

samj1912:
@elliotgunton when we eventually release the version, it needs to be a strictly greater version. The semver labels are aggregated between the current release and the next release and the highest semver label is applied to generate the next release version.

PR #499, opened 2023-03-22
use MISSING vs None in parameter inference

flaviuvadan:
CC: @JacobHayes 

Problem: when `None` is used for a parameter it gets serialized as `null`. This forces the parameter inference to get a default value of `null` for inferred parameters. Therefore, this leads Hera to not set the non default parameters, which are required to automatically create the Parameter values that have `{{item.X}}` as their defaults when users pass in `with_param`. Hence, the last check in parameter inference raises a key error because no parameters are identified to contain `{{item...`


JacobHayes:
Ah, thanks for the fix!

PR #504, opened 2023-03-26
Add type-hints to ensure IDEs provide right auto-completion

samj1912:
Currently pydantic auto-complete uses the alias field in the auto-complete. This is not ideal as it is not the actual field name that is used in the code. This commit adds type-hints to the models to ensure that the IDEs provide the right auto-complete.

## Before

image

## After

image


flaviuvadan (approved):
🎖️

PR #507, opened 2023-03-26
Add decorator for script

flaviuvadan:
This PR adds a decorator called `script`. The two are the same. The main benefit of having a decorator is so that users do not have to instantiate a function with dunders to avoid name conflicts with templates. Using a decorator allows users to bypass that step while having the full flexibility of script. If users do not wish to use the decorator, they still have the liberty to use `Script`, of course! This also modifies the global config to use default as the namespace, which was always the case for Hera. Users who have simple Argo setups likely submit to the default namespace, and this facilitates that by removing the need to set default in the first place. For users who have different namespaces, they of course still have the liberty to set the namespace. This PR updated a few upstream examples to use the script/task decorator in order to have a few comparisons with upstream to ensure that the mix of use cases (Script vs script vs task) is covered and 1:1 with upstream examples

## Example
```python
from hera.workflows import Workflow, script, DAG


@script()
def flip():
    import random

    result = "heads" if random.randint(0, 1) == 0 else "tails"
    print(result)


@script()
def heads():
    print("it was heads")


@script()
def tails():
    print("it was tails")


with Workflow(generate_name="coinflip-", entrypoint="d") as w:
    with DAG(name="d") as s:
        f = flip()
        heads().on_other_result(f, "heads")
        tails().on_other_result(f, "tails")

```


src/hera/shared/_global_config.py
```diff
@@ -24,7 +24,7 @@ class _GlobalConfig:
     host: Optional[str] = None
     verify_ssl: bool = True
     api_version: str = "argoproj.io/v1alpha1"
-    namespace: Optional[str] = None
+    namespace: Optional[str] = "default"
```
flaviuvadan, line 27: This is worthy of review

src/hera/workflows/_context.py
```diff
@@ -23,15 +23,27 @@ def __init__(self) -> None:
...
+            raise InvalidType(type(p))
         self._pieces.append(p)
 
     def exit(self) -> None:
         self._pieces.pop()
 
     def add_sub_node(self, node: Union[SubNodeMixin, TTemplate]) -> None:
         if self._pieces:
-            self._pieces[-1]._add_sub(node)
+            try:
```
flaviuvadan, line 34: This is worthy of review

src/hera/workflows/_mixins.py
```diff
@@ -424,8 +424,7 @@ def __call__(self, *args, **kwargs) -> SubNodeMixin:
             return Task(*args, template=self, **kwargs)
         except InvalidType:
             pass
-
-        raise InvalidTemplateCall("Container is not under a Steps, Parallel, or DAG context")
+        return None
```
flaviuvadan, line 427: This is worthy of review

src/hera/workflows/script.py
```diff
@@ -224,4 +223,18 @@ def _get_parameters_from_callable(source: Callable) -> Optional[List[Parameter]]
     return [Parameter(name=n, default=v) for n, v in source_signature.items()]
 
 
-__all__ = ["Script"]
+def script(**script_kwargs):
```
flaviuvadan, line 226: This is worthy of review

tests/test_workflow.py
```diff
@@ -36,6 +36,7 @@ def _transform_workflow(obj):
     w = ModelWorkflow.parse_obj(obj)
     w.metadata.annotations = {}
     w.metadata.labels = {}
+    w.metadata.namespace = None
```
flaviuvadan, line 39: This is worthy of review

tests/test_workflow.py
```diff
@@ -50,6 +51,7 @@ def _transform_cron_workflow(obj):
     wt.spec.workflow_spec.templates.sort(key=lambda t: t.name)
     wt.metadata.annotations = {}
     wt.metadata.labels = {}
+    wt.metadata.namespace = None
```
flaviuvadan, line 54: This is worthy of review

src/hera/shared/_global_config.py
elliotgunton: I think we had this as `None` because from [the Argo spec](https://argoproj.github.io/argo-workflows/fields/#objectmeta), a missing namespace *is* "default":
>Namespace defines the space within which each name must be unique. An empty namespace is equivalent to the "default" namespace, but "default" is the canonical representation. Not all objects are required to be scoped to a namespace - the value of this field for those objects will be empty.Must be a DNS_LABEL. Cannot be updated. More info: http://kubernetes.io/docs/user-guide/namespaces

So I think it's fine either way but I would argue it's an extra bit of logic to maintain!

flaviuvadan: I did not know that! Thanks for pointing out. The only reason I added it, tbh, is that upon submitting workflows via `w.create()` I got errors on local that `namespace` is `None`. How about we keep it for that specific use case but we also link this documentation as justification for adding `default`?

samj1912: I would prefer if we did not default to the `default` namespace if possible. 

It makes it difficult to provide re-usable/copy-pastable examples or code.

Related - https://release.com/blog/10-kubernetes-namespace-best-practices-to-start-following#:~:text=Don't%20Use%20the%20Default,or%20implement%20RBAC%20and%20NetworkPolicies.
image

If no namespace is provided it would default to the `default` namespace as @elliotgunton mentioned.

src/hera/workflows/http_template.py
samj1912, line 1: @flaviuvadan did you have to rename this because of some name conflict?

src/hera/workflows/_mixins.py
samj1912, line 427: @flaviuvadan curious around the reasoning behind this change. Could you provide some color?

docs/examples/workflows/upstream/colored_logs.md
```diff
@@ -29,15 +30,10 @@ def colored_logs():
...
-        source=colored_logs,
-        image="python:3.7",
-        add_cwd_to_sys_path=False,
-        env=[Env(name="PYTHONUNBUFFERED", value="1")],
-    )
-    w.entrypoint = s.name
+with Workflow(generate_name="colored-logs-", entrypoint="whalesay") as w:
+    whalesay(name="whalesay")
+
+print(w.to_yaml())
```
samj1912, line 36: Leftover print?

src/hera/workflows/_context.py
samj1912: @flaviuvadan it would be great to have some comments explaining these steps for future contributions.

src/hera/workflows/_mixins.py
samj1912: Could we still raise the exception and catch it in the right place where it's expected?

src/hera/workflows/script.py
samj1912, line 226: Worth adding a doc string explaining the behavior of this decorator?

src/hera/workflows/script.py
```diff
@@ -224,4 +223,18 @@ def _get_parameters_from_callable(source: Callable) -> Optional[List[Parameter]]
...
+
+        def task_wrapper(**task_params):
+            return s.__call__(**task_params)
+
+        return task_wrapper
+
+    return script_wrapper
+
+
+task = script
```
samj1912, line 238: I would prefer if we didn't name script as task here. Might confuse users and might cause a hindrance if in the future we do want to add tasks of other types. It also doesn't match with the step type where it can also be used.

src/hera/workflows/http_template.py
flaviuvadan: Yea, exactly. I was getting circular imports because `requests` imports `urllib3`, and that one imports some `http` package. I am not sure exactly _why_ it pointed to _this_ `http` file but.. it's probably safer anyway. I will add a comment to the top of the file about this

src/hera/workflows/_mixins.py
flaviuvadan: This was leftover from testing, readded

samj1912 (approved):

PR #508, opened 2023-03-26
Switch to mkdocs for documentation

samj1912:
Moves the v5 documentation to mkdocs to keep it in-line with Argo docs. Also did a make-over of search/side-bar to make it parseable.


docs/versions.json
```diff
@@ -0,0 +1,18 @@
...
+            "v5"
+        ],
+        "version": "stable"
+    },
+    {
+        "title": "4.x",
+        "aliases": [
+            "v4"
+        ],
+        "version": "hera-v4-n-n"
```
flaviuvadan, line 16:
```suggestion
        "version": "v4"
```

flaviuvadan (approved):

PR #514, opened 2023-03-27
Add PayIt to hera adopters

menzenski:
We are using hera-workflows at [PayIt](https://payitgov.com/) to define all our Argo workflows.


menzenski:
> Thank you @menzenski! ❤️

Thank you @flaviuvadan and the other contributors for writing this tool! My team is happy to stay working in Python as much as we can.

PR #516, opened 2023-03-27
Publish to both `hera-workflows` and `hera` on PyPI

flaviuvadan:
The maintainers behind the original `hera` PyPi project were extremely kind and removed the project from PyPi because it has not been maintained since 2012. Therefore, we claim `hera` and switching from `hera-workflows` to `hera`. I already published rc5 to that project to avoid any potential bot taking the name ahead of this project

### DIFF
--- a/.github/workflows/cicd.yaml
+++ b/.github/workflows/cicd.yaml
@@ -4,12 +4,9 @@ on:
   push:
     branches:
       - main
-    paths:
-      - "src/**"
   pull_request:
-    paths:
-      - "src/**"
-
+    branches:
+      - main
 jobs:
   test:
     name: build + test py${{ matrix.python-version }} on ${{ matrix.os }}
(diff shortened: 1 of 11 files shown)


samj1912 (dismissed):

samj1912:
@flaviuvadan I believe CI might be failing because of outdated docs. But actions are not working. Can you please avoid merging until CI passes?

samj1912:
@flaviuvadan I think there are a couple of references to hera-workflows - can we make sure to tidy up all of them before we make the switch?

samj1912:
@flaviuvadan note - I didn't replace all the readme references yet. We can figure out how we want to handle the announcement around changes in package names. Meanwhile I have just updated the CI/CD to publish to both pypi projects so future releases can be consumed from either. We can make a call before 5.0 release on what name we want to keep as the canonical name.

Issue #498, opened 2023-03-21
Passing outputs from one step/task to the inputs of another


mshatkhin23:
In hera v4 I remember we had a well documented way of doing the above.  Something like:
```
task1 = Task(..., outputs=Parameter("output1", "hey"))
task2 = Task(..., inputs=Parameter("input1, task1.get_parameter("output1").value))
```

I'm not seeing any examples that show the same for v5.  The closest I've found is [here](https://github.com/argoproj-labs/hera-workflows/blob/main/examples/workflows/resource_flags.py#L35) which uses Argo simple expression syntax directly.  Can we add some documentation on best practices for this? 

(happy to contribute an example after input from v5 designers on how they thought about this)

elliotgunton:
Yeah currently the only syntactic sugar we have is for task results here https://github.com/argoproj-labs/hera-workflows/blob/af275245fb37206df15267c3a97a4511a5859118/src/hera/workflows/task.py#L81

We're aiming for syntactic sugar improvements like this for subsequent patches/minor versions but we might be able to add this today 🚀

PR #517, opened 2023-03-27
Add ability to get task output parameters, more syntactic sugar around complex dependencies, and easier way to define env vars

flaviuvadan:
Closes: #498 
CC: @mshatkhin23

### DIFF
--- a/docs/examples/workflows/dag_with_param_passing.md
+++ b/docs/examples/workflows/dag_with_param_passing.md
@@ -1,0 +1,75 @@
+# Dag With Param Passing
+
+
+
+
+
+
+=== "Hera"
+
+    ```python linenums="1"
+    from hera.workflows import DAG, Container, Parameter, Task, Workflow
+
+    with Workflow(generate_name="param-passing-", entrypoint="d") as w:
+        out = Container(
+            name="out",
+            image="docker/whalesay:latest",
+            command=["cowsay"],
+            outputs=Parameter(name="x", value=42),
+        )
+        in_ = Container(
+            name="in",
+            image="docker/whalesay:latest",
+            command=["cowsay"],
+            args=["{{inputs.parameters.a}}"],
+            inputs=Parameter(name="a"),
+        )
+        with DAG(name="d"):
+            t1 = Task(name="a", template=out)
+            t2 = Task(name="b", template=in_, arguments=t1.get_parameter("x").with_name("a"))
+            t1 >> t2
+    ```
+
+=== "YAML"
+
+    ```yaml linenums="1"
+    apiVersion: argoproj.io/v1alpha1
+    kind: Workflow
+    metadata:
+      generateName: param-passing-
+    spec:
+      entrypoint: d
+      templates:
+      - container:
+          command:
+          - cowsay
+          image: docker/whalesay:latest
+        name: out
+        outputs:
+          parameters:
+          - name: x
+            value: '42'
+      - container:
+          args:
+          - '{{inputs.parameters.a}}'
+          command:
+          - cowsay
+          image: docker/whalesay:latest
+        inputs:
+          parameters:
+          - name: a
+        name: in
+      - dag:
+          tasks:
+          - name: a
+            template: out
+          - arguments:
+              parameters:
+              - name: a
+                value: '{{tasks.a.outputs.parameters.x}}'
+            depends: a
+            name: b
+            template: in
+        name: d
+    ```
+
--- a/src/hera/workflows/parameter.py
+++ b/src/hera/workflows/parameter.py
@@ -5,7 +5,6 @@ for a tutorial on Parameters.
 """
 from __future__ import annotations
 
-import copy
 from typing import Any, Optional
 
 from pydantic import root_validator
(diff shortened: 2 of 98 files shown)


src/hera/workflows/task.py
```diff
@@ -145,6 +147,46 @@ def finished_at(self) -> str:
     def result(self) -> str:
         return f"{{{{tasks.{self.name}.outputs.result}}}}"
 
+    def get_parameters_as(self, name):
+        """Gets all the output parameters from this task"""
```
samj1912, line 151: @flaviuvadan couple of suggestions - can we instead introduce get_output? which gets an appropriate artifact or parameter?

The other thing we noticed is that `valueFrom` + expression is a better way to reference variables rather than using the substitution syntax. The latter doesn't work for inline template outputs.

Artifact similarly has a fromExpression field.

flaviuvadan: > can we instead introduce get_output? which gets an appropriate artifact or parameter?

Can one have the same name for a parameter and an artifact? If that is the case then `get_parameters` seems more precise 🤔 

> The other thing we noticed is that valueFrom + expression is a better way to reference variables rather than using the substitution syntax. The latter doesn't work for inline template outputs.

> Artifact similarly has a fromExpression field

I am not sure I fully grasp this - do you mind telling me more? My hunch is that you are referring to [this field](https://github.com/argoproj-labs/hera-workflows/blob/main/src/hera/workflows/artifact.py#L36) and implying that building an expression via `hera.expr` is better, perhaps?

samj1912: For the latter,

valueFrom: expression: tasks.name.output.result

works everywhere but

value: {{tasks.name.output.result}} doesn't

flaviuvadan: Oh, ok, got it

flaviuvadan:
@samj1912 adjusted the above!

docs/examples/workflows/dag_with_param_passing.md
```diff
@@ -0,0 +1,75 @@
...
+          - name: a
+        name: in
+      - dag:
+          tasks:
+          - name: a
+            template: out
+          - arguments:
+              parameters:
+              - name: a
+                value: '{{tasks.a.outputs.parameters.x}}'
```
samj1912, line 69: @flaviuvadan this should ideally be
```suggestion
                valueFrom:
                  expression: 'tasks.a.outputs.parameters.x'
```

src/hera/workflows/steps.py
```diff
@@ -62,6 +65,52 @@ def finished_at(self) -> str:
...
+        if obj is not None:
+            if isinstance(obj, _ModelParameter):
+                obj.value = f"{{{{steps.{self.name}.outputs.parameters.{name}}}}}"
+                return Parameter(
+                    name=obj.name,
+                    value=obj.value,
+                    value_from=obj.value_from,
+                    global_name=obj.global_name,
+                    description=obj.description,
+                )
```
samj1912, line 111: @flaviuvadan I am not sure what this code is doing. Can you explain a bit on what is happening here?

src/hera/workflows/task.py
```diff
@@ -145,6 +148,52 @@ def finished_at(self) -> str:
...
+        return Parameter(name=name, value_from=ValueFrom(expression=f"{{{{tasks.{self.name}.outputs.parameters}}}}"))
```
samj1912, line 153:
```suggestion
        return Parameter(name=name, value_from=ValueFrom(expression=f"{{tasks.{self.name}.outputs.parameters}}"))
```

docs/examples/workflows/dag_with_param_passing.md
flaviuvadan: I tried this but the Argo server was returning an error saying that parameter `a` is not set 😕

src/hera/workflows/steps.py
```diff
@@ -62,6 +65,52 @@ def finished_at(self) -> str:
     def result(self) -> str:
         return f"{{{{steps.{self.name}.outputs.result}}}}"
 
+    def get_parameters_as(self, name):
+        """Gets all the output parameters from this task"""
+        return Parameter(name=name, value_from=ValueFrom(expression=f"{{{{steps.{self.name}.outputs.parameters}}}}"))
```
samj1912, line 70:
```suggestion
        return Parameter(name=name, value=f"{{{{steps.{self.name}.outputs.parameters}}}}"))
```

src/hera/workflows/parameter.py
```diff
@@ -43,6 +49,12 @@ def __str__(self):
             raise ValueError("Cannot represent `Parameter` as string as `value` is not set")
         return self.value
 
+    def with_name(self, name: str) -> Parameter:
+        """Returns a copy of the parameter with the name set to the value"""
+        p = copy.deepcopy(self)
+        p.name = name
+        return p
```
samj1912, line 56:
```suggestion
        p = self.copy(deep=True)
        p.name = name
        return p
```

src/hera/workflows/steps.py
flaviuvadan: I removed the check for `_ModelParameter` because it's not useful - all the objects in parameters are of type `_ModelParameter` because the code invokes `build_template` just above. Here we assemble the identified output parameter and return a Hera `Parameter` type.

PR #543, opened 2023-04-01
Add a script constructor that allows for flexible script construction and add a robust runner mode

samj1912:

This PR introduces a new mode for script called callable mode, which runs the function using the hera runner.

hera runner is a robust version of the current script mode which can accomodate large files, imports and calling functions with kwargs. It is highly compatible/tied to pydantic and makes writing argo commands a breeze using pydantic + hera.

It automatically serializes/deserializes inputs/outputs using pydantic and also updates the script callable to only return a task when under an active hera context.

This mode is only possible if the user source code is installed in the output image along with hera.

---

After conversations with @flaviuvadan, updated the implementation to allow for the following - 

- Added a new ScriptConstructor class that allows users to define their own ScriptConstructor that can inspect the callable source and generate the output source. Provided two default implementations - inline and runner. The default value is `inline` right now, but there are placeholders in the code to allow for the default value to be detected automatically. Additionally, users can provide their own ScriptConstructor if they wish to and set it as a default.
- Added the ability to set class defaults for any hera class using global_config. This allows users to introduce default values of hera classes without needing us to implement it for every single use case. This also removes any circular dependencies between hera global config and any hera specific classes.

### DIFF
--- a/docs/examples/workflows/callable_script.md
+++ b/docs/examples/workflows/callable_script.md
@@ -22,6 +22,9 @@
     # and serializes the output.
     global_config.image = "my-image-with-python-source-code-and-dependencies"
     global_config.set_class_defaults(Script, constructor="runner")
+    # Runner script constructor is still and experimental feature and we need to explicitly opt in to it
+    # Note that experimental features are subject to breaking changes in future releases of the same major version
+    global_config.experimental_features["script_runner"] = True
 
 
     # An optional pydantic input type
--- a/docs/examples/workflows/callable_script.md
+++ b/docs/examples/workflows/callable_script.md
@@ -13,14 +13,16 @@
     from pydantic import BaseModel
 
     from hera.shared import global_config
-    from hera.workflows import Steps, Workflow, script
+    from hera.workflows import Script, Steps, Workflow, script
 
-    # Note, callable mode is only possible if the source code is available
+    # Note, setting constructor to runner is only possible if the source code is available
     # along with dependencies include hera in the image.
     # Callable is a robust mode that allows you to run any python function
     # and is compatible with pydantic. It automatically parses the input
     # and serializes the output.
     global_config.image = "my-image-with-python-source-code-and-dependencies"
+    global_config.set_class_defaults(Script, constructor="runner")
+
 
     # An optional pydantic input type
     # hera can automatically de-serialize argo
@@ -32,6 +34,7 @@
         a: int
         b: str = "foo"
 
+
     # An optional pydantic output type
     # hera can automatically serialize the output
     # of your function into a json string
@@ -42,7 +45,7 @@
         output: List[Input]
 
 
-    @script(callable=True)
+    @script()
     def my_function(input: Input) -> Output:
         return Output(output=[input])
 
@@ -50,7 +53,7 @@
     # Note that the input type is a list of Input
     # hera can also automatically de-serialize
     # composite types like lists and dicts
-    @script(callable=True)
+    @script()
     def another_function(inputs: List[Input]) -> Output:
         return Output(output=inputs)
 
@@ -58,7 +61,7 @@
     # it also works with raw json strings
     # but those must be explicitly marked as
     # a string type
-    @script(callable=True)
+    @script()
     def str_function(input: str) -> Output:
         # Example function to ensure we are not json parsing
         # string types before passing it to the function
--- a/docs/examples/workflows/callable_script.md
+++ b/docs/examples/workflows/callable_script.md
@@ -1,0 +1,147 @@
+# Callable Script
+
+
+
+
+
+
+=== "Hera"
+
+    ```python linenums="1"
+    from typing import List
+
+    from pydantic import BaseModel
+
+    from hera.shared import global_config
+    from hera.workflows import Steps, Workflow, script
+
+    # Note, callable mode is only possible if the source code is available
+    # along with dependencies include hera in the image.
+    # Callable is a robust mode that allows you to run any python function
+    # and is compatible with pydantic. It automatically parses the input
+    # and serializes the output.
+    global_config.image = "my-image-with-python-source-code-and-dependencies"
+
+    # An optional pydantic input type
+    # hera can automatically de-serialize argo
+    # arguments into types denoted by your function's signature
+    # as long as they are de-serializable by pydantic
+    # This provides auto-magic input parsing with validation
+    # provided by pydantic.
+    class Input(BaseModel):
+        a: int
+        b: str = "foo"
+
+    # An optional pydantic output type
+    # hera can automatically serialize the output
+    # of your function into a json string
+    # as long as they are serializable by pydantic or json serializable
+    # This provides auto-magic output serialization with validation
+    # provided by pydantic.
+    class Output(BaseModel):
+        output: List[Input]
+
+
+    @script(callable=True)
+    def my_function(input: Input) -> Output:
+        return Output(output=[input])
+
+
+    # Note that the input type is a list of Input
+    # hera can also automatically de-serialize
+    # composite types like lists and dicts
+    @script(callable=True)
+    def another_function(inputs: List[Input]) -> Output:
+        return Output(output=inputs)
+
+
+    # it also works with raw json strings
+    # but those must be explicitly marked as
+    # a string type
+    @script(callable=True)
+    def str_function(input: str) -> Output:
+        # Example function to ensure we are not json parsing
+        # string types before passing it to the function
+        return Output(output=[Input.parse_raw(input)])
+
+
+    with Workflow(name="my-workflow") as w:
+        with Steps(name="my-steps") as s:
+            my_function(arguments={"input": Input(a=2, b="bar")})
+            str_function(arguments={"input": Input(a=2, b="bar").json()})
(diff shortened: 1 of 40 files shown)


examples/workflows/callable_coinflip.py
```diff
@@ -0,0 +1,33 @@
+import random
```
flaviuvadan, line 1: The overall experience of this looks amazing!

src/hera/workflows/_context.py
```diff
@@ -29,6 +29,10 @@ def enter(self, p: Subbable) -> None:
     def exit(self) -> None:
         self._pieces.pop()
 
+    @property
+    def active(self) -> bool:
+        return bool(self._pieces)
+
```
samj1912, line 35: To check if context is active. Useful in determining what the call to a script decorated function should return.

src/hera/shared/serialization.py
```diff
@@ -1,13 +1,23 @@
...
+
+
 def serialize(value: Any):
     if value == MISSING:
         return None
     elif isinstance(value, str):
         return value
     else:
-        return json.dumps(value)  # None serialized as `null`
+        return json.dumps(value, cls=PydanticEncoder)  # None serialized as `null`
```
samj1912, line 23: Adds support for serializing pydantic models to json by default instead of raising a value error.

src/hera/shared/_global_config.py
```diff
@@ -74,6 +82,50 @@ def _get_pre_build_hooks(self, instance: TBase) -> List[Hook]:
...
+    def __hera_init__(self):
+        ...
+
+    @root_validator(pre=True)
+    def _set_defaults(cls, values):
+        defaults = global_config._get_class_defaults(cls)
+        for key, value in defaults.items():
+            if values.get(key) is None:
+                values[key] = value
+        return values
```
samj1912, line 127: This piece of code is what allows us to set global defaults on any hera class.

src/hera/shared/_base_model.py
```diff
@@ -12,19 +8,3 @@ class Config:
         use_enum_values = True
         arbitrary_types_allowed = True
         smart_union = True
-
-
-class BaseMixin(BaseModel):
```
samj1912, line 17: Was moved to `_global_config` since it was causing circular deps otherwise.

examples/workflows/volume-mounts.yaml
```diff
@@ -33,7 +33,7 @@ spec:
     script:
       command:
       - python
-      image: python:3.7
+      image: python:3.8
```
samj1912, line 36: Changed default python version to 3.8 as that's the minimum required version for hera.

docs/examples/workflows/callable_script.md
```diff
@@ -0,0 +1,150 @@
...
+        script:
+          args:
+          - -m
+          - hera.workflows.runner
+          - -e
+          - examples.workflows.callable_script:another_function
+          command:
+          - python
+          image: my-image-with-python-source-code-and-dependencies
+          source: '{{inputs.parameters}}'
```
samj1912, line 148: We put the input parameters in a source file and then parse it as json to avoid python string interpolation issues with quotes or brackets. This also leads to cleaner yamls which are readable.

src/hera/workflows/runner.py
```diff
@@ -0,0 +1,127 @@
...
+import json
+from pathlib import Path
+from typing import Any
+
+from pydantic import validate_arguments
+
+from hera.shared.serialization import serialize
+
+
+def _ignore_unmatched_kwargs(f):
```
samj1912, line 14: This ensures that we are only passing valid args to our callable.

src/hera/workflows/runner.py
```diff
@@ -0,0 +1,127 @@
...
+    Returns:
+        The parsed value.
+
+    """
+    if _is_str_kwarg_of(key, f):
+        return value
+    try:
+        return json.loads(value)
+    except json.JSONDecodeError:
+        return value
```
samj1912, line 60: We do and additional json.loads on all input params since argo can only contain string values.

If pydantic is passed a json encoded string, it will not coerce it. So we need to convert it to a dictionary which then lets pydantic coerce it into the right type.

samj1912: We do not json parse the type if the input argument is explicitly defined as string.

src/hera/workflows/script.py
```diff
@@ -31,6 +34,21 @@
...
+        raise NotImplementedError
+
+    def transform_inputs(self, cls: Type["Script"], values: Any) -> Any:
+        return values
+
+    def transform_script_template(self, instance: "Script", script: _ModelScriptTemplate) -> _ModelScriptTemplate:
+        return script
+
+    def transform_template(self, instance: "Script", template: _ModelTemplate) -> _ModelTemplate:
+        return template
```
samj1912, line 49: This is the general interface anyone can implement to create their own script constructor.

Hera allows users to pass a custom script constructor of their own or a string constant which corresponds to hera only script constructor (inline and runner)

src/hera/workflows/script.py
```diff
@@ -51,92 +69,36 @@ class Script(
...
-                return textwrap.dedent(script)
-        else:
-            assert isinstance(self.source, str)
-            return self.source
+    @root_validator
+    @classmethod
+    def _constructor_validate(cls, values):
+        constructor = values.get("constructor")
+        assert isinstance(constructor, ScriptConstructor)
+        return constructor.transform_inputs(cls, values)
```
samj1912, line 101: This allows users to modify input args based on the constructor type. This is the first script constructor hook.

src/hera/workflows/script.py
```diff
@@ -156,66 +118,74 @@ def _build_inputs(self) -> Optional[ModelInputs]:
...
+                image=self.image,
+                image_pull_policy=self._build_image_pull_policy(),
+                lifecycle=self.lifecycle,
+                liveness_probe=self.liveness_probe,
+                name=self.container_name,
+                ports=self.ports,
+                readiness_probe=self.readiness_probe,
+                resources=self._build_resources(),
+                security_context=self.security_context,
+                source=self.constructor.generate_source(self),
```
samj1912, line 178: This is the main method every script constructor MUST implement.

src/hera/workflows/script.py
```diff
@@ -156,66 +118,74 @@ def _build_inputs(self) -> Optional[ModelInputs]:
...
-            stdin=self.stdin,
-            stdin_once=self.stdin_once,
-            termination_message_path=self.termination_message_path,
-            termination_message_policy=self.termination_message_policy,
-            tty=self.tty,
-            volume_devices=self.volume_devices,
-            volume_mounts=self._build_volume_mounts(),
-            working_dir=self.working_dir,
+        assert isinstance(self.constructor, ScriptConstructor)
+        return self.constructor.transform_script_template(
```
samj1912, line 162: Some additional hooks to allow script constructors to modify things even further if they wish to influence other params.

src/hera/workflows/script.py
```diff
@@ -268,13 +238,105 @@ def script_wrapper(func: Callable) -> Callable:
...
-        def task_wrapper(**task_params) -> Union[Task, Step]:
+        def task_wrapper(**kwargs) -> Union[Task, Step]:
             """Invokes a `Script` object's `__call__` method using the given `task_params`"""
-            return s.__call__(**task_params)  # type: ignore
+            if _context.active:
+                return s.__call__(**kwargs)  # type: ignore
+            return func(**kwargs)
 
+        # Set the wrapped function to the original function so that we can use it later
+        task_wrapper.wrapped_function = func  # type: ignore
```
samj1912, line 248: We set the original function as an attribute to allow hera runner to correctly call the original function regardless of whether it was wrapped or not.

src/hera/workflows/script.py
```diff
@@ -31,6 +36,21 @@
 from hera.workflows.task import Task
 
 
+class ScriptConstructor(BaseMixin):
+    @abstractmethod
+    def generate_source(self, instance: "Script") -> str:
+        raise NotImplementedError
+
+    def transform_inputs(self, cls: Type["Script"], values: Any) -> Any:
```
flaviuvadan, line 44:
```suggestion
    def transform_values(self, cls: Type["Script"], values: Any) -> Any:
```

src/hera/workflows/script.py
```diff
@@ -31,6 +36,21 @@
...
+        return values
+
+    def transform_script_template(self, instance: "Script", script: _ModelScriptTemplate) -> _ModelScriptTemplate:
+        return script
+
+    def transform_template(self, instance: "Script", template: _ModelTemplate) -> _ModelTemplate:
```
flaviuvadan, line 50:
```suggestion
    def transform_template_post_build(self, instance: "Script", template: _ModelTemplate) -> _ModelTemplate:
```

src/hera/workflows/script.py
```diff
@@ -31,6 +36,21 @@
 from hera.workflows.task import Task
 
 
+class ScriptConstructor(BaseMixin):
```
flaviuvadan, line 39: Can this have some docs on what users need to implement primarily and what independent transforms can perform? Just a bit to provide some clarity

flaviuvadan (approved):

PR #544, opened 2023-04-01
Add check that auto-genned params do not have already set artifact names

flaviuvadan:
Noticed that when adding a param to a script and then setting an artifact as the value of that script input can result in both an artifact and a parameter added to the template. 

# Reproducible example (main)

```python
from hera.workflows import DAG, Artifact, Workflow, script


@script(outputs=Artifact(name="result", path="/tmp/result"))
def produce():
    import pickle

    result = "foo testing"
    with open("/tmp/result", "wb") as f:
        pickle.dump(result, f)


@script(inputs=Artifact(name="i", path="/tmp/i"))
def consume(i):
    import pickle

    with open("/tmp/i", "rb") as f:
        i = pickle.load(f)
    print(i)


with Workflow(generate_name="fv-test-", entrypoint="d") as w:
    with DAG(name="d"):
        p = produce()
        c = consume(arguments=Artifact(name="i", from_="{{tasks.produce.outputs.artifacts.result}}"))
        p >> c

print(w.to_yaml())
```

Above results in:
```yaml
...
      image: python:3.7
      source: "import os\nimport sys\nsys.path.append(os.getcwd())\nimport pickle\n\
        \nresult = \"foo testing\"\nwith open(\"/tmp/result\", \"wb\") as f:\n   \
        \ pickle.dump(result, f)\n"
  - inputs:
      artifacts:
      - name: i
        path: /tmp/i
      parameters:
      - name: i
    name: consume
    script:
      command:
...
```

# Post-fix (this branch, isolated to problem section)

```yaml
...
        \nresult = \"foo testing\"\nwith open(\"/tmp/result\", \"wb\") as f:\n   \
        \ pickle.dump(result, f)\n"
  - inputs:
      artifacts:
      - name: i
        path: /tmp/i
    name: consume
    script:
      command:
...
```

### DIFF
--- a/docs/examples/workflows/script_auto_infer.md
+++ b/docs/examples/workflows/script_auto_infer.md
@@ -1,0 +1,86 @@
+# Script Auto Infer
+
+
+
+
+
+
+=== "Hera"
+
+    ```python linenums="1"
+    from hera.workflows import DAG, Artifact, Workflow, script
+
+
+    @script(outputs=Artifact(name="result", path="/tmp/result"))
+    def produce():
+        import pickle
+
+        result = "foo testing"
+        with open("/tmp/result", "wb") as f:
+            pickle.dump(result, f)
+
+
+    @script(inputs=Artifact(name="i", path="/tmp/i"))
+    def consume(i):
+        import pickle
+
+        with open("/tmp/i", "rb") as f:
+            i = pickle.load(f)
+        print(i)
+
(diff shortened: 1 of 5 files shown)


samj1912:
@flaviuvadan can we add an example so it serves as a regression test?

samj1912 (approved):

PR #551, opened 2023-04-04
WorkflowsService doesn't respect verify_ssl=False

jplikesbikes:


jplikesbikes:
Since the major rewrite for v5 I'm not sure what needs to be done to pass ci. This is a one line change if someone with more knowledge of the system just wants to do it or walk me through it, either would be appreciated. Thanks

samj1912:
@jplikesbikes you will need to change https://github.com/argoproj-labs/hera/blob/main/scripts/service.py#L403 and run `make codegen`

samj1912 (approved):

KathrynFowler91:
Is it possible to add the same change to token and namespace also so that they also respect the `global_config`?

samj1912:
@KathrynFowler91 those properties should work. Do you have a reproducible example?

KathrynFowler91:
"""python
from hera.workflows import WorkflowsService

from omegaconf import OmegaConf

from src.utils.hera_utils import get_token  # create token


app_settings = OmegaConf.load("app_settings.yaml")

global_config.namespace = app_settings.argoNamespace
global_config.service_account_name = app_settings.argoServiceAccountName
global_config.host = app_settings.argoHost
global_config.token = get_token()
global_config.verify_ssl = True
list_templates = WorkflowsService().list_workflow_templates()
print(list_templates)
"""

The error gives: `TypeError: WorkflowsService.list_workflow_templates() missing 1 required positional argument: 'namespace'`

samj1912:
Ah, that seems to be an issue with the generated methods instead of the class itself. Can you please create an enhancement request for this?

KathrynFowler91:
Will do. Thank you for the quick response.

PR #554, opened 2023-04-05
Script.resources is not the same as ModelTemplate.resource

jplikesbikes:
This looks like a possible typo. I don't believe `self._build_resources(),` should be passed to the ModelTemplate constructor as resource.

The following code currently fails with a pydantic error, but works correctly with the constructor argument removed.

```python
from hera.workflows import Task, Workflow, Resources, script


@script(resources=Resources(memory_request='5Gi'))
def task_with_lots_of_memory():
    print('ok')


with Workflow(
    name="test",
    entrypoint='main',
) as test_wf:

    with DAG(name='main', ) as main:
        this_should_pass: Task = task_with_lots_of_memory()

if __name__ == "__main__":
    print(test_wf.to_yaml())
```

Error:
```
pydantic.error_wrappers.ValidationError: 1 validation error for Template
resource -> action
  field required (type=value_error.missing)
```

### DIFF
--- a/examples/workflows/script_with_resources.py
+++ b/examples/workflows/script_with_resources.py
@@ -1,0 +1,10 @@
+from hera.workflows import Resources, Workflow, script
+
+
+@script(resources=Resources(memory_request="5Gi"))
+def task_with_memory_request():
+    print("ok")
+
+
+with Workflow(generate_name="script-with-resources-", entrypoint="task-with-memory-request") as w:
+    task_with_memory_request()
--- a/docs/examples/workflows/script_with_resources.md
+++ b/docs/examples/workflows/script_with_resources.md
@@ -1,0 +1,51 @@
+# Script With Resources
+
+
+
+
+
+
+=== "Hera"
+
+    ```python linenums="1"
+    from hera.workflows import Resources, Workflow, script
+
+
+    @script(resources=Resources(memory_request="5Gi"))
+    def task_with_memory_request():
+        print("ok")
+
+
+    with Workflow(generate_name="script-with-resources-", entrypoint="task-with-memory-request") as w:
+        task_with_memory_request()
+    ```
+
+=== "YAML"
+
+    ```yaml linenums="1"
+    apiVersion: argoproj.io/v1alpha1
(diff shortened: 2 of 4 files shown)


samj1912:
@jplikesbikes could you please add an example with this test case as a regression test?

We automatically run through all of our examples. Please see - https://github.com/argoproj-labs/hera/blob/main/CONTRIBUTING.md#adding-new-workflow-tests

jplikesbikes:
@samj1912  Sure no problem. I followed the readme and added the test `script_with_resources.py`.
Am I missing a step for the docs?

examples/workflows/script_with_resources.py
```diff
@@ -0,0 +1,10 @@
+from hera.workflows import Resources, Workflow, script
+
+
+@script(resources=Resources(memory_request="5Gi"))
+def task_with_memory_request():
+    print("ok")
+
+
+with Workflow(generate_name="script-with-resources-", entrypoint="s") as w:
```
flaviuvadan, line 9: I think this should be 
```suggestion
with Workflow(generate_name="script-with-resources-", entrypoint="task-with-memory-request") as w:
```

`task_with_memory_request(name="s")` sets the name of the independent task/step but in this workflow this does not run as part of a task/step so the template takes the name of the function but parsed for Argo/K8s

I tried the YAML as a test: 
```yaml
kind: Workflow
metadata:
  generateName: script-with-resources-
spec:
  entrypoint: task-with-memory-request
  templates:
  - name: task-with-memory-request
    script:
      command:
      - python
      image: python:3.8
      resources:
        requests:
          memory: 5Gi
      source: |
                import os
                import sys
                sys.path.append(os.getcwd())
                print("ok")
```

flaviuvadan: @samj1912 maybe we should have a check in the workflow that sets the right entrypoint when a single template is added?

jplikesbikes: Oh I see. I've pushed an update.

samj1912 (approved):

Issue #552, opened 2023-04-04
Validation Error when creating a RetryStrategy


jplikesbikes:
I'm not terribly familiar with Pydantic, but I think there is an error with the RetryStrategy model

```python
from hera.workflows import RetryStrategy

if __name__ == "__main__":
    RetryStrategy(limit=3)
```

The above fails with
```
pydantic.error_wrappers.ValidationError: 2 validation errors for RetryStrategy
limit
  value is not a valid integer (type=type_error.integer)
limit
  str type expected (type=type_error.str)
```

lakhanjindam:
hey i can pick this up, if no one is working, can you assign to me ?

dmerrick:
I am also experiencing this issue

PR #558, opened 2023-04-06
Fix validation error for retry strategy

lakhanjindam:
Fix issue #552

### DIFF
--- a/src/hera/workflows/retry_strategy.py
+++ b/src/hera/workflows/retry_strategy.py
@@ -65,4 +65,3 @@ class RetryStrategy(_BaseModel):
 
 
 __all__ = ["RetryPolicy", "RetryStrategy"]
-r = RetryStrategy(limit=3)
--- a/src/hera/workflows/retry_strategy.py
+++ b/src/hera/workflows/retry_strategy.py
@@ -36,7 +36,7 @@ class RetryStrategy(_BaseModel):
     affinity: Optional[RetryAffinity] = None
     backoff: Optional[Backoff] = None
     expression: Optional[str] = None
-    limit: Optional[IntOrString] = None
+    limit: Optional[Union[str, int, IntOrString]] = None
     retry_policy: Optional[Union[str, RetryPolicy]] = None
 
     @validator("retry_policy", pre=True)
@@ -52,7 +52,7 @@ class RetryStrategy(_BaseModel):
         if v is None or isinstance(v, IntOrString):
             return v
 
-        return IntOrString(__root__=str(v))  # int or str
+        return str(v)  # int or str
 
     def build(self) -> _ModelRetryStrategy:
         return _ModelRetryStrategy(
@@ -65,3 +65,4 @@ class RetryStrategy(_BaseModel):
 
 
 __all__ = ["RetryPolicy", "RetryStrategy"]
+r = RetryStrategy(limit=3)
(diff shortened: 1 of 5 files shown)


src/hera/workflows/retry_strategy.py
```diff
@@ -36,7 +36,7 @@ class RetryStrategy(_BaseModel):
     affinity: Optional[RetryAffinity] = None
     backoff: Optional[Backoff] = None
     expression: Optional[str] = None
-    limit: Optional[Union[int, str]] = None
+    limit: Optional[IntOrString] = None
```
flaviuvadan, line 39: Thanks for submitting the PR @lakhanjindam! 🚀 

There might be something strange happening here.. there's a `validator` for this field that is supposed to convert the `limit` from `str | int` to `IntOrString`

lakhanjindam: problems seems to be happening with this line https://github.com/argoproj-labs/hera/blob/main/src/hera/workflows/retry_strategy.py#L55 
since i made change where `limit` is mapped to `IntOrString` type so it goes into if block and returns from there.

lakhanjindam: what we can do is instead of that line we can add something like 
```python
def _convert_limit(cls, v):
  if v is None or isinstance(v, IntOrString):
    return v

  v = cast(IntOrStr, v)
  return v.value
```

flaviuvadan: Ah, yes, you are right... I think, though, that we should enrich this type so that users can pass in 1 of 3 types:
```suggestion
    limit: Optional[Union[int, str, IntOrString]] = None
```

Also, you pointed out the `__root__` usage issue. If we change line 55 from 
```python
return IntOrString(__root__=str(v)) # int or str
```
to
```python
return str(v)  # int or str
```

everything should work.

flaviuvadan (commented):
I left a comment on an implementation that I think will solve the problem! Also, any chance you can please add an upstream example in `examples/workflows/upstream/retry_script.py`? This will map to the following example: https://github.com/argoproj/argo-workflows/blob/master/examples/retry-script.yaml

Note that you will have to run `make regenerate-test-data` to generate the YAML for comparison/testing! 

```
from hera.workflows import Workflow, script, RetryStrategy


@script(image="python:alpine3.6", retry_strategy=RetryStrategy(limit=10), add_cwd_to_sys_path=False)
def retry_script():
    import random
    import sys
    exit_code = random.choice([0, 1, 1])
    sys.exit(exit_code)


with Workflow(generate_name="retry-script-", entrypoint="retry-script") as w:
    retry_script()

```

src/hera/workflows/retry_strategy.py
lakhan-twilio: yeah its working that way, if we just return **stringified** version of value but what if the value passed is `bool` then it will just convert that to string and no err will be raised ?

lakhanjindam: yeah this will work. But what if the value passed is `bool` for var `limit`, in that case it wont raise any err since it will be converting it to `str` type

src/hera/workflows/retry_strategy.py
```diff
@@ -65,3 +65,4 @@ def build(self) -> _ModelRetryStrategy:
 
 
 __all__ = ["RetryPolicy", "RetryStrategy"]
+r = RetryStrategy(limit=3)
```
samj1912, line 68: Can we please remove this line?

samj1912, line 68:
```suggestion
```

flaviuvadan (approved):
Thank you for the fix @lakhanjindam!

Might need a quick `make format` before we can merge

flaviuvadan:
@lakhanjindam the PR looks great to me so I am merging it into an upstream branch that I will run make format for you on then merge that one! Thank you again 🚀

Issue #550, opened 2023-04-04
Parameter Validation Error when using `get_parameter` with a parameter set using `ValueFrom`


inari-ldenton:
## Overview
Currently if you attempt to generate a workflow with the following snippet you will get a validation error, this is occuring on hera version 5.1.0.  I believe this is occurring due to the way the [get_parameter function returns a Parameter](https://github.com/argoproj-labs/hera/blob/593c0bf0641f4e01b1f8215252659457594bc1ef/src/hera/workflows/task.py#L184-L191).  It will copy over the existing `value_from` field and set the `value` field for the task output.  This leads to both `value_from` and `value` being set on the new parameter causing the validation to fail.  In my testing, removing the `value_from=obj.value_from,` from the parameter being returned fixes the issue. 

## Example Code
```
from hera.workflows import DAG, Parameter, Task, Workflow, script
from hera.workflows.models import ValueFrom


@script(
    outputs=[
        Parameter(
            name="a",
            value_from=ValueFrom(path="/test_out"),
        )
    ]
)
def out():
    with open("/test_out", "w") as f_out:
        f_out.write("TESTEST")


@script()
def in_(a):
    print(a)


with Workflow(generate_name="script-param-passing-", entrypoint="d") as w:
    with DAG(name="d"):
        t1: Task = out()
        t2 = in_(arguments=t1.get_parameter("a"))
        t1 >> t2

# Write the workflow
with open("param_test.yaml", "w") as f_out:
    f_out.write(w.to_yaml())

```

## Error
```
Traceback (most recent call last):
  File "**/param_passing.py", line 27, in <module>
    t2 = in_(arguments=t1.get_parameter("a"))
  File "**/lib/python3.10/site-packages/hera/workflows/task.py", line 185, in get_parameter
    return Parameter(
  File "pydantic/main.py", line 341, in pydantic.main.BaseModel.__init__
pydantic.error_wrappers.ValidationError: 1 validation error for Parameter
__root__
  Cannot specify both `value` and `value_from` when instantiating `Parameter` (type=value_error)
```

PR #565, opened 2023-04-08
Fix output param getter

flaviuvadan:
Fixes: #550

### DIFF
--- a/src/hera/workflows/task.py
+++ b/src/hera/workflows/task.py
@@ -146,7 +146,7 @@ class Task(
     def result(self) -> str:
         return f"{{{{tasks.{self.name}.outputs.result}}}}"
 
-    def get_parameters_as(self, name):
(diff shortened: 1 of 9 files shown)


src/hera/workflows/task.py
```diff
@@ -181,11 +181,9 @@ def get_parameter(self, name: str) -> Parameter:
 
         obj = next((output for output in parameters if output.name == name), None)
         if obj is not None:
-            obj.value = f"{{{{tasks.{self.name}.outputs.parameters.{name}}}}}"
             return Parameter(
                 name=obj.name,
-                value=obj.value,
-                value_from=obj.value_from,
+                value=f"{{{{tasks.{self.name}.outputs.parameters.{name}}}}}",
                 global_name=obj.global_name,
```
samj1912, line 187: Do we need these other things like global name and description for arguments?

samj1912 (approved):

PR #574, opened 2023-04-13
Fix inline script parsing issues

### LINKED ISSUE #572: Script inline constructor is flakey
Tried generating a script template with something like this:
```python
@script()
def echo(message: str) -> None:  # pragma: no cover
    """Prints out a message."""
    print(message)
```
and got something like this:
```yaml
source: "import os\nimport sys\nsys.path.append(os.getcwd())\nimport json\n\
        try: message = json.loads(r'''{{inputs.parameters.message}}''')\nexcept: message\
        \ = r'''{{inputs.parameters.message}}'''\n\ndef echo(message: str) -> None:\
        \  # pragma: no cover\n    \"\"\"Prints out a message.\"\"\"\n    # TODO Remove\
        \ function or replace with your logic\n    print(message)\n"
```

This seems to be because of the reliance on the function signature ending with a `:\n` string, from the code [here](https://github.com/argoproj-labs/hera/blob/main/src/hera/workflows/script.py#L334).

samj1912:
Fixes #572 

The fix makes the function parser a bit more robust by round-tripping through ast parse/unparse.

This ensures we get consistently formatted code each time and removes any extraneous comments which were previously causing parsing issues.

### DIFF
--- a/docs/examples/workflows/coinflip.md
+++ b/docs/examples/workflows/coinflip.md
@@ -72,12 +72,9 @@
 
             import random
 
+            result = ''heads'' if random.randint(0, 1) == 0 else ''tails''
 
-            result = "heads" if random.randint(0, 1) == 0 else "tails"
-
-            print(result)
-
-            '
+            print(result)'
       - name: heads
         script:
           command:
@@ -89,9 +86,7 @@
 
             sys.path.append(os.getcwd())
 
-            print("it was heads")
-
-            '
+            print(''it was heads'')'
       - name: tails
         script:
           command:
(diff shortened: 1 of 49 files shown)


docs/examples/workflows/coinflip.md
```diff
@@ -72,12 +72,9 @@
 
             import random
 
+            result = ''heads'' if random.randint(0, 1) == 0 else ''tails''
```
elliotgunton, line 75: The double single quotes makes me nervous, does this run in python?

samj1912: It's yaml first, if you convert it to json, the double single quotes go away.

elliotgunton (approved):
LGTM

flaviuvadan (approved):
🚀

PR #578, opened 2023-04-17
Fix conditional on task status to string conversion

janowskijak:
The conversion of TaskResult Enum to string creates an incorrect depends field e.g. :  
`depends: fail-or-succeed.TaskResult.Failed` instead of  `depends: fail-or-succeed.Failed`. 
See here this link for more info on [depends](https://argoproj.github.io/argo-workflows/enhanced-depends-logic/).  

This PR fixes the issue by converting to string the value of the Enum instead of the Enum itself.

### DIFF
--- a/docs/examples/workflows/dag_conditional_on_task_status.md
+++ b/docs/examples/workflows/dag_conditional_on_task_status.md
@@ -1,0 +1,93 @@
+# Dag Conditional On Task Status
+
+
+
+
+
+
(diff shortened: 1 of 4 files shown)


janowskijak:
Signed commits

janowskijak:
@flaviuvadan anything else I have to do here? I see that some checks are still pending.

samj1912 (approved):

PR #587, opened 2023-04-19
fix: broken URLs on Windows due to os.path.join()

abschm:
When trying to submit a workflow with Hera v5 on Windows, the use of os.path.join breaks the URLs because it joins with backslashes on Windows. Using urllib.parse.urljoin instead makes it compatible with both Linux and Windows.

The error when trying to create a workflow on Windows is "json.decoder.JSONDecodeError: Expecting value: line 1 
column 1 (char 0)" because Hera assumes a JSON response, but gets HTML because of the bad URL.

No tests because you test on Linux.

### DIFF
--- a/docs/examples/workflows/coinflip.md
+++ b/docs/examples/workflows/coinflip.md
@@ -72,12 +72,9 @@
 
             import random
 
+            result = ''heads'' if random.randint(0, 1) == 0 else ''tails''
 
-            result = "heads" if random.randint(0, 1) == 0 else "tails"
-
-            print(result)
-
-            '
+            print(result)'
       - name: heads
         script:
           command:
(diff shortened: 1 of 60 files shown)


tachylatus:
For the CI to succeed, you need to make the following additional changes (e.g. pipe the diff to `git apply`):

```diff
diff --git a/scripts/service.py b/scripts/service.py
index 59331c5..a480f60 100644
--- a/scripts/service.py
+++ b/scripts/service.py
@@ -400,8 +400,8 @@ def get_endpoints(
 def get_service_def() -> str:
     """Assembles the service definition string/code representation"""
     return """
+from urllib.parse import urljoin
 import requests
-import os
 from hera.{module}.models import {imports}
 from hera.shared import global_config
 from typing import Optional, cast
```

You can also pull it from my fork: https://github.com/tachylatus/hera-workflows/tree/fix_broken_urls_on_windows

samj1912 (approved):

PR #592, opened 2023-04-24
Fix script decorator to properly wrap the underlying function

samj1912:
Currently users are unable to use the underlying function outside of a hera context since we don't pass all of the args. This generates errors that reference the task wrapper since we don't use functool wraps. 

This fixes the script decorator so that - 
- Args are passed as intended
- The wrapper is wrapped with functools.wraps to ensure proper stack traces. See https://docs.python.org/3/library/functools.html#functools.wraps

### DIFF
--- a/src/hera/workflows/script.py
+++ b/src/hera/workflows/script.py
@@ -8,6 +8,7 @@ import copy
 import inspect
 import textwrap
 from abc import abstractmethod
+from functools import wraps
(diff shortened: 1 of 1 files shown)


elliotgunton:
Does this change usage (in the examples) at all? Or just lets you use args then kwargs when calling?

samj1912:
@elliotgunton it doesn't. It just allows users to use the original function as is.

Issue #567, opened 2023-04-09
Add more upstream Argo Workflows examples


flaviuvadan:

azro352:
What is the difference between "hera" and "upstream" examples ?

flaviuvadan:
Upstream examples are [examples from the Argo Workflows repository](https://github.com/argoproj/argo-workflows/tree/master/examples)! The [test bed of Hera](https://github.com/argoproj-labs/hera/blob/main/tests/test_workflow.py) generates all the YAMLs from the examples in the Hera repo and compares them with the upstream YAMLs. This ensures that the output of Hera matches the upstream YAMLs, which is an approach of guaranteeing/checking 1:1 feature parity. Hera examples are more "lenient", so to say, and meant to illustrate features of Hera

elliotgunton:
Current count is 90 missing upstream examples

<details>
<summary>Table of examples</summary>

| Name | Link |
|------|------|
| k8s-patch-basic | https://github.com/argoproj/argo-workflows/blob/master/examples/k8s-patch-basic.yaml |
| resubmit | https://github.com/argoproj/argo-workflows/blob/master/examples/resubmit.yaml |
| synchronization-mutex-wf-level | https://github.com/argoproj/argo-workflows/blob/master/examples/synchronization-mutex-wf-level.yaml |
| pod-metadata-wf-field | https://github.com/argoproj/argo-workflows/blob/master/examples/pod-metadata-wf-field.yaml |
| secrets | https://github.com/argoproj/argo-workflows/blob/master/examples/secrets.yaml |
| pod-spec-patch | https://github.com/argoproj/argo-workflows/blob/master/examples/pod-spec-patch.yaml |
| exit-code-output-variable | https://github.com/argoproj/argo-workflows/blob/master/examples/exit-code-output-variable.yaml |
| parameter-aggregation-script | https://github.com/argoproj/argo-workflows/blob/master/examples/parameter-aggregation-script.yaml |
| loops-sequence | https://github.com/argoproj/argo-workflows/blob/master/examples/loops-sequence.yaml |
| conditionals | https://github.com/argoproj/argo-workflows/blob/master/examples/conditionals.yaml |
| conditional-parameters | https://github.com/argoproj/argo-workflows/blob/master/examples/conditional-parameters.yaml |
| k8s-set-owner-reference | https://github.com/argoproj/argo-workflows/blob/master/examples/k8s-set-owner-reference.yaml |
[...]

PR #597, opened 2023-04-26
Add missing examples test

elliotgunton:
For tracking #567 

* Local `make regenerate-test-data` will notify user after running the normal tests
* CI/CD will not fail as the test is marked as `xfail`, and the `ci` make target doesn't have the `regenerate-test-data` as a dependency
* We can later remove the `xfail` so that CI/CD will check for missing examples as part of the regular `test` target when we've reached good coverage (we can change the assertion to e.g. `len(missing) < 10`)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -88,8 +88,9 @@ init-files:
 	$(MAKE) format
 
 .PHONY: examples
-examples:  ## Generate all the examples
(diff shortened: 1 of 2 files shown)


flaviuvadan (approved):
🚀

Issue #595, opened 2023-04-26
Better get_parameter/get_artifact support 


elliotgunton:
A goal of Hera should be to avoid the use of Argo expression syntax for Parameters and Artifacts, as without seeing the yaml itself it's not obvious what the keys are. E.g in the [script_auto_infer example](https://github.com/argoproj-labs/hera/blob/main/examples/workflows/script_auto_infer.py), a good syntax for the artifact would be:

```py
    with DAG(name="d"):
        p = produce()
        c = consume(arguments=Artifact(name="i", from_=p.get_artifact("result")))
        p >> c
```

---

Currently, there's no Hera syntax for fetching a Workflow argument, example taking inspiration from https://argoproj.github.io/argo-workflows/workflow-inputs/
```py
@script(constructor="inline")
def echo(message: str) -> None:
    print(message)

with WorkflowTemplate(
    name="my-wt-with-args",
    entrypoint="main",
    arguments=[
        Parameter(name="arg-1"),
        Parameter(name="arg-2"),
    ],
) as wt:
    Steps(name="main"):
        echo(
            name="test",
            arguments={
                "message": "{{workflow.parameters.arg-1}}",
            },
        )
```

Something like `wt.get_argument("arg-1")` would be better.

---

Addressing this issue would also be a good time to consolidate the `get_parameter[s_as]` functions that are duplicated (~except for "steps." vs "tasks." in `obj.value` assignment~ _this was a bug addressed in #624_) in [Step](https://github.com/argoproj-labs/hera/blob/69091e1ed77723746bc57db439a3a0f78f420990/src/hera/workflows/steps.py#L70) and [Task](https://github.com/argoproj-labs/hera/blob/69091e1ed77723746bc57db439a3a0f78f420990/src/hera/workflows/task.py#L153)

---

We've also had feedback that knowing whether to use `value` or `value_from` can be confusing (e.g. you could do `value="{{inputs.parameters.my-value}}` or (I think) `value_from=ValueFrom(expression="inputs.parameters.my-value")` - maybe we can figure out a way to assign to one of these automatically based on the input? Both of these are still using Argo expressions though.

---

Tasks:
- [x] Get artifacts in native Hera - [fixed](https://github.com/argoproj-labs/hera/pull/613/files#diff-2a0ef4257136d9b2d5dd30da1111e478658c6cb55a0f3a8ff7585ee2043b28daR111) in #613 
[...]

PR #598, opened 2023-04-26
Add script_artifact_passing example

### LINKED ISSUE #591: Cannot create artifacts in scripts without using "outputs=[Artifact(...)]" in @script decorator
I've stumbled upon something that might be a bug or might be an intended change from Hera 4, but that is confusing to me.

In Hera 4, I was used to specifying artifacts inside `with Workflow as w:` with `task_name = Task("name", func, inputs=[...], outputs=[Artifact(...)],)`. 
When trying something similar in Hera 5 with `task_name = func(arguments=[...], outputs=[Artifact(...)],)`, no artifact is created and no error is printed. 
However, using `@script(outputs=[Artifact(...)],)` works as expected.

So this code works in Hera 5 and the artifact called "decorator_created_file" is created:
```
@script(
    outputs=[
        Artifact(
            name="decorator_created_file",
            path="/home/app/decorator_file.txt",
        )
    ],
)
def make_file_with_decorator():
    with open("decorator_file.txt", "w") as f:
        f.write("hello world!")


with Workflow(
    generate_name="script-artifact-output",
    entrypoint="dag",
) as w:
    with DAG(name="dag"):
        make_file_with_decorator_task = make_file_with_decorator()

    make_file_with_decorator_task
w.create()
```

But this code doesn't work in Hera 5 because the artifact called "no_decorator_created_file" is not created:
```
@script()
def make_file_without_decorator():
    with open("no_decorator_file.txt", "w") as f:
        f.write("hello world!")


with Workflow(
    generate_name="script-artifact-output",
    entrypoint="dag",
) as w:
    with DAG(name="dag"):
        make_file_without_decorator_task = make_file_without_decorator(
            outputs=[
                Artifact(
                    name="no_decorator_created_file",
                    path="/home/app/no_decorator_file.txt",
                )
            ],
        )

    make_file_without_decorator_task
w.create()
```

I tried tracking down the difference on my own but didn't find it.
Also, it would be really helpful to have an official example of script artifact passing.

elliotgunton:
Fixes #591 with a script example to mirror `artifact_passing.py` upstream example.

Will also be useful as an example of artifact passing noted in #595

### DIFF
--- a/examples/workflows/script_artifact_passing.py
+++ b/examples/workflows/script_artifact_passing.py
@@ -1,0 +1,23 @@
+from hera.workflows import Artifact, Steps, Workflow, script
+
+
+@script(outputs=Artifact(name="hello-art", path="/tmp/hello_world.txt"))
+def whalesay():
+    with open("/tmp/hello_world.txt", "wb") as f:
+        f.write("hello world")
+
+
+@script(inputs=Artifact(name="message", path="/tmp/message"))
+def print_message():
+    with open("//tmp/message", "rb") as f:
+        message = f.readline()
+    print(message)
+
+
+with Workflow(generate_name="artifact-passing-", entrypoint="artifact-example") as w:
+    with Steps(name="artifact-example") as s:
+        whalesay(name="generate-artifact")
+        print_message(
+            name="consume-artifact",
+            arguments=[Artifact(name="message", from_="{{steps.generate-artifact.outputs.artifacts.hello-art}}")],
+        )
--- a/docs/examples/workflows/script_artifact_passing.md
+++ b/docs/examples/workflows/script_artifact_passing.md
@@ -1,0 +1,79 @@
+# Script Artifact Passing
+
+
+
+
+
+
+=== "Hera"
+
+    ```python linenums="1"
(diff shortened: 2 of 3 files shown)


examples/workflows/script_artifact_passing.py
```diff
@@ -0,0 +1,23 @@
...
+        message = f.readline()
+    print(message)
+
+
+with Workflow(generate_name="artifact-passing-", entrypoint="artifact-example") as w:
+    with Steps(name="artifact-example") as s:
+        whalesay(name="generate-artifact")
+        print_message(
+            name="consume-artifact",
+            arguments=[Artifact(name="message", from_="{{steps.generate-artifact.outputs.artifacts.hello-art}}")],
```
flaviuvadan, line 22: Is this PR a nice opportunity to implement `get_artifact` on step/task? 🤔 The same way there is a `get_parameter`

elliotgunton: I want to make a new PR where I'll do all the tasks in #595, and then the diff will show the changes from argo expressions to native hera 😄

PR #606, opened 2023-05-01
Add map reduce example

### LINKED ISSUE #588: Add upstream map-reduce example to Hera examples
https://github.com/argoproj/argo-workflows/blob/master/examples/map-reduce.yaml

Discussion w/ Anders, Helge, and Mohsin @ KubeCon

flaviuvadan:
**Pull Request Checklist**
- [x] Fixes #588
- [ ] Tests added

**Description of PR**

While working on #588 I noticed that the extra `{{item}}` based parameter was not added to a `Task.arguments` field. This happened because Hera performs a check on `source` + `with_param` when inferring the new parameters to set. So, added the `source` to the `kwargs` the same way the `name` is added. This PR also enriches the used arguments in the parameter inference to use the `kwargs` ones too, since that is where the incoming arguments are set. 

Added the upstream mapreduce example to the regular examples folder rather than upstream. The upstream example is not formatted properly.

~### Question~

~- @samj1912 or @elliotgunton - I am having a hard time determining why example comparison fails on this branch. It _seems_ like [this line](https://github.com/argoproj-labs/hera-workflows/blob/266f65eeaa11fbe6861f33bc93e56e7175a7e454/tests/test_workflow.py#L45) does not yield reproducible results between upstream and the Hera generated YAML?~


examples/workflows/map_reduce.py
```diff
@@ -0,0 +1,75 @@
+"""
+This is a map reduce example from the upstream Argo Workflows repository. This is not part of the upstream examples
+folder for workflows because the upstream example is not formatted properly
```
elliotgunton, line 3: Adds the markdown link to render in the docs

```suggestion
folder for workflows because the upstream example is not formatted properly.

See the upstream example [here](https://github.com/argoproj/argo-workflows/blob/master/examples/map-reduce.yaml)
```

elliotgunton: Added in #649

src/hera/workflows/_unparse.py
```diff
@@ -944,6 +946,10 @@ def unparse(ast_obj):
 
 def roundtrip(source):
     tree = ast.parse(source)
-    if hasattr(ast, "unparse"):
+    # python 3.11 ast.unparse stopped using brackets around tuples used in assignment,
+    # see https://github.com/python/cpython/commit/51cef8be8c77dff522bec6f95d6853031bf19038
+    # Previous versions will unparse with brackets
+    if sys.version_info[1] >= 11 and hasattr(ast, "unparse"):
```
samj1912, line 952: Not the current implementation of unparse was taken from python 3.8 compat syntax. Python 3.9+ have added other experssions to the grammar and the parser/unparser may not work as expected. I would prefer if we drop support for 3.8 entirely if possible rather than produce incorrect parse trees.

samj1912: We already see some issues here for eg https://github.com/argoproj-labs/hera/issues/584

samj1912, line 952:
```suggestion
    if hasattr(ast, "unparse"):
```

elliotgunton:
I don't think this PR is needed anymore, not sure how the map reduce example got into main 🤔😂

samj1912:
@elliotgunton did the map reduce example get lost in the merge? The diff doesn't seem to be accurate?

elliotgunton:
@samj1912 https://github.com/argoproj-labs/hera/blob/main/examples/workflows/map_reduce.py 🤷‍♂️

PR #626, opened 2023-05-11
Adding Intuition Machines to Users.md

tmi:
Just adding another Hera user.


flaviuvadan (approved):
Thank you @tmi! :rocket:

PR #635, opened 2023-05-25
Fix types in script decorator

elliotgunton:
**Pull Request Checklist**
- [x] Tests added - manual testing of typing system with screenshots

**Description of PR**

Currently, using a function decorated with a `@script` decorator loses type information when in a steps context or when used outside of a workflow e.g. in tests.

This PR fixes the type information in the script decorator.

Tested with functions such as:

```py
@script()
def hello_world():  # no args
    print("Hello world")

@script()
def hello(message: str): # one str arg
    print(f"Hello {message}")

@script()
def echo_a_dict(my_dict: Dict[str, str]): # dict arg
    for k, v in my_dict.items():
        print(f"{k}={v}")

@script()
def echo_multiple(message: str, times: int):  # multiple args, different types
    print(message * times)

@script()
def return_str(message: str) -> str:  # has a return type
    return f"Your message: '{message}'"
```

See screenshots:

Under a steps context we can see we'll get back a Task or a Step. Currently not able to infer types of the Script kwargs that are passed through the decorator. The closest we can get I think is to check for missing values/correctness described [here](https://stackoverflow.com/questions/68995170/pydantic-get-a-fields-type-hint) - we need the types of the `Script` class members which I think are only available at runtime.

under-steps

Outside of a steps we can get the full type info of the underlying function (note that the chosen hint isn't automatic)

out-of-steps

Func with no args

no-args

Func with multiple args, different types

echo-multiple

Func with return type 

return-type

Hover over at any location

hovering-over

Automatic intellisense within a Steps context (!!!) (Note it shows `depends` which is from `Task` only, because `hello_step` has a type of `Union[Step, Task, None]`, so it seems like a best effort by pylance)

intellisense

### DIFF
--- a/src/hera/workflows/script.py
+++ b/src/hera/workflows/script.py
@@ -243,21 +243,39 @@ FuncIns = ParamSpec("FuncIns")  # For input types of given func to script decora
 FuncR = TypeVar("FuncR")  # For return type of given func to script decorator
 ScriptIns = ParamSpec("ScriptIns")  # For attribute types of Script
 
-# R = Union[Callable[FuncIns, FuncR], Callable[ScriptIns, Union[Task, Step]]]
-# def _take_annotation_from(_: Callable[ScriptIns, R]) -> Callable[[Callable], Callable[ScriptIns, R]]:
-#     def decorator(real_function: Callable) -> Callable[ScriptIns, R]:
-#         def new_function(*args: ScriptIns.args, **kwargs: ScriptIns.kwargs) -> R:
-#             return real_function(*args, **kwargs)
 
-#         return new_function
+def _take_annotation_from(
+    _: Callable[
+        ScriptIns,
+        Callable[[Callable[FuncIns, FuncR]], Union[Callable[FuncIns, FuncR], Callable[ScriptIns, Union[Task, Step]]]],
+    ]
+) -> Callable[
+    [Callable],
+    Callable[
+        ScriptIns,
+        Callable[[Callable[FuncIns, FuncR]], Union[Callable[FuncIns, FuncR], Callable[ScriptIns, Union[Task, Step]]]],
+    ],
+]:
+    def decorator(
+        real_function: Callable,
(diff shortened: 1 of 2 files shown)


flaviuvadan (approved):
LGTM! Curious what @samj1912 thinks as well

elliotgunton:
We're almost there! Figuring out how to correctly expand the type of the ScriptIn ParamSpec 
image

Rest of the screenshots are the same except for `...` replaced with `**ScriptIns@script`, and I seem to have broken automatic intellisense within the steps context.

WIP: I can get type hinting within the decorator 💯 
image

**BUT** it breaks the type hinting of the function call site which gets stuck as the function itself (it doesn't know it will return `Step | Task`)
image

elliotgunton:
It works!! `script` decorator intellisense:

image

And at the call site! All the Script kwargs:

image

So we can find `arguments` in there
image

Although this results in `No value for argument 'name' in function call pylint` which I would say we can safely ignore at this point

src/hera/workflows/script.py
```diff
@@ -228,6 +239,42 @@ def _get_parameters_from_callable(source: Callable) -> Optional[List[Parameter]]
...
 
+FuncIns = ParamSpec("FuncIns")  # For input types of given func to script decorator
+FuncR = TypeVar("FuncR")  # For return type of given func to script decorator
+ScriptIns = ParamSpec("ScriptIns")  # For attribute types of Script
+
+
+def _take_annotation_from(
+    _: Callable[
+        ScriptIns,
+        Callable[[Callable[FuncIns, FuncR]], Union[Callable[FuncIns, FuncR], Callable[ScriptIns, Union[Task, Step]]]],
```
elliotgunton, line 250: Will see if I can type alias this but it seemed to be losing the underlying type info for intellisense suggetsions. Tomorrow's problem 😴

samj1912 (approved):

PR #638, opened 2023-05-26
[#614] Add a Hera exceptions module

### LINKED ISSUE #614: Add `hera.exceptions` module and intercept Argo server errors to transform into Hera exceptions
Context: https://cloud-native.slack.com/archives/C03NRMD9KPY/p1683213144163779
CC: @zach-jablons-hinge

flaviuvadan:
**Pull Request Checklist**
- [x] Fixes #614
- [x] Tests added

Currently, Hera simply raises a `ValueError` for any type of error that is intercepted from the Argo server. This PR adds an exceptions shared module to Hera. This will help users have a more pleasant experience intercepting any exceptions that are raised by Hera.


src/hera/workflows/service.py
```diff
@@ -97,8 +99,17 @@ def list_archived_workflows(
 
         if resp.ok:
             return WorkflowList(**resp.json())
-        else:
-            raise Exception(f"Server returned status code {resp.status_code} with error: {resp.json()}")
+
+        try:
```
samj1912, line 103: @flaviuvadan we seem to be repeating these lines everywhere. Should we move these under the function call as well?

elliotgunton (commented):
I agree with Sambhav's suggestion. We can then simplify the exception function (plus I would suggest a rename) to:

```py
def exception_from_server_response(resp: Response) -> HeraException:
    """Returns a `HeraException` mapped from the given status code initialized with the given message"""
    try:
        return status_code_to_exception_map.get(resp.status_code, HeraException)(
            f"Server returned status code {resp.status_code} with message: `{resp.json()['message']}`"
        )
    except json.JSONDecodeError:
        return status_code_to_exception_map.get(resp.status_code, HeraException)(
            f"Server returned status code {resp.status_code} with message: `{resp.text}`"
        )
```

and then the try /raise / excepts become
```py
        if resp.ok:
            return ...

        raise exception_from_server_response(resp)
```

src/hera/workflows/service.py
flaviuvadan: I think what you and Elliot suggested makes perfect sense. My aim with the independent lines was to have the raise inside the service function so that the backtrace does not go outside the service. However, if we make the helper func take a Response as @elliotgunton suggested, then we get the best of both worlds:
- isolated func + code
- users going through a backtrace still get to only focus on the service function because that's the spot that raises

scripts/service.py
```diff
@@ -430,8 +430,10 @@ def get_service_def() -> str:
     return """
 from urllib.parse import urljoin
 import requests
+import json
```
flaviuvadan, line 433: remove this

elliotgunton (approved):
LGreatTM! 🤘

PR #613, opened 2023-05-04
Add support for templatable exit handlers and V4 examples

flaviuvadan:
**Pull Request Checklist**
- [x] Fixes #610
- [ ] Tests added
- [ ] Documentation/examples added
- [ ] [Good commit messages](https://cbea.ms/git-commit/) and/or PR title

**Description of PR**

Currently, the Hera v4 examples are mostly missing. This PR adds numerous examples from V4 and also fixes a few tiny bugs that were identified:
- add source to callable kwargs for otherwise the correct parameters are not generated
- add support for templates to the on_exit field of workflows

PR #639, opened 2023-05-26
[#595] Improve artifact/parameter access

### LINKED ISSUE #595: Better get_parameter/get_artifact support

flaviuvadan:
**Pull Request Checklist**
- [x] Fixes #595 
- [ ] Tests added

**Description of PR**

@elliotgunton and I bounced some thoughts on improving access to parameters/artifact via #595 and #613. This PR adds the implementation @elliotgunton suggested to the template invokator class. Open to having these on a different mixin!


src/hera/workflows/_mixins.py
```diff
@@ -675,6 +676,124 @@ def one(xs: List):
...
+            When something else other than an `Parameter` is found for the specified name.
+        """
+        if isinstance(self.template, str):
+            raise ValueError(f"Cannot get output parameters when the template was set via a name: {self.template}")
+
+        # here, we build the template early to verify that we can get the outputs
+        if isinstance(self.template, Templatable):
+            template = self.template._build_template()
+        else:
+            template = self.template
```
elliotgunton, line 731: A general point here - we've been finding it hard to work with TemplateRefs that we know output certain parameters, so we might want to relax/change this validation. Something we'll look to address in a later PR, JFYI

flaviuvadan: I see. Do you imagine this can actually accommodate template refs by actually fetching the template based on the ref if self.template is a string, and search the params like that?

elliotgunton: >fetching the template

This would be pretty awesome, and along the lines of what we're thinking - if we're able to statically generate stubs from a given WorkflowTemplate, so that we have a class that inherits from a Hera `TemplateRef` and works with our functions like `get_parameter`.

```py
class TemplateRef(IOMixin):
    pass  # let's say we've implemented this


from hera.workflow import TemplateRef


class MyTemplateRefStub(TemplateRef):  # we generate this in some hera.workflow.template_ref_stubs module or something
    inputs = [Parameter("a-parameter", default="some-default")]
    outputs = [Parameter("an-output-parameter")]
    ... etc
```

elliotgunton (approved):
LGTM

PR #643, opened 2023-05-26
[#522] Add a wait + synchronous `Workflow.create`

### LINKED ISSUE #522: Add `wait` method to `Workflow`
**Background**

To help perform live tests of workflow templates, we would like to be able to write the following in test cases:

```python
wflo_template = WorkflowTemplate(...)
wflo_template.apply(...)
wflo = wflo_template.run_as_worklow(...)
wflo.wait()
# Check post-conditions on workflow
```

**Proposal**

Add a `wait` method to `Workflow` that returns only after the workflow has been completed, or a timeout has been reached.

flaviuvadan:
**Pull Request Checklist**
- [x] Fixes #522
- [ ] Tests added

**Description of PR**

See #522 for the feature request!

### DIFF
--- a/src/hera/workflows/workflow.py
+++ b/src/hera/workflows/workflow.py
@@ -5,6 +5,7 @@ for more on Workflows.
 """
 from __future__ import annotations
 
+import time
 from pathlib import Path
 from types import ModuleType
 from typing import Any, Dict, List, Optional, Union
@@ -44,11 +45,12 @@ from hera.workflows.models import (
     WorkflowLintRequest,
     WorkflowMetadata,
     WorkflowSpec as _ModelWorkflowSpec,
-    WorkflowStatus,
+    WorkflowStatus as _ModelWorkflowStatus,
     WorkflowTemplateRef,
 )
 from hera.workflows.protocol import Templatable, TTemplate, TWorkflow, VolumeClaimable
 from hera.workflows.service import WorkflowsService
+from hera.workflows.workflow_status import WorkflowStatus
 
 _yaml: Optional[ModuleType] = None
 try:
@@ -82,7 +84,7 @@ class Workflow(
     # Workflow fields - https://argoproj.github.io/argo-workflows/fields/#workflow
     api_version: Optional[str] = None
     kind: Optional[str] = None
-    status: Optional[WorkflowStatus] = None
+    status: Optional[_ModelWorkflowStatus] = None
 
     # ObjectMeta fields - https://argoproj.github.io/argo-workflows/fields/#objectmeta
     annotations: Optional[Dict[str, str]] = None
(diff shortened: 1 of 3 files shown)


src/hera/workflows/workflow.py
```diff
@@ -330,13 +332,44 @@ def to_yaml(self, *args, **kwargs) -> str:
...
         assert self.workflows_service, "workflow service not initialized"
         assert self.namespace, "workflow namespace not defined"
-        return self.workflows_service.create_workflow(
+        if async_:
+            return self.workflows_service.create_workflow(
+                WorkflowCreateRequest(workflow=self.build()), namespace=self.namespace
+            )
+        return self.wait(poll_interval=poll_interval)
+
+    def wait(self, poll_interval: int = 5) -> TWorkflow:
```
elliotgunton, line 355: What do we think about a timeout parameter as well? We could then also allow a cleanup callback hook if it times outs and e.g. send a `WorkflowTerminateRequest` to kill the workflow

flaviuvadan: Sounds like an amazing idea! What do you think about doing that in a follow up PR? We should also discuss how that will play with `active_deadline_seconds`: https://argoproj.github.io/argo-workflows/fields/#fields_3

elliotgunton: Sgtm! Won't be a breaking change so fine by me

elliotgunton (approved):

src/hera/workflows/workflow.py
```diff
@@ -330,13 +332,44 @@ def to_yaml(self, *args, **kwargs) -> str:
...
+        """Waits for the Workflow to complete execution.
+
+        Parameters
+        ----------
+        poll_interval: int = 5
+            The interval in seconds to poll the workflow status.
+        """
+        wf = self.workflows_service.create_workflow(
```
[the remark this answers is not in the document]
elliotgunton: Yep that's a bug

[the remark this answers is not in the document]
flaviuvadan: I aimed for a design in which Hera does not have to track the name of a workflow based on users' use of `generated_name=True`. Not performing a create inside the wait implies users have to call create first, which is ok, or call create with the wait flag (replaced async_ with wait). I made the adjustment and now there's a get inside the wait!

src/hera/workflows/workflow.py
```diff
@@ -330,13 +332,57 @@ def to_yaml(self, *args, **kwargs) -> str:
...
+        """
+        assert self.workflows_service is not None, "workflow service not initialized"
+        assert self.namespace is not None, "workflow namespace not defined"
+        assert self.name is not None, "workflow name not defined"
+
+        wf = self.workflows_service.get_workflow(self.name, namespace=self.namespace)
+        assert wf.status is not None, f"workflow status not defined for workflow {wf.metadata.name}"
+        assert wf.status.phase is not None, f"workflow phase not defined for workflow status {wf.status}"
+        status = WorkflowStatus.from_argo_status(wf.status.phase)
+        assert wf.metadata.name is not None, f"workflow name not defined for workflow {self.name}"
```
elliotgunton, line 376: Think it makes more sense to have the name assert first as the status assert uses it. Added a whitespace for readability
```suggestion
        wf = self.workflows_service.get_workflow(self.name, namespace=self.namespace)
        assert wf.metadata.name is not None, f"workflow name not defined for workflow {self.name}"

        assert wf.status is not None, f"workflow status not defined for workflow {wf.metadata.name}"
        assert wf.status.phase is not None, f"workflow phase not defined for workflow status {wf.status}"
        status = WorkflowStatus.from_argo_status(wf.status.phase)
```

samj1912 (approved):

PR #644, opened 2023-05-26
Move properties/funcs into TISubNodeMixin

elliotgunton:
Small PR on top of a WIP to avoid making suggestions/commits


elliotgunton:
Hold up, we can reduce duplication of the properties too, WIP

flaviuvadan (approved):
Amazing suggestion @elliotgunton !! 🎖️

PR #647, opened 2023-05-26
[#513] Add source args inference to callable template mixin to support source kwargs in regular Py call

### LINKED ISSUE #513: Make `CallableTemplate` accept the kwargs of the function that is called with a `Script`'s `source`
It would be amazing if Hera would support this syntax:
```python
@script(...)
foo(a, b, c):
    print(a, b, c)

with Workflow(...) as w:
    foo(a=1, b=2, c=3, name="testing")
```

For this to happen Hera needs to evaluate the `kwargs` passed to the `__call__` inside `CallableTemplate`, match them against a `source`, and create the necessary parameters.

flaviuvadan:
**Pull Request Checklist**
- [x] Fixes #513
- [x] Tests added

**Description of PR**

Currently, users are forced to use the `arguments` keyword to set up the parameters to be set for a Python source executed on Argo. 

This PR adds the possibility to set source callable kwargs as input to the callable, effectively removing the need for using `arguments` for simple function calls. This also accepts parameters/artifacts to be set the same way!

### DIFF
--- a/docs/examples/workflows/script_artifact_passing.md
+++ b/docs/examples/workflows/script_artifact_passing.md
@@ -48,7 +48,6 @@
               artifacts:
               - from: '{{steps.generate-artifact.outputs.artifacts.hello-art}}'
                 name: message
-                path: /tmp/hello_world.txt
             name: consume-artifact
             template: print-message
       - name: whalesay
--- a/docs/examples/workflows/script_artifact_passing.md
+++ b/docs/examples/workflows/script_artifact_passing.md
@@ -26,11 +26,8 @@
 
     with Workflow(generate_name="artifact-passing-", entrypoint="artifact-example") as w:
         with Steps(name="artifact-example") as s:
-            whalesay(name="generate-artifact")
-            print_message(
-                name="consume-artifact",
-                arguments=[Artifact(name="message", from_="{{steps.generate-artifact.outputs.artifacts.hello-art}}")],
-            )
+            ga = whalesay(name="generate-artifact")
+            print_message(name="consume-artifact", message=ga.get_artifact("hello-art"))
     ```
 
 === "YAML"
@@ -47,11 +44,7 @@
         steps:
         - - name: generate-artifact
             template: whalesay
-        - - arguments:
-              artifacts:
-              - from: '{{steps.generate-artifact.outputs.artifacts.hello-art}}'
-                name: message
-            name: consume-artifact
+        - - name: consume-artifact
             template: print-message
       - name: whalesay
         outputs:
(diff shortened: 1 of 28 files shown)


examples/workflows/script-artifact-passing.yaml
```diff
@@ -9,11 +9,7 @@ spec:
     steps:
     - - name: generate-artifact
         template: whalesay
-    - - arguments:
```
flaviuvadan, line 12: TODO: find out why this is broken

src/hera/workflows/_mixins.py
```diff
@@ -479,6 +478,22 @@ def __call__(self, *args, **kwargs) -> Union[Step, Task]:
         try:
             from hera.workflows.steps import Step
 
+            arguments = self._get_arguments(**kwargs)
```
elliotgunton, line 481: We might as well move the duped code to before the first `try`

examples/workflows/script-artifact-passing.yaml
elliotgunton: I figure out why it's broken. PR up at #648

Leaving this here as notes for my debugging process:

`_get_arguments` in CallableTemplateMixin isn't finding your assignment of `message`

E.g. we know this works:
```py
        print_message(name="consume-artifact", arguments=ga.get_artifact("hello-art").as_name("message"))
```

But when we try
```py
        print_message(name="consume-artifact", message=ga.get_artifact("hello-art"))
```

This is assigning the `message` kwarg to an artifact that looks like
```
Artifact(name='hello-art', from_='{{steps.generate-artifact.outputs.artifacts.hello-art}}', path='/tmp/hello_world.txt', ...)
```

So _get_arguments seeing the given artifact as it's hiding in the `message` kwarg, plus it should have the name "message" otherwise the step will fail. The fix is to look through the kwargs for anything that looks like a Parameter or Artifact, and rename it to the kwarg name. Bonus is that we can check for clashing kwargs too!

src/hera/workflows/_mixins.py
elliotgunton: Done this in #648

docs/examples/workflows/script_artifact_passing.md
```diff
@@ -51,6 +48,7 @@
               artifacts:
               - from: '{{steps.generate-artifact.outputs.artifacts.hello-art}}'
                 name: message
+                path: /tmp/hello_world.txt
```
flaviuvadan, line 51: Maybe only `name` and `from` should be copied over, thoughts here @elliotgunton ?

flaviuvadan: Adjusted, please see https://github.com/argoproj-labs/hera/pull/647/files#diff-079dae561586ca661f24348fd6560e265e3c6eba6650abd05491201598ae9288R537

elliotgunton: Sounds right to me 💯

elliotgunton (approved):
I love this change!! 💯 👑 🥇

PR #648, opened 2023-05-27
Get any Artifacts/Parameters contained in function kwargs

elliotgunton:
Adds kwarg scanning for Artifacts/Parameters that aren't the standard
`arguments` or `with_param` when calling a function decorated with `@script`

### DIFF
--- a/src/hera/workflows/_mixins.py
+++ b/src/hera/workflows/_mixins.py
@@ -77,6 +77,7 @@ OutputsT = Optional[
         List[Union[Parameter, ModelParameter, Artifact, ModelArtifact]],
     ]
 ]
+ArgumentT = Union[Parameter, ModelParameter, Artifact, ModelArtifact]
 ArgumentsT = Optional[
     Union[
         ModelArguments,
@@ -467,6 +468,9 @@ class ArgumentsMixin(BaseMixin):
 
 class CallableTemplateMixin(ArgumentsMixin):
     def __call__(self, *args, **kwargs) -> Union[Step, Task]:
+        from hera.workflows.steps import Step
+        from hera.workflows.task import Task
+
         if "name" not in kwargs:
             kwargs["name"] = self.name  # type: ignore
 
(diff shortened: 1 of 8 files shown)


src/hera/workflows/script.py
```diff
@@ -48,6 +49,9 @@
 from hera.workflows.steps import Step
 from hera.workflows.task import Task
 
+CLASHING_KWARGS = list(Step.__fields__.keys()) + list(Task.__fields__.keys())
+ARGUMENT_TYPE_KWARGS = ("arguments", "with_param")
```
flaviuvadan, line 53: Any chance we can add a small piece of documentation that explains why these 2 fields?

flaviuvadan: Suggesting a small set for deduplication and slightly faster search ```suggestion
CLASHING_KWARGS = set(list(Step.__fields__.keys()) + list(Task.__fields__.keys()))
ARGUMENT_TYPE_KWARGS = ("arguments", "with_param")
```

File: src/hera/workflows/_mixins.py [hunk 2]
```diff
@@ -467,62 +467,88 @@ def _build_arguments(self) -> Optional[ModelArguments]:
...
+            if isinstance(arg_value, (Parameter, ModelParameter, Artifact, ModelArtifact)):
+                if arg_name not in CLASHING_KWARGS:
+                    # replace names for each argo argument so that this  step/task receives them with correct names
+                    arg_value.name = arg_name
+                    arguments.append(arg_value)
+                elif arg_name not in ARGUMENT_TYPE_KWARGS:
+                    raise ValueError(
+                        f"'{arg_name}' clashes with Step/Task kwargs. Rename '{arg_name}' or "
+                        "pass your Parameter/Artifact through 'arguments' or 'with_param'"
+                    )
```
flaviuvadan, line 545: Small piece of feedback, ignore if you disagree. I generally suggest using truthy `is` vs falsy `is not` statements as they seem to be slightly easier to understand:
```suggestion
                if arg_name in CLASHING_KWARGS:
                    raise ValueError(
                        f"'{arg_name}' clashes with Step/Task kwargs. Rename '{arg_name}' or "
                        "pass your Parameter/Artifact through 'arguments' or 'with_param'"
                    )
                elif arg_name in ARGUMENT_TYPE_KWARGS:
                    # replace names for each argo argument so that this  step/task receives them with correct names
                    arg_value.name = arg_name
                    arguments.append(arg_value)
```

flaviuvadan: > arg_value.name = arg_name

Is this replacement safe to do? 🤔 I wonder if this name change affects references downstream of other consumer step/tasks

flaviuvadan (approved):
Approving as it's overall good and functional but left a few small questions and two suggestions!

flaviuvadan:
@elliotgunton maybe we should also enrich the `Parameter` documentation you wrote recently with a note about this? Basically almost copy/paste the comments from the code that explains the 3 use cases of parameter inference/clashing

src/hera/workflows/_mixins.py
elliotgunton: >Is this replacement safe to do? 

I don't believe so, no! Good catch!

elliotgunton: >slightly easier to understand

I would usually agree, but in this case the if statements in your comment aren't equivalent to mine.

The `if`s are set up in this way because `ARGUMENT_TYPE_KWARGS` is a subset of `CLASHING_KWARGS`.

i.e. this means if we're currently looking at a Parameter/Artifact value (from the `isinstance`):
- we first check the `not in CLASHING_KWARGS` condition as a means of saying "is this a func param?" in which case, full steam ahead.
- Then, if it _is_ `in CLASHING_KWARGS` so it _could_ be in `ARGUMENT_TYPE_KWARGS` which we need to allow (which your code would not). So then we're looking at the `elif arg_name not in ARGUMENT_TYPE_KWARGS`. i.e. we can say the arg is clashing, but we can still allow arguments and with_param usage.

An alternative is a single `if` checking if in clashing_kwargs and not in arg_type_kwargs like so:

```suggestion
                if arg_name in CLASHING_KWARGS and arg_name not in ARGUMENT_TYPE_KWARGS:
                    raise ValueError(
                        f"'{arg_name}' clashes with Step/Task kwargs. Rename '{arg_name}' or "
                        "pass your Parameter/Artifact through 'arguments' or 'with_param'"
                    )
                # replace names for each argo argument so that this  step/task receives them with correct names
                arg_value.name = arg_name
                arguments.append(arg_value)
```

I think this is better actually so will use it. Thanks for your suggestion to help rethink it!

flaviuvadan: Ah, got it, thank you for the discussion!

flaviuvadan (approved):

PR #661, opened 2023-05-31
Common arg inference code for steps/tasks

elliotgunton:
**Pull Request Checklist**
- [ ] Tests added

**Description of PR**
_Copied from [comment](https://github.com/argoproj-labs/hera/pull/660#issuecomment-1570430679)_
>>I think things like "{{item.name}}" are set automatically for Task arguments. Should we expose that to Step as well?

>@elliotgunton can confirm these are set automatically for tasks based on arguments inference. So, Task invocation does not need the arguments field when with_param is passed, or with_item, but Step does for now. We should add the same logic to Step as well: https://github.com/argoproj-labs/hera/blob/main/src/hera/workflows/_mixins.py#L469 Motivation here is to not confuse users with these differences of "sometimes I can set a thing on Task but not on Step"

### DIFF
--- a/examples/workflows/script_loops_maps.py
+++ b/examples/workflows/script_loops_maps.py
@@ -1,0 +1,21 @@
+from hera.workflows import Steps, Workflow, script
+
+
+@script()
+def test_key_mapping(key_1: str, key_2: str):  # pragma: no cover
+    print(f"{key_1}:{key_2}")
+
+
+with Workflow(
+    generate_name="loops-maps-",
+    entrypoint="loop-map-example",
+) as w:
+    with Steps(name="loop-map-example") as loop_map_example:
+        test_key_mapping(
+            with_items=[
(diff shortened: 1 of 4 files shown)


examples/workflows/script_loops_maps.py
```diff
@@ -0,0 +1,21 @@
+from hera.workflows import Steps, Workflow, script
+
+
+@script()
+def test_key_mapping(key_1: str, key_2: str):  # pragma: no cover
+    print(f"{key_1}:{key_2}")
```
flaviuvadan, line 6: Until we figure out the parse/unparse problem, we need to use `format`. Other examples already do this :( 

```suggestion
    print(f"{key_1}:{key_2}")
    print("{key_1}:{key_2}".format(key_1=key_1, key_2=key_2))
```

PR #642, opened 2023-05-26
[#520] Add template apply

flaviuvadan:
**Pull Request Checklist**
- [x] Fixes #520
- [ ] Tests added
- [x] Documentation/examples added
- [x] [Good commit messages](https://cbea.ms/git-commit/) and/or PR title

**Description of PR**

See #520 for the feature request!

PR #669, opened 2023-06-05
WorkflowTemplate/ClusterWorkflowTemplate create_as_workflow

### LINKED ISSUE #521: Add `run_as_workflow` method
**Background**

To help perform live tests of workflow templates, we would like to be able to write the following in test cases:

```python
wflo_template = WorkflowTemplate(...)
wflo_template.apply(...)
wflo = wflo_template.run_as_worklow(...)
wflo.wait()
# Check post-conditions on workflow
```

**Proposal**

Add a method `run_as_worklow` to `WorkflowTemplate` which takes the following parameters:
- `template`: the name of the template to run
- `inputs`: the list of inputs parameters

The method returns a `Workflow` object.

elliotgunton:
**Pull Request Checklist**
- [x] Fixes #521 
- [x] Tests added

**Description of PR**
Currently, the DevEx for a workflow template is either:
* you have to iterate on a Workflow until it's ready to submit as a WorkflowTemplate; or
* you submit a WorkflowTemplate to the cluster, and run it via `workflowTemplateRef`, `templateRef` or via CLI `submit --from`

This PR adds
* the `create_as_workflow` function (reflecting existing naming in the `Workflow` class rather than `run_as`)
  * this modifies the `kind` in place, then submits to Argo through `Workflow.create`
  * we pass through `wait` and `poll_interval` (was hesitant to just use `kwargs`)
* `ClusterWorkflowTemplate` as a subclass of `WorkflowTemplate` as it was missing, added to `TWorkflow` type
  * acts as a fairly light wrapper around `WorkflowTemplate`, as we need to call the right `WorkflowService` functions
  * `build` just converts the pydantic model to `ClusterWorkflowTemplate`.


src/hera/workflows/cluster_workflow_template.py
```diff
@@ -0,0 +1,70 @@
...
+
+    def create(self) -> TWorkflow:  # type: ignore
+        """Creates the ClusterWorkflowTemplate on the Argo cluster."""
+        assert self.workflows_service, "workflow service not initialized"
+        assert self.namespace, "workflow namespace not defined"
+        return self.workflows_service.create_cluster_workflow_template(
+            ClusterWorkflowTemplateCreateRequest(template=self.build())
+        )
+
+    def get(self) -> TWorkflow:
```
flaviuvadan, line 25: Should `get`s on classes such as workflows, workflow template, etc, be class methods? 🤔 Or static methods that return a new `Workflow` based on the details of the upstream object if one is found? The service _does_ have to be instantiated, so I understand if this implementation is a _necessary_ approach

src/hera/workflows/workflow_template.py
```diff
@@ -153,5 +163,21 @@ def build(self) -> TWorkflow:
...
+        Note: this does not require the WorkflowTemplate to exist on the cluster
+        """
+
+        workflow = self.build()
+        workflow.kind = "Workflow"
+        if workflow.metadata.name:
+            # In case of a name being > TRUNCATE_LENGTH, we need to truncate the value to assign to generate_name
+            # otherwise the random suffix will make the generated name too long
+            workflow.metadata.generate_name = workflow.metadata.name[:TRUNCATE_LENGTH] + "-"
+            workflow.metadata.name = None
```
flaviuvadan, line 178: This is an implicit modification of a user submitted workflow that Hera will perform - what's a good way to "advertise" this is happening? Perhaps a log? 🤔 Also, should this logic be isolated and shared in a validator, perhaps? Maybe wf template + cluster wf template might also benefit from inheriting this. Thoughts?

src/hera/workflows/cluster_workflow_template.py
```diff
@@ -0,0 +1,70 @@
...
+            ClusterWorkflowTemplateLintRequest(template=self.build())
+        )
+
+    def build(self) -> TWorkflow:
+        """Builds the ClusterWorkflowTemplate and its components into an Argo schema ClusterWorkflowTemplate object."""
+        # Note that ClusterWorkflowTemplates are exactly the same as WorkflowTemplates except for the kind which is
+        # handled in Workflow._set_kind (by __name__). When using ClusterWorkflowTemplates via templateRef, clients
+        # should specify cluster_scope=True, but that is an intrinsic property of ClusterWorkflowTemplates from our
+        # perspective.
+        return _ModelClusterWorkflowTemplate(**super().build().dict())
```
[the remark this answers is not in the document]
elliotgunton: Good catch, it has exactly the same `metadata: v1_1.ObjectMeta` member, so it will let you set the namespace, but I guess Argo will ignore it on submission. Would still be part of the metadata though so maybe we should validate/remove it?

src/hera/workflows/cluster_workflow_template.py
elliotgunton: Hmm yeah maybe this should be a private method (as it's needed for `update`), and a `get` could be a static method to get via a `name` (+any other info needed)

src/hera/workflows/cluster_workflow_template.py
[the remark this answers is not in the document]
elliotgunton: Added `_set_namespace` to raise an error if trying to set a value

src/hera/workflows/cluster_workflow_template.py
elliotgunton: Well, we have `WorkflowsService.get_workflow` with more explicit parameters? Maybe we just keep this `Workflow.get` private

elliotgunton: We already released the `WorkflowTemplate.get` method in v5.3.0, from #642 [here](https://github.com/argoproj-labs/hera/pull/642/files#diff-578cbe46ffddea5458bdb9b1701e5ec32fe9ef4b8bb5334c5c09791e77da7ce0R42) - I need to use the same interface to override it and avoid calling get_workflow_template instead of \*\_cluster_*. Tbh I don't think it's a "mistake" to add it to the public interface, but now that we have we should probably leave it as is.

flaviuvadan: I think we could introduce a kwarg called `cluster: bool = False` to support creating cluster level workflow templates while maintaining backwards compatibility 🤔 thoughts?

src/hera/workflows/workflow_template.py
```diff
@@ -153,5 +162,26 @@ def build(self) -> TWorkflow:
...
+
+        if workflow.name:
+            # In case of a name being > TRUNCATE_LENGTH, we need to truncate the value to assign to generate_name
+            # otherwise the random suffix will make the generated name too long
+            workflow.generate_name = workflow.name[:TRUNCATE_LENGTH] + "-"
+            workflow.name = None
+
+        return workflow
+
+    def create_as_workflow(self, wait: bool = False, poll_interval: int = 5) -> TWorkflow:
```
flaviuvadan, line 177: Should generate name be an option here, which gets passed to `get_as_workflow`, perhaps? 🤔 That might give users more control over the behavior of the workflow

flaviuvadan (commented):
Looks great @elliotgunton 🎖️ I left a question re: interfacing with the generated name option. Will approve after the discussion/potential change

src/hera/workflows/workflow_template.py
elliotgunton: If you mean as a bool, I did consider this and it would be more flexible, but I don't think there's a use case for _not_ using the generated name as a workflow? Otherwise you'd `create_as_workflow(generate_name=False, ...` and not call it again.

So then if you mean as a string, that sounds more reasonable, optional and we can default it to `workflow.name[:TRUNCATE_LENGTH] + "-"` as before if not given 💯

elliotgunton: >we can truncate that one

I let the user pass in whatever they want and we don't truncate here which matches normal `Workflow` code - maybe we can add a validator on `generate_name`? Argo will truncate it regardless to add the random suffix if the string is too long

flaviuvadan: > Argo will truncate it regardless
Oh, I see, did not know this

> I let the user pass in whatever they want and we don't truncate here which matches normal Workflow code
This sounds great!

elliotgunton: >truncate it regardless
Seems I was wrong!

```
argo submit long-name.yaml
ERRO[2023-06-09T15:58:48.404Z] Create request failed                         error="Workflow.argoproj.io \"aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaj2vwz\" is invalid: metadata.labels: Invalid value: \"aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa\": must be no more than 63 characters"
FATA[2023-06-09T15:58:48.404Z] Failed to submit workflow: rpc error: code = InvalidArgument desc = Workflow.argoproj.io "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaj2vwz" is invalid: metadata.labels: Invalid value: "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa": must be no more than 63 characters
```

elliotgunton: Okay no it's weirder than that. If you submit with a generateName > 63 it will error. But if you submit with len(generateName) == 63, Argo will truncate the name to add the suffix... Think it would be helpful to add validators to `Workflow`

flaviuvadan: 😅 maybe this is a problem/inconsistency that upstream should fix? Or maybe there's a good reason for it

flaviuvadan (approved):
🚀

PR #673, opened 2023-06-06
Move to Annotated for mapping hera to models. Add from/to functions

### LINKED ISSUE #609: Read Workflows from yaml
Currently the code supports writing hera classes to a yaml files:
workflow.to_yaml()

It would be very helpful to be able to read yaml files to hear classes.
This way we can have partial class definitions in yaml files, and then load it into the code

workflow.from_yaml(file)

container.from_yaml(file)

elliotgunton:
**Pull Request Checklist**
- [x] Fixes #609
- [x] Tests added

**Description of PR**
This PR changes the Hera dev experience so we don't have to maintain hera<->model mappings in build functions.

This PR adds `from`/`to` function pairs for dict/yaml/file.


src/hera/workflows/workflow.py
```diff
@@ -436,5 +444,81 @@ def to_file(self, output_directory: Union[Path, str] = ".", name: str = "", *arg
         output_path.write_text(self.to_yaml(*args, **kwargs))
         return output_path.absolute()
 
+    @classmethod
+    def _from_model(cls: Workflow, model: _ModelWorkflow) -> Workflow:
```
flaviuvadan, line 448: This looks great! 

- What do you think about making use of this in the APIs? create, get, lint, etc
- What do you think about making this public?

src/hera/workflows/workflow.py
```diff
@@ -436,5 +444,81 @@ def to_file(self, output_directory: Union[Path, str] = ".", name: str = "", *arg
...
+
+    @classmethod
+    def from_yaml(cls: Workflow, yaml_file: Union[Path, str]) -> Workflow:
+        """Create a Workflow from a Workflow contained in a YAML file.
+
+        Usage:
+            my_workflow = Workflow.from_yaml(yaml_file)
+        """
+        model_workflow = _ModelWorkflow.parse_file(yaml_file)
+        return cls.from_model(model_workflow)
```
flaviuvadan, line 521: Small - it might be prudent to add some validation here unless `parse_file` doesn't already do it

src/hera/workflows/workflow.py
elliotgunton: >create, get, lint

Yes was thinking that would be a great use of it! (But would need separate functions for bw-compat.) Makes sense that you would get Hera objects from the workflows service

>making this public

I was leaning towards public but went private because we can go private -> public if we want. Also if the workflows service gives back hera objects then would clients of the library ever need to deal with `_ModelWorkflows`? They would be able to get hera objects from yaml and the live service

src/hera/workflows/workflow.py
elliotgunton: `parse_file` calls `parse_obj` which I believe will error if the object isn't well-formed. Copied the func from pydantic below

```
    @classmethod
    def parse_obj(cls: Type['Model'], obj: Any) -> 'Model':
        obj = cls._enforce_dict_if_root(obj)
        if not isinstance(obj, dict):
            try:
                obj = dict(obj)
            except (TypeError, ValueError) as e:
                exc = TypeError(f'{cls.__name__} expected dict not {obj.__class__.__name__}')
                raise ValidationError([ErrorWrapper(exc, loc=ROOT_KEY)], cls) from e
        return cls(**obj)
```

src/hera/workflows/_mixins.py
```diff
@@ -898,6 +900,18 @@ def _get_params_from_items(with_items: List[Any]) -> Optional[List[Parameter]]:
     return [Parameter(name=n, value=f"{{{{item.{n}}}}}") for n in with_items[0].keys()]
 
 
+class ParseFromYamlMixin(BaseMixin):
+    @classmethod
+    def _from_model(cls: ParseableT, model: Any) -> ParseableT:
```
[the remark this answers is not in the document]
elliotgunton: Replied in https://github.com/argoproj-labs/hera/pull/673/files/826179e63a2487deda705ecbca59b706ad5c8a4c#r1223018963

src/hera/workflows/workflow.py
flaviuvadan: >I was leaning towards public but went private because we can go private -> public if we want

Sounds great!

flaviuvadan (commented):
This approach LGTM @elliotgunton! Thanks for tackling it 🚀 Is the plan to introduce these transformations to other objects as well? Could I perhaps help extend it to other objects? I can open follow up PRs and we can coordinate on this one with a checklist of what needs to change

elliotgunton:
Yep exactly @flaviuvadan! This PR is a proof of concept for `Workflow` to confirm the approach. However, I was looking into `typing.Annotated` to see if we can avoid doing the attribute mapping manually in `Workflow.build` and then backwards in `Workflow.from_model` - see https://github.com/argoproj-labs/hera/pull/675 for the forwards direction, still thinking about how to do from_model with `Annotated`. Basically the approach using `Annotated` means Hera's Workflow's attributes know through their metadata stored with the attribute, where they map to in the model Workflow.

src/hera/workflows/workflow.py
elliotgunton: >the service should primarily use the generated objects

Makes sense to me! 💯 

>interfacing with argo through maintained objects should return maintained objects

Also makes sense, thanks for the example! 💯

flaviuvadan:
> Basically the approach using Annotated means Hera's Workflow's attributes know (through their metadata stored with the attribute) where they map to in the model Workflow.

This sounds great! How does it affect autocomplete and the experience of looking at types through IDE pop-ups?

elliotgunton:
@flaviuvadan autocomplete still works and uses the underlying type, so still LGTM! 
image

src/hera/workflows/_mixins.py
```diff
@@ -898,6 +921,121 @@ def _get_params_from_items(with_items: List[Any]) -> Optional[List[Parameter]]:
     return [Parameter(name=n, value=f"{{{{item.{n}}}}}") for n in with_items[0].keys()]
 
 
+def _model_attr_setter(attrs: List[str], model: BaseModel, value: Any):
```
flaviuvadan, line 924: I am slightly confused by the logic here - why is it necessary to check all but the last attr then set the last one only? Also, very picky comment, but `_set_model_attr` seems more aligned with the _action_ this performs

src/hera/workflows/cron_workflow.py
```diff
@@ -90,40 +137,44 @@ def build(self) -> TWorkflow:
...
             spec=CronWorkflowSpec(
-                concurrency_policy=self.concurrency_policy,
-                failed_jobs_history_limit=self.failed_jobs_history_limit,
-                schedule=self.schedule,
-                starting_deadline_seconds=self.starting_deadline_seconds,
-                successful_jobs_history_limit=self.successful_jobs_history_limit,
-                suspend=self.cron_suspend,
-                timezone=self.timezone,
-                workflow_metadata=None,
+                schedule="",
```
flaviuvadan, line 143:
```suggestion
                schedule=self.schedule,
```

elliotgunton: Haha yes that makes more sense as a default value! Then the annotation for `schedule` isn't strictly needed 🤔

flaviuvadan (dismissed):
This looks really amazing @elliotgunton 🚀 I checked out the branch to better understand what's going on. I love the ClusterWorkflowTemplate did not have to be touched in this PR 💪🏻 My only small piece of feedback was on using `self.schedule` on cron workflow

src/hera/workflows/_mixins.py
```diff
@@ -898,6 +921,130 @@ def _get_params_from_items(with_items: List[Any]) -> Optional[List[Parameter]]:
...
+    # BaseModel set to `curr`, then `getattr` on that BaseModel, using the last attribute
+    # name in attrs.
+    curr: BaseModel = model
+    for attr in attrs[:-1]:
+        curr = getattr(curr, attr)
+
+    return getattr(curr, attrs[-1])
+
+
+class ParseFromYamlMixin(BaseMixin):
```
flaviuvadan, line 949: Again, reiterating that I think a rename of this makes sense but it's not critical!

src/hera/workflows/cron_workflow.py
elliotgunton: Turns out this helped me realise there's a bug with this - `CronWorkflow._from_model(w.build())` - this will fail because the default init for a CronWorkflow needs a value for the schedule, so the line `hera_obj = cls()` fails. Making a change for this

samj1912 (approved):

Issue #670, opened 2023-06-05
Hera CLI design


elliotgunton:
Following a catchup with @flaviuvadan and @samj1912, this issue is to expand/formalise the idea of a Hera CLI, features it should support and a small roadmap/priority list for the CLI.

<html>
<body>
<b style="font-weight:normal;" id="docs-internal-guid-21d2b251-7fff-5f1a-a95f-40aefe9d75a9"><div dir="ltr" style="margin-left:0pt;" align="center">

Feature (priority order) | Depends on | Impact | T shirt size | Proof of concept / PR / links
-- | -- | -- | -- | --
CLI framework | - | Required to enable features | S |  #886
WT stub gen | CLI | Medium/high - would make TemplateRef usage much better, key part of WorkflowTemplate use/distribution | M |  
Detect Workflows / WorkflowTemplates | To be part of CLI but can figure out functionality in parallel | Low/Medium (might be required to enable features. Ease of use feature) | S/M |  #886
Submit from WT | CLI, Detect Workflows / WorkflowTemplates | Medium - would improve dev ex a lot | S | As a function on WorkflowTemplates, issue/suggestion here https://github.com/argoproj-labs/hera/issues/521 -> Done in - https://github.com/argoproj-labs/hera/pull/669
Python to YAML converter | CLI | Low/Medium - to_yaml() already exists on Workflows, this puts convenience in the CLI. Could be used to generate many at once. Useful for CI/CD and DevOps use cases | S |  #1440 
YAML to Python converter | CLI | Medium - existing users of Argo can instantly convert to using Hera, opens the door for better integrations | M | https://github.com/argoproj-labs/hera/issues/609 and https://github.com/argoproj-labs/hera/pull/673 Done in #886
User Workflow automatic documentation generator | CLI | Medium - we love docs! | S |  

</div></b>
</body>
</html>

# Example CLI Usage
## WorkflowTemplate Stub gen
`hera generate stubs <template-name> [--namespace <namespace>] [--cluster-template]`

Comments:
* Stubs wouldn’t be useful for Workflows, so just WTs and CWTs are accepted

## Submit from WorkflowTemplate
`hera submit --from <python-file-with-def>/<template-name> --e <entrypoint> --namespace <namespace>`

Comments:
* Unsure of the --from syntax - pass a file + template name? Separate args for each?

## ~Python to YAML converter~ Done in #1440 
`hera generate yaml --from <python-file-with-def>[/<template-name>] [--to <filepath>]`

[...]

PR #674, opened 2023-06-06
Add CLI framework files

elliotgunton:
**Pull Request Checklist**
- [x] Work towards #670 
- [ ] Tests added

**Description of PR**
Currently, ...

This PR adds/changes/fixes...


src/hera/_cli/__init__.py
```diff
@@ -0,0 +1 @@
+"""Hera's command line interface."""
```
flaviuvadan, line 1: I suggest we leave this public as there might be uses for it. Thoughts?

elliotgunton: Hmm, I think it depends on the API of the CLI, but it should be passing through to public functions we'll add elsewhere. We can always go public if want too, which is my preference unless the "might" becomes definite

flaviuvadan (commented):
Looks _awesome_ so far 🎖️ posted a question in our private chat about the organization of the CLI modules

flaviuvadan (approved):
This LGTM + we aligned on Slack re: the structure + direction of the CLI. Would you like to merge this PR @elliotgunton? Happy to review/facilitate integration in whatever order you'd like to use for delivering this

elliotgunton:
@flaviuvadan I want to get the docs together too (e.g. installation) for a minimal PR, but also leaving this branch to develop this on a feature branch could help? Either way would rather not merge for the moment

samj1912:
I'm going to close this PR for now. We can reopen this once we are working on the CLI.

Issue #679, opened 2023-06-12
UpdateCronWorkflowRequest complains about properly formatted CronWorkflow parameter


iameskild:
<h2>Pre-bug-report checklist</h2>

**1. This bug can be reproduced using YAML**
- [ ] Yes [👉 Please report a bug to the Argo Workflows GitHub 👈](https://github.com/argoproj/argo-workflows/issues/new/choose)
- [x] No

**2. This bug occurs when...**
- [x] running Hera code without submitting to Argo (e.g. when exporting to YAML)
- [ ] running Hera code and submitting to Argo

<h2>Bug report</h2>

**Describe the bug**
_A clear and concise description of what the bug is:_

I want to update my running CronWorkflow using the WorkflowServices API (`update_cron_workflow`), however I have been update to do so because the `UpdateCronWorkflowRequest` object complains whenever I provide the `CronWorkflow` I want to update.

_Error log if applicable_:
```python
    req = UpdateCronWorkflowRequest(cron_workflow=w)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "pydantic/main.py", line 341, in pydantic.main.BaseModel.__init__
pydantic.error_wrappers.ValidationError: 2 validation errors for UpdateCronWorkflowRequest
cronWorkflow -> metadata
  field required (type=value_error.missing)
cronWorkflow -> spec
  field required (type=value_error.missing)
```

**To Reproduce**
Full Hera code to reproduce the bug:
```py
# my wrapper function that creates a cron_workflow
# this function is used elsewhere in the codebase to return a `CronWorkflow` object 
# which is then can be created via `w.create()`
w = create_cwf_oject(...)
req = UpdateCronWorkflowRequest(cron_workflow=w)
wfs = WorkflowsService()
wfs.update_cron_workflow(
    req=req,
    name="my-cwf",
    namespace=global_config.namespace,
)
```

**Expected behavior**
_A clear and concise description of what you expected to happen:_

**Environment**
 - OS: [e.g. iOS]: MacOS
 - Browser: Chrome
 - Version of Hera: `5.3.0`
 - Version of Argo: `3.4.4` (server)  

**Additional context**

elliotgunton:
This is because the type for `UpdateCronWorkflowRequest` is a `hera.workflows.models.CronWorkflow` - it has required `metadata` and `spec`:

```py
class CronWorkflow(BaseModel):
[...]

PR #686, opened 2023-06-19
Set status=None if any status field is None

iameskild:
**Pull Request Checklist**
- [x] Fixes # Related to #679
- [ ] Tests added

**Description of PR**

I kept running into the following error message when I would try to update my CronWorkflow. 

```py
pydantic.error_wrappers.ValidationError: 2 validation errors for CronWorkflow
status -> active
  none is not an allowed value (type=type_error.none.not_allowed)
status -> lastScheduledTime
  none is not an allowed value (type=type_error.none.not_allowed)
```

Unfortunately this error would only appear if at least workflow has been created. Comparing the two models, `WorkflowStatus`  `CronWorkflowStatus` in v1alpha1.py, it seems that the former has all of it's fields set to optional and it makes sense that the same should apply to `CronWorkflowStatus`; this is what this PR does. 

With this fix, I was able to update my CronWorkflows regardless of the current status.

### DIFF
--- a/scripts/service.py
+++ b/scripts/service.py
@@ -190,7 +190,7 @@ class ServiceEndpoint:
 
         if resp.ok:
             resp_json = resp.json()
-            if "status" in resp_json and resp_json["status"]['active'] is None and resp_json["status"]['lastScheduledTime'] is None and resp_json["status"]['conditions'] is None:
+            if "status" in resp_json or resp_json["status"]['active'] is None or resp_json["status"]['lastScheduledTime'] is None or resp_json["status"]['conditions'] is None:
                 # this is a necessary special case as the status fields cannot be empty on the `CronWorkflowStatus`
                 # object. So, we overwrite the response with a value that allows the response to pass through safely.
                 # See `hera.scripts.service.ServiceEndpoint.__str__` for more details.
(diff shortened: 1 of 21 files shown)


elliotgunton:
This would be changing the auto-generated classes so doesn't look like the right solution here. I think we might need to change the `and`s to `or`s in [scripts/service.py#L193](https://github.com/argoproj-labs/hera/blob/2eb239e6c3af72e3726528e8123139116c98a0d3/scripts/service.py#L193) @flaviuvadan?

iameskild:
@elliotgunton that would make more sense. I just tested this for CronWorkflow and it seems to work. I just made that update to this PR.

elliotgunton:
Hey @iameskild - you'll need to change the specific line in main/scripts/service.py to `or`s and then run `make codegen` (and commit) as the code you've changed is generated from the script so is currently failing CI/CD because of the diff

elliotgunton (commented):
LGTM but would like @flaviuvadan to also take a look

flaviuvadan (approved):
Thank you for the fix @iameskild! 🎖️ LGTM

elliotgunton (approved):

PR #696, opened 2023-06-25
Add more documentation

flaviuvadan:
**Pull Request Checklist**
- [ ] Fixes #
- [ ] Tests added

**Description of PR**
This PR adds more inline documentation


mkdocs.yml
```diff
@@ -118,6 +118,10 @@ plugins:
             filters: ["!^_.*$"]
             show_bases: false
             separate_signature: true
+          selection:
+            docstring_style: numpy
```
samj1912, line 122: @flaviuvadan could we use google docstyle? It is easier to write docs using it as we don't have to worry about underlining with hyphens and it follows python syntax more closely (colons and tabs).

samj1912 (commented):
Thanks for this PR! Looks great, just a comment on the choice of docstyle. Would prefer if we use google style as it is more succinct to write docs using it and I find it easier to read alongisde python code.

mkdocs.yml
flaviuvadan: Yep, we can definitely do so! I am going to change the rest of the docs then

PR #698, opened 2023-06-27
Fix the artifact fanout example

elliotgunton:
* The artifact was being passed as a parameter
* The filepath being used got `PermissionError: [Errno 13] Permission denied: '/file'`, now using `/tmp/file`
* Use the None ArchiveStrategy so it's viewable in the UI

**Pull Request Checklist**
- [x] Fixes example
- [x] Tests added

### DIFF
--- a/examples/workflows/artifact_with_fanout.py
+++ b/examples/workflows/artifact_with_fanout.py
@@ -1,5 +1,4 @@
-from hera.workflows import DAG, Artifact, Workflow, script
-from hera.workflows.archive import NoneArchiveStrategy
+from hera.workflows import DAG, Artifact, NoneArchiveStrategy, Workflow, script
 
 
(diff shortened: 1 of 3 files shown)


examples/workflows/artifact_with_fanout.py
```diff
@@ -1,22 +1,23 @@
 from hera.workflows import DAG, Artifact, Workflow, script
+from hera.workflows.archive import NoneArchiveStrategy
```
flaviuvadan, line 2: Is this importable as `from hera.workflows import NoneArchiveStrategy`?

flaviuvadan (approved):
LGTM! Just left a small Q about an import

examples/workflows/artifact_with_fanout.py
elliotgunton: Yes it is! Fixing

PR #708, opened 2023-07-11
Script annotations for input Parameters

### LINKED ISSUE #705: Support annotations for input parameters for script
Make writing annotations for scripts easier by using typing.Annotated for additional information about parameters such as enum, default, desc. 

Current way of writing that is
```python
@script(inputs=[Parameter(name="an_int", enum=[1, 2, 3]), default=2])
def echo_int(an_int=2):
    print(an_int)
```

but with Annotated we can avoid the duplication of "an_int" and the default value

```python
@script()
def echo_int(an_int: Annotated[int, Parameter(enum=[1, 2, 3])]):
    print(an_int)
```
Or for all annotations
```python
@script()
def echo_int(an_int: Annotated[int, Parameter(default=1, description="an_int parameter", enum=[1, 2, 3])]):
    print(an_int)
```

The ValueFrom has a problem described in #704 so we might want to deal with that later
```python
@script()
def echo_string(a_string: Annotated[str, InputParameter(value_from=ValueFrom( config_map_key_ref=ConfigMapKeySelector(
                    name="simple-parameters",
                    key="a_string_param"
                )))]):
    print(a_string)
```

dejamiko:
**Pull Request Checklist**
- [x] Fixes #705 
- [x] Tests added

**Description of PR**
This PR adds some basic support for annotations for script. We can avoid duplication of parameter names and default values. See the old version

```python
@script(
    inputs=[
        Parameter(name="an_int", description="an_int parameter", default=1, enum=[1, 2, 3]), 
        Parameter(name="a_bool", description="a_bool parameter", default=True, enum=[True, False]), 
        Parameter(name="a_string", description="a_string parameter", default="a", enum=["a", "b", "c"])
    ]
)
def echo_all(an_int=1, a_bool=True, a_string="a"):
    print(an_int)
    print(a_bool)
    print(a_string)
```

vs the new one:

```python
@script()
def echo_all(
    an_int: Annotated[int, Parameter(description="an_int parameter", default=1, enum=[1, 2, 3])], 
    a_bool: Annotated[bool, Parameter(description="a_bool parameter", default=True, enum=[True, False])], 
    a_string: Annotated[str, Parameter(description="a_string parameter", default="a", enum=["a", "b", "c"])]
):
    print(an_int)
    print(a_bool)
    print(a_string)
```

We also allow for renaming parameters, using YAML rules for naming

```python
@script()
def echo(
    a_name: Annotated[str, Parameter(name="another-name")],
):
    print(a_name)
```

## kebab-case support for runner
Since we now allow for parameters to have a new name and the naming rules come from YAML, we allow kebab-case names for parameters which could be used in a script runner. A mapping of kwargs was created that maps the parameter name from annotations to the one found in the python function signature.

```python
@script()
def function_kebab(
    a_but_kebab: Annotated[int, Parameter(name="a-but-kebab")] = 2,
    b_but_kebab: Annotated[str, Parameter(name="b-but-kebab")] = "foo",
) -> Output:
    return Output(output=[Input(a=a_but_kebab, b=b_but_kebab)])
```


docs/getting-started/walk-through/advanced-hera-features.md
```diff
@@ -89,12 +89,30 @@ conflicts.
...
 dictionary before using the feature:
 
+```py
+global_config.experimental_features["NAME_OF_FEATURE"] = True
+```
+
+### Currently supported experimental features:
+
+#### `RunnerScriptConstructor`
+What is this used for? [runner script example](../../examples/workflows/callable_script.md)
```
elliotgunton, line 104:
```suggestion
The `RunnerScriptConstructor` found in `hera.workflows.script` and seen in the
[callable script example](../../examples/workflows/callable_script.md) is a robust way to run Python functions on Argo.
The image used by the script should be built from the source code package itself and its dependencies, so that the
source code's functions, dependencies, and Hera itself are available to run. The `RunnerScriptConstructor` is also
compatible with Pydantic so supports deserializing inputs to Python objects and serializing outputs to json strings. It
must be enabled with the `script_runner` feature flag as below.
```

src/hera/workflows/parameter.py
```diff
@@ -93,4 +94,15 @@ def as_output(self) -> _ModelParameter:
         )
 
 
-__all__ = ["Parameter"]
+class InputParameterAnnotation(BaseModel):
```
samj1912, line 97: Ideally would like this to be the same as the Parameter class so that we don't diverge too much. We can also add some annotation specific fields that don't get exported to the hera.models.Parameters.

One such field would be `alias`

src/hera/workflows/runner.py
```diff
@@ -24,6 +31,7 @@ def _ignore_unmatched_kwargs(f):
     def inner(**kwargs):
         # filter out kwargs that are not part of the function signature
         # and transform them to the correct type
+
```
elliotgunton, line 34: Extra line added

src/hera/workflows/runner.py
```diff
@@ -76,6 +84,17 @@ def _is_str_kwarg_of(key: str, f: Callable):
         return False
 
 
+def map_kwargs(function: Callable, kwargs: dict):
```
elliotgunton, line 87: Tiny suggestion because we're changing the kwarg keys only. Function is 💯 
```suggestion
def map_kwargs_keys(function: Callable, kwargs: dict):
```

src/hera/workflows/script.py
```diff
@@ -232,16 +236,32 @@ def _build_script(self) -> _ModelScriptTemplate:
...
+        param = Parameter(name=p.name, default=default)
 
-    if len(source_signature) == 0:
+        # add docs for this feature in docs/getting-started/walk-through/advanced-hera-features.md
+        if global_config.experimental_features["script_annotations"]:
+            if get_origin(p.annotation) is Annotated and isinstance(get_args(p.annotation)[1], Parameter):
+                annotation = get_args(p.annotation)[1]
+                for attr in ["enum", "description", "default", "name"]:
+                    if hasattr(annotation, attr) and getattr(annotation, attr) is not None:
+                        setattr(param, attr, getattr(annotation, attr))
```
elliotgunton, line 257: This will overwrite the `default` of `param` set on L249, maybe we want to raise an error in that case (if `param.default is not MISSING`) so a user can't try to set default in two places?

tests/test_runner.py
```diff
@@ -9,8 +9,8 @@
     [
         (
             "examples.workflows.callable_script:my_function",
-            [{"name": "input", "value": '{"a": 2, "b": "bar"}'}],
-            '{"output": [{"a": 2, "b": "bar"}]}',
+            [{"name": "input", "value": '{"a": 3, "b": "barbar"}'}],
+            '{"output": [{"a": 3, "b": "barbar"}]}',
```
elliotgunton, line 13: Unrelated change

src/hera/workflows/script.py
```diff
@@ -232,16 +236,34 @@ def _build_script(self) -> _ModelScriptTemplate:
...
+        param = Parameter(name=p.name, default=default)
 
-    if len(source_signature) == 0:
+        # add docs for this feature in docs/getting-started/walk-through/advanced-hera-features.md
+        if global_config.experimental_features["script_annotations"]:
+            if get_origin(p.annotation) is Annotated and isinstance(get_args(p.annotation)[1], Parameter):
+                annotation = get_args(p.annotation)[1]
+                for attr in ["enum", "description", "default", "name"]:
+                    if hasattr(annotation, attr) and getattr(annotation, attr) is not None:
+                        if attr == "default" and getattr(param, attr) is not None:
```
elliotgunton, line 257: We can stick to the direct access for this special case (easier to understand)
```suggestion
                        if attr == "default" and param.default is not None:
```

docs/examples/workflows/callable_script.md
```diff
@@ -71,11 +77,28 @@
...
+    def function_kebab_object(input_values: Annotated[Input, Parameter(name="input-values")]) -> Output:
+        return Output(output=[input_values])
+
+
     with Workflow(name="my-workflow") as w:
         with Steps(name="my-steps") as s:
             my_function(arguments={"input": Input(a=2, b="bar")})
             str_function(arguments={"input": Input(a=2, b="bar").json()})
             another_function(arguments={"inputs": [Input(a=2, b="bar"), Input(a=2, b="bar")]})
+            function_kebab(arguments={"a_but_kebab": 3, "b_but_kebab": "bar"})
```
samj1912, line 100: Are we sure Argo lint will let us do this? At a yaml level the mapping has a discrepancy. Let's confirm.

dejamiko: I think this was an accident, should be `a-but-kebab` instead

elliotgunton: It does actually work though, interestingly... not sure about naming limits on parameters. This is accepted and runs:

```yaml
kind: Workflow
metadata:
  generateName: my-workflow-
spec:
  entrypoint: whalesay
  templates:
    - name: whalesay
      container:
        args:
          - '{{inputs.parameters.test_parameter}}'
        command:
          - cowsay
        image: docker/whalesay:latest
      inputs:
        parameters:
          - name: test_parameter
            value: The name has an underscore
```

src/hera/workflows/script.py
```diff
@@ -232,16 +236,35 @@ def _build_script(self) -> _ModelScriptTemplate:
...
-            source_signature[p.name] = p.default
+            default = p.default
         else:
-            source_signature[p.name] = MISSING
+            default = MISSING
+
+        param = Parameter(name=p.name, default=default)
 
-    if len(source_signature) == 0:
+        if global_config.experimental_features["script_annotations"]:
```
samj1912, line 251: We should try and reduce the complexity in terms of nesting here so that the logic becomes easier to follow.

We should be able to check for the negation of these conditions and use `continue`.

dejamiko: Good call, I did that now

src/hera/workflows/parameter.py
```diff
@@ -20,6 +20,13 @@ class Parameter(_ModelParameter):
     for Steps and Tasks to assign values.
     """
 
+    name: Optional[str] = None  # type: ignore
+
+    def check_name(self):
```
elliotgunton, line 25: Should be a private function
```suggestion
    def _check_name(self):
```

src/hera/workflows/script.py
```diff
@@ -232,16 +236,41 @@ def _build_script(self) -> _ModelScriptTemplate:
 def _get_parameters_from_callable(source: Callable) -> Optional[List[Parameter]]:
     # If there are any kwargs arguments associated with the function signature,
     # we store these as we can set them as default values for argo arguments
-    source_signature: Dict[str, Optional[object]] = {}
+    #
+    # Take all the annotation information and make sure it is stored.
```
elliotgunton, line 240: Comment not relevant here - something like this might be helpful above the `get_args` below
```py
        # Take parameter attribute values from annotation
        annotation = get_args(p.annotation)[1]
        for attr in ["enum", "description", "default", "name"]:
```

(But the code is quite self explanatory so it's up to you - comment would be helpful to newbies)

src/hera/workflows/script.py
```diff
@@ -232,16 +236,41 @@ def _build_script(self) -> _ModelScriptTemplate:
...
+                    "The default cannot be set via both the function parameter default and the annotation's default"
+                )
+
+            setattr(param, attr, getattr(annotation, attr))
+
+    if len(parameters) == 0:
         return None
-    return [Parameter(name=n, default=v) for n, v in source_signature.items()]
+
+    return parameters
```
elliotgunton, line 273: Thanks to your changes we can now actually streamline this even more 😎

```suggestion
    return parameters or None
```

src/hera/workflows/script.py
```diff
@@ -232,16 +236,41 @@ def _build_script(self) -> _ModelScriptTemplate:
...
+                continue
+
+            if attr == "default" and param.default is not None:
+                raise ValueError(
+                    "The default cannot be set via both the function parameter default and the annotation's default"
+                )
+
+            setattr(param, attr, getattr(annotation, attr))
+
+    if len(parameters) == 0:
```
elliotgunton, line 270: This line could then be removed

src/hera/workflows/runner.py
```diff
@@ -101,11 +119,13 @@ def _runner(entrypoint: str, kwargs_list: Any) -> str:
         if "name" not in kwarg or "value" not in kwarg:
             continue
         # sanitize the key for python
-        key = cast(str, serialize(kwarg["name"])).replace("-", "_")
+        key = cast(str, serialize(kwarg["name"]))
```
elliotgunton, line 122: We've lost the `.replace("-", "_")` behaviour for non-experimental users. We could either do it here or in the `_map_keys` function?

I'm not sure whether the script runner will pick up a global config defined next to the given function in the module though, might need an extra parameter or alternative runner 🤔

src/hera/workflows/runner.py
```diff
@@ -76,6 +83,17 @@ def _is_str_kwarg_of(key: str, f: Callable):
         return False
 
 
+def _map_keys(function: Callable, kwargs: dict):
+    """Change the kwargs's keys to use the python name instead of the parameter name which could be kebab case."""
+    mapped_kwargs = {}
```
elliotgunton, line 88:
```suggestion
def _map_keys(function: Callable, kwargs: dict):
    """Change the kwargs's keys to use the python name instead of the parameter name which could be kebab case."""
    if not global_config.experimental_features["script_annotations"]:
        return {key.replace("-", "_"): value for key, value in kwargs.items()}

    mapped_kwargs = {}
```

elliotgunton (commented):
We need to add the old behaviour back in but otherwise LGTM!

src/hera/workflows/runner.py
samj1912: I think it's OK to break in this case as it's a experimental feature.

samj1912: Actually I see the suggestion you have made. That makes more sense.

src/hera/workflows/script.py
```diff
@@ -198,12 +205,18 @@ def _build_script(self) -> _ModelScriptTemplate:
         assert isinstance(self.constructor, ScriptConstructor)
         image_pull_policy = self._build_image_pull_policy()
 
+        env = self._build_env()
+        if global_config.experimental_features["script_annotations"]:
```
elliotgunton, line 209: For this run-time side of things I think it makes sense to only add the env var for `RunnerScriptConstructor`s? Inline scripts won't have the function definition itself to derive from.

```suggestion
        if isinstance(self.constructor, RunnerScriptConstructor) and global_config.experimental_features["script_annotations"]:
```

docs/getting-started/walk-through/advanced-hera-features.md
```diff
@@ -89,12 +89,35 @@ conflicts.
...
+compatible with Pydantic so supports deserializing inputs to Python objects and serializing outputs to json strings. It
+must be enabled with the `script_runner` feature flag as below.
+
 ```py
 global_config.experimental_features["script_runner"] = True
 ```
+
+
+#### Script Annotations
+An annotation based system for simplifying script parameter creation, as seen in the [script annotations example](../../examples/workflows/script_annotations_combined_new.md)
```
elliotgunton, line 117: Would be good to demonstrate the name aliasing aspect and expand this bit of documentation as well (as it's the only docs we have for this). Feel free to expand and sell it a bit more!

```suggestion
An annotation based system for simplifying script parameter creation, as seen in the [script annotations example](../../examples/workflows/script_annotations_combined_new.md).

Script annotations can work on top of the `RunnerScriptConstructor` for name aliasing of function
parameters, in particular to ensure a public `kebab-case` parameter, while using a `snake_case`
Python function parameter. When using a `RunnerScriptConstructor`, an environment variable
`hera__script_annotations` will be added to the Script template.

Script annotations also work with the regular `InlineScriptConstructor` for
generating valid template parameters in the yaml instead of using the `@script` decorator.
```

Maybe add a Python snippet showing old vs new?

src/hera/workflows/runner.py
```diff
@@ -76,6 +84,20 @@ def _is_str_kwarg_of(key: str, f: Callable):
         return False
 
 
+def _map_keys(function: Callable, kwargs: dict):
+    """Change the kwargs's keys to use the python name instead of the parameter name which could be kebab case."""
+    if os.environ.get("script_annotations", None) is not None:
```
elliotgunton, line 89: Just to be clear it's Hera's annotation and avoid naming collisions

```suggestion
    if os.environ.get("hera__script_annotations", None) is not None:
```

src/hera/workflows/script.py
```diff
@@ -198,12 +205,18 @@ def _build_script(self) -> _ModelScriptTemplate:
...
+            if not env:
+                env = []
+            env.append(EnvVar(name="script_annotations", value=""))
```
elliotgunton, line 212:
```suggestion
            env.append(EnvVar(name="hera__script_annotations", value=""))
```

elliotgunton (commented):
More docs, slight change on env var then LGTM! :shipit:

elliotgunton (approved):

PR #710, opened 2023-07-12
Fix: script function called under workflow no longer returns a Step

### LINKED ISSUE #709: Bug: Workflow-level callable scripts

- [x] No

- [ ] running Hera code and submitting to Argo

<h2>Bug report</h2>

**Describe the bug**
Calling a decorated function under a Workflow context creates a Step as the returned object. The template itself gets added to the Workflow (correctly? This pattern is used in various examples: `script_with_resources`, `script_with_resources`, `colored_logs`, `retry_script`, `hello_world` but is working by accident).

In particular, in the [hello world example](https://hera.readthedocs.io/en/stable/examples/workflows/hello_world/) yaml, the assignment `s="hello"` doesn't make it into the yaml, nor does it work using `arguments` so I would call it a bug here definitely:

```py
with Workflow(generate_name="hello-world-", entrypoint="s") as w:
    hello(arguments={"s": "world"})
```

**To Reproduce**
```py
from hera.workflows import Workflow, Script, Step, script

@script()
def hello_world():
     print("Hello world")

    x = hello_world()
    assert isinstance(x, Step)  # passes (but should not!)
    assert isinstance(w.templates[0], Script)  # passes - do we want to keep this behaviour?
```

**Expected behavior**
Calling a function within a Workflow context but not in a Steps or DAG should either error with `InvalidType` or more preferred would be adding the script as a template to the Workflow so we can create `WorkflowTemplates` easily consisting of many `@script` decorated functions, so I think `x` should be a `Template` in the example above?

**Environment**
 - OS: WSL
 - Browser: N/A
 - Version of Hera: 5.5.2
 - Version of Argo: N/A

**Additional context**

elliotgunton:
**Pull Request Checklist**
- [x] Fixes #709 
- [x] Tests added

**Description of PR**
* Fix hello_world example
* Keep the behaviour that the script template itself is added to the
list of templates in the Workflow
* Remove the behaviour that a `Step` object is returned from the call
of the script
* Change to look-before-you-leap for returned object from the
CallableTemplateMixin's `__call__` to make it clearer what's happening

### DIFF
--- a/examples/workflows/hello_world.py
+++ b/examples/workflows/hello_world.py
@@ -6,5 +6,9 @@ def hello(s: str):
     print("Hello, {s}!".format(s=s))
 
 
-with Workflow(generate_name="task-exit-handler-", entrypoint="s") as w:
-    hello(s="hello")
+with Workflow(
+    generate_name="hello-world-",
+    entrypoint="hello",
+    arguments={"s": "world"},
+) as w:
+    hello()
--- a/src/hera/workflows/_mixins.py
+++ b/src/hera/workflows/_mixins.py
@@ -26,7 +26,7 @@ from hera.workflows._context import SubNodeMixin, _context
 from hera.workflows.artifact import Artifact
 from hera.workflows.env import Env, _BaseEnv
 from hera.workflows.env_from import _BaseEnvFrom
-from hera.workflows.exceptions import InvalidTemplateCall, InvalidType
+from hera.workflows.exceptions import InvalidTemplateCall
 from hera.workflows.metrics import Metrics, _BaseMetric
 from hera.workflows.models import (
     HTTP,
@@ -600,16 +600,17 @@ class CallableTemplateMixin(ArgumentsMixin):
     """`CallableTemplateMixin` provides the ability to 'call' the template like a regular Python function.
 
     The callable template implements the `__call__` method for the inheritor. The `__call__` method supports invoking
-    the template as a regular Python function. The call must be executed within an active context, such as a `DAG` or
-    `Steps` since the call returns either a `Step` or a `Task` depending on the active context (`Step` for `Steps` and
-    `Task` for `DAG`, respectively). Note that `Steps` also supports calling templates in a parallel steps context
-    via using `Steps(...).parallel()`. When the call is executed and the template does not exist on the active
-    context, i.e. the workflow, it is automatically added for the user. Note that invoking the same template multiple
-    times does *not* result in the creation/addition of the same template to the active context/workflow. Rather, a
-    union is performed, so space is saved for users on the templates field and templates are not duplicated.
+    the template as a regular Python function. The call must be executed within an active context, which is a
+    `Workflow`, `DAG` or `Steps` context since the call optionally returns a `Step` or a `Task` depending on the active
+    context (`None` for `Workflow`, `Step` for `Steps` and `Task` for `DAG`). Note that `Steps` also supports calling
+    templates in a parallel steps context via using `Steps(...).parallel()`. When the call is executed and the template
+    does not exist on the active context, i.e. the workflow, it is automatically added for the user. Note that invoking
+    the same template multiple times does *not* result in the creation/addition of the same template to the active
+    context/workflow. Rather, a union is performed, so space is saved for users on the templates field and templates are
+    not duplicated.
     """
 
-    def __call__(self, *args, **kwargs) -> Union[Step, Task]:
+    def __call__(self, *args, **kwargs) -> Union[None, Step, Task]:
         if "name" not in kwargs:
             kwargs["name"] = self.name  # type: ignore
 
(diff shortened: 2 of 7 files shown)


tests/test_unit/test_workflow.py
```diff
@@ -65,3 +69,36 @@ def test_workflow_from_yaml():
...
+    # GIVEN
+    with Workflow(name="w"):
+        returned_from_script = hello()
+
+    # THEN
+    assert returned_from_script is None
+
+
+def test_workflow_callable_container_raises_error():
+    with pytest.raises(InvalidTemplateCall) as e:
```
dejamiko, line 89: Maybe we should have more negative testing throughout for combining different constructs (Workflows, Scripts, DAGs, Steps)

elliotgunton: Currently I think only a `ContainerSet` would be another negative example. However I've just realised we should separate out the error for Script/Container as the error says `Callable Template 'whalesay' is not under a Workflow` when it is

dejamiko: Well spotted!

examples/workflows/hello_world.py
```diff
@@ -6,5 +6,9 @@ def hello(s: str):
     print("Hello, {s}!".format(s=s))
 
 
-with Workflow(generate_name="task-exit-handler-", entrypoint="s") as w:
-    hello(s="hello")
+with Workflow(
+    generate_name="hello-world-",
+    entrypoint="hello",
+    arguments={"s": "world"},
```
flaviuvadan, line 12: Should `arguments={"s": "world"},` go to `hello`?

src/hera/workflows/_mixins.py
```diff
@@ -632,21 +633,39 @@ def __call__(self, *args, **kwargs) -> Union[Step, Task]:
...
+                    f"Callable Template '{self.name}' is not callable under a Workflow"  # type: ignore
+                )
+            if isinstance(_context.pieces[-1], (Steps, Parallel)):
+                return Step(*args, template=self, **kwargs)
 
-            return Task(*args, template=self, **kwargs)
-        except InvalidType:
-            pass
+            if isinstance(_context.pieces[-1], DAG):
+                return Task(*args, template=self, **kwargs)
```
flaviuvadan, line 664: This is very very nice!

examples/workflows/hello_world.py
elliotgunton: They do go to `hello` - this example is actually the same as the upstream `arguments_parameters` [here](https://hera.readthedocs.io/en/stable/examples/workflows/upstream/arguments_parameters/) except using a script!

src/hera/workflows/_mixins.py
```diff
@@ -632,21 +633,39 @@ def __call__(self, *args, **kwargs) -> Union[Step, Task]:
...
+        from hera.workflows.steps import Parallel, Step, Steps
+        from hera.workflows.task import Task
+        from hera.workflows.workflow import Workflow
+
+        if _context.pieces:
+            if isinstance(_context.pieces[-1], Workflow):
+                # Notes on callable templates under a Workflow:
+                # * If the user calls a script directly under a Workflow (outside of a Steps/DAG) then we add the script
+                #   template to the workflow and return None.
+                # * Containers, ContainerSets and Data objects (i.e. subclasses of CallableTemplateMixin) are already
```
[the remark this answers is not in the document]
elliotgunton: This is a question of whether you want to be able to add containers as a template through a callable (I would say no). The options are:

### 1. WAD: use a function to retrieve the container defined elsewhere (but returning a fresh instantation)
```py
def get_container():
    return Container(
    name="whalesay",
    image="docker/whalesay:latest",
    command=["cowsay"],
    args=["{{inputs.parameters.message}}"],
    inputs=Parameter(name="message"),
)

with Workflow(
    generate_name="arguments-parameters-",
    entrypoint="whalesay",
    arguments=Parameter(name="message", value="hello world"),
) as w:
    c = get_container()  # Container is instantiated in the function call and added correctly, c is now callable in steps/dags
```

### 2. Enhancement: make Containers callable under `Workflow`s
```py
whalesay = Container(
    name="whalesay",
    image="docker/whalesay:latest",
    command=["cowsay"],
    args=["{{inputs.parameters.message}}"],
    inputs=Parameter(name="message"),
)

with Workflow(
    generate_name="arguments-parameters-",
    entrypoint="whalesay",
    arguments=Parameter(name="message", value="hello world"),
) as w:
    whalesay()  # Container is already instantiated outside so this doesn't currently work, should this return something or None?
```

I'm hesitant to add that functionality as we somewhat accidentally added it for script-decorated functions. It encourages bad practice (IMO) of global variables - e.g. what if the `whalesay` container is modified in code elsewhere (changing the image etc) without me realising? We should try to stay close to being a definition language (as opposed to going full OOP) as a yaml alternative. The script-decorator works as a workflow-level callable because it's a fresh function call each time. Making the container callable under Workflows also confuses the callable container pattern under steps/dags - should a user be given a container from the call? e.g.:

```py
with Workflow(
    generate_name="arguments-parameters-",
    entrypoint="whalesay",
    arguments=Parameter(name="message", value="hello world"),
) as w:
    container = whalesay()
    with Steps(name="s"):
        # Do I call `container()` or `whalesay()` here?
        whalesay()
        container()
```

PR #718, opened 2023-07-20
Script annotations artifacts

### LINKED ISSUE #717: Support annotations for input artifacts for script
Similarly as in #705 , we can use annotations for an easier way of using arifacts as inputs for scripts.

The cases could be:

Artifact as a Path
```python
@script()
def read_artifact(an_artifact: Annotated[Path, Artifact(name="my-artifact", path=ARTIFACT_PATH)]) -> str:
    output = an_artifact.read_text()
    print(output)
    return output
```
and as an object
```python
class MyArtifact(BaseModel):
    a = "a"
    b = "b"


@script()
def read_artifact(an_artifact: Annotated[MyArtifact, Artifact(name="my-artifact", path=ARTIFACT_PATH)]) -> str:
    output = an_artifact.a + an_artifact.b
    print(output)
    return output
```

dejamiko:
**Pull Request Checklist**
- [x] Fixes #717 
- [x] Tests added

**Description of PR**
In Hera we are able to specify `Artifact`s in `inputs`:

```python
@script(inputs=Artifact(name="my-artifact", path="/tmp/file"))
def read_artifact():
    with open("/tmp/file") as a_file:
        print(a_file.read())
```

But by using annotations we can avoid repeating the `path` of the file, and access the artifact as if it's a regular Python argument of the given type:

```python
@script()
def read_artifact(an_artifact: Annotated[Path, Artifact(name="my-artifact", path="/tmp/file")]):
    print(an_artifact.read_text())
```

The path duplication is no longer necessary since `an_artifact` is a `Path` object.

The fields allowed in the `Artifact` annotations are: `name`, `path`, and `loader`.
For `Artifact`s, we allow three types of loaders. Using `ArtifactLoader`, we have `file`, `json`, 
and `None`.
With no loader, the `path` attribute of `Artifact` is extracted and can be subsequently used in the function body by referring to the function parameter. This can be seen above.
When the loader is set to `file`, the function parameter will be the contents of the file 
stored at `path`. 

```python
@script()
def read_artifact(
    an_artifact: Annotated[str, Artifact(name="my-artifact", path=ARTIFACT_PATH, loader=ArtifactLoader.file)]
) -> str:
    return an_artifact
```

This loads the contents of the file at `ARTIFACT_PATH` to the argument `an_artifact` and subsequently can be used as a string inside the function.

When the loader is set to `json`, the contents of the file at `path` are read and parsed to `json`.

```python
class MyArtifact(BaseModel):
    a = "a"
    b = "b"


@script()
def read_artifact(
    an_artifact: Annotated[MyArtifact, Artifact(name="my-artifact", path=ARTIFACT_PATH, loader=ArtifactLoader.json)]
) -> str:
    return an_artifact.a + an_artifact.b
```

Here, we have a json representation of `MyArtifact` stored at `ARTIFACT_PATH`. We can load it with `ArtifactLoader.json` and then use `an_artifact` as an instance of `MyArtifact` inside the function.


elliotgunton:
We'll need to merge #708 first, and then you can rebase this branch to still merge to main. atm this PR includes all the #708 changes

src/hera/workflows/artifact.py
```diff
@@ -24,6 +25,13 @@
 )
 
 
+class LoaderEnum(Enum):
+    """The enum with possible loader options."""
```
elliotgunton, line 29:
```suggestion
    """Enum for artifact loader options."""
```

tests/test_runner.py
```diff
@@ -39,3 +43,58 @@ def test(entrypoint, kwargs_list, expected_output):
...
+
+
+@pytest.mark.parametrize(
+    "entrypoint",
+    ["tests.artifact_annotations.script_annotations_artifact_wrong_loader:read_artifact"],
+)
+def test_artifact_wrong_loader(entrypoint):
+    # GIVEN
+    kwargs_list = []
+    global_config.experimental_features["script_annotations"] = True
```
elliotgunton, line 93: actual tests should be using the `global_config_fixture` now https://github.com/argoproj-labs/hera/blob/4bf33d05343efcc580fcb270a7e7ec951e51c556/tests/conftest.py#L7

tests/test_unit/test_script.py
```diff
@@ -0,0 +1,70 @@
...
+    md = importlib.import_module(module)
+    importlib.reload(md)
+    function = getattr(md, function_name)
+
+    # WHEN
+    params, artifacts = _get_parameters_and_artifacts_from_callable(function)
+
+    # THEN
+    assert params is None
+    assert artifacts is not None
```
elliotgunton, line 63: Will need to be updated with the change to remove usage of `None`

src/hera/workflows/runner.py
```diff
@@ -84,15 +84,25 @@ def _is_str_kwarg_of(key: str, f: Callable):
         return False
 
 
-def _map_keys(function: Callable, kwargs: dict):
-    """Change the kwargs's keys to use the python name instead of the parameter name which could be kebab case."""
-    if os.environ.get("hera__script_annotations", None) is not None:
-        return {key.replace("-", "_"): value for key, value in kwargs.items()}
```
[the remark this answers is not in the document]
dejamiko: Must be some merging mistake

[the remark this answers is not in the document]
dejamiko: Also, just looking at it now, should it not be inversed?

```python
if os.environ.get("hera__script_annotations", None) is None:
    return {key.replace("-", "_"): value for key, value in kwargs.items()}
```

tests/test_runner.py
```diff
@@ -39,3 +43,58 @@ def test(entrypoint, kwargs_list, expected_output):
...
+        (
+            "tests.artifact_annotations.script_annotations_artifact_object:read_artifact",
+            """{"a": "Hello ", "b": "there!"}""",
+            "Hello there!",
+        ),
+        (
+            "tests.artifact_annotations.script_annotations_artifact_file:read_artifact",
+            "Hello there!",
+            "Hello there!",
+        ),
```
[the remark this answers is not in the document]
dejamiko: We want to load a path from an artifact?

src/hera/workflows/runner.py
```diff
@@ -67,7 +67,7 @@ def _parse(value, key, f):
         return value
     try:
         return json.loads(value)
-    except json.JSONDecodeError:
+    except (json.JSONDecodeError, TypeError):
```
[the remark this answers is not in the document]
dejamiko: It's because even though the JSON is parsed in the loader.json case, the code attempts to parse it again and fails

src/hera/workflows/runner.py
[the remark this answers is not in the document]
elliotgunton: Yes it should be inverted! Good spot

tests/test_runner.py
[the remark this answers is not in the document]
elliotgunton: Yes, we came up with some sensible use-case predictions - a user might want a file artifact containing lots of file paths!

src/hera/workflows/artifact.py
```diff
@@ -97,6 +108,11 @@ def as_name(self, name: str) -> _ModelArtifact:
         artifact.name = name
         return artifact
 
+    @classmethod
+    def get_input_attributes(cls):
+        """Return the attributes used for input artifact annotations."""
+        return ["name", "path", "archive"]
```
elliotgunton, line 114: Archive would only be for outputs. More info on https://argoproj.github.io/argo-workflows/fields/#artifact

src/hera/workflows/artifact.py
```diff
@@ -24,6 +25,13 @@
 )
 
 
+class LoaderEnum(Enum):
```
flaviuvadan, line 28: Small - can we call this `Loader`? This would be consistent with the other use of `Enum` such as `ImagePullPolicy`

flaviuvadan: Just in case we may use `Loader` in other places in the future, an alternative is `ArtifactLoader`

src/hera/workflows/artifact.py
```diff
@@ -97,6 +108,11 @@ def as_name(self, name: str) -> _ModelArtifact:
         artifact.name = name
         return artifact
 
+    @classmethod
+    def get_input_attributes(cls):
+        """Return the attributes used for input artifact annotations."""
+        return ["name", "path"]
```
elliotgunton, line 114: I think we need to review the input attributes - `from_`, `from_expression` and `artifact_gc` I think should all be possible inputs as well

docs/examples/workflows/script_annotations_artifacts_raw_old.md
```diff
@@ -0,0 +1,80 @@
+# Script Annotations Artifacts Raw Old
```
flaviuvadan, line 1: @dejamiko I wonder if the upstream examples can act as the "old" ones, and then the new ones here can primarily focus on the `Annotated` version? 🤔 cc: @elliotgunton

examples/workflows/script_annotations_artifacts_oss_new.py
```diff
@@ -0,0 +1,38 @@
...
+            create_bucket_if_not_present=True,
+            endpoint="http://oss-cn-hangzhou-zmf.aliyuncs.com",
+            key="test/mydirectory/",
+            lifecycle_rule={"name": "my-oss-rule", "key": "ruleKey"},
+            secret_key_secret={"name": "my-oss-credentials", "key": "secretKey"},
+            security_token="oss-token",
+        ),
+    ]
+) -> str:
+    return my_artifact
```
flaviuvadan, line 33: Will this return work by default if someone does not use the experimental script runner?

docs/examples/workflows/script_annotations_artifacts_raw_old.md
elliotgunton: I agree we should make annotations the standard method going forward and avoid overloading the examples section (the old examples don't add value once someone knows about the annotation method - it's only good for the walkthrough section).

dejamiko: I added them to make sure the annotations generate the same exact yaml as the old method. I am happy to move all those tests to some sort of unit tests instead of polluting examples

src/hera/workflows/task.py
```diff
@@ -49,7 +49,91 @@ class Task(
     ParameterMixin,
     ItemMixin,
 ):
-    """Task is used to run a given template within a DAG. Must be instantiated under a DAG context."""
+    r"""Task is used to run a given template within a DAG. Must be instantiated under a DAG context.
```
elliotgunton, line 52: Restore file

src/hera/workflows/workflow.py
```diff
@@ -405,13 +405,6 @@ def wait(self, poll_interval: int = 5) -> TWorkflow:
         assert self.namespace is not None, "workflow namespace not defined"
         assert self.name is not None, "workflow name not defined"
 
-        # here we use the sleep interval to wait for the workflow post creation. This is to address a potential
```
elliotgunton, line 408: Restore file

src/hera/workflows/workflow_template.py
```diff
@@ -116,9 +116,8 @@ def from_dict(cls, model_dict: Dict) -> ModelMapperMixin:
         """Create a WorkflowTemplate from a WorkflowTemplate contained in a dict.
 
         Examples:
-            >>> my_workflow_template = WorkflowTemplate(name="my-wft")
-            >>> my_workflow_template == WorkflowTemplate.from_dict(my_workflow_template.to_dict())
-            True
+            my_workflow_template = WorkflowTemplate(...)
```
elliotgunton, line 119: Restore file

examples/workflows/script_annotations_artifacts_oss_new.py
elliotgunton: If someone's not using the runner then I think this would return an empty string? Not using the runner with Artifact annotations seems to be the wrong thing to be doing

examples/workflows/script_annotations.py
```diff
@@ -0,0 +1,36 @@
...
+
+global_config.experimental_features["script_annotations"] = True
+global_config.experimental_features["script_runner"] = True
+
+
+@script()
+def echo_all(
+    an_int: Annotated[int, Parameter(description="an_int parameter", default=1, enum=[1, 2, 3])],
+    a_bool: Annotated[bool, Parameter(description="a_bool parameter", default=True, enum=[True, False])],
+    a_string: Annotated[str, Parameter(description="a_string parameter", default="a", enum=["a", "b", "c"])],
```
elliotgunton, line 17: We can remove the enums here as we found they're only used in suspend templates.

elliotgunton (approved):
Wooo! 🚀

PR #734, opened 2023-08-02
Move extensive task docstring to getting-started.md

elliotgunton:
* Also fix "Examples:" sections breaking mkdocs rendering

**Pull Request Checklist**
- [x] Fixes bad doc rendering
- [ ] ~Tests added~

**Description of PR**
Currently, mkdocs is having trouble rendering the doc website

This PR fixes some of the docstrings escaping the rendering which results in sections like 
image
image


src/hera/workflows/task.py
```diff
@@ -49,91 +49,7 @@ class Task(
...
-
-    As a diagram:
-
-    ```
-    A   B
-    | X |
-    C   D
-    ```
-    """
+    """Task is used to run a given template within a DAG. Must be instantiated under a DAG context."""
```
dejamiko, line 52: Maybe it's worth linking back to the doc?

elliotgunton: I would rather keep the inline docstrings self-contained as the getting started guide structure may change/evolve so we don't want to have to update lots of docstrings in that case.

flaviuvadan (approved):
🎖️

elliotgunton:
image 
image


ToC now fixed all the way down 😎

Might just update links as they're not clickable which I think need to be markdown `[text](link)` style

image

samj1912 (approved):

PR #735, opened 2023-08-02
Add testable support for configmaps

### LINKED ISSUE #704: Add Arguments Parameters From Configmap for Scripts
There doesn't seem to be a way to use a parameter from configmap in a script now and have that Python function unit testable.

We'd need to create some sort of mocking system for allowing the unit testing or change the behaviour of the yaml generation (ignore subsequent usages of a parameter if it has a valueFrom?)

A Hera example of Arguments Parameters From Configmap within a container.
https://hera.readthedocs.io/en/stable/examples/workflows/upstream/arguments_parameters_from_configmap/
And the corresponding yaml
https://github.com/argoproj/argo-workflows/blob/master/examples/arguments-parameters-from-configmap.yaml

We would like to do something like:

```python
@script(inputs=[parameter(name="message", value_from=m.ValueFrom(
                config_map_key_ref=m.ConfigMapKeySelector(
                    name="simple-parameters",
                    key="msg",
                )
            ))])
def echo(message):
    print(message)
```

which behaves like a normal python function and generates a similar yaml

The current behaviour is that if we have the argument in the function signature, it will not take it from the configmap

```python
@script(inputs=[Parameter(name="a_string", value_from=ValueFrom( config_map_key_ref=ConfigMapKeySelector(
                    name="simple-parameters",
                    key="a_string_param"
                )))])
def echo_string(a_string):
    print(a_string)
```

```yaml
- inputs:
      parameters:
      - name: a_string
        valueFrom:
          configMapKeyRef:
            key: a_string_param
            name: simple-parameters
    name: echo-string
    script:
      command:
      - python
      image: python:3.8
      source: 'import os

        import sys

        sys.path.append(os.getcwd())

        import json

        try: a_string = json.loads(r''''''{{inputs.parameters.a_string}}'''''')

        except: a_string = r''''''{{inputs.parameters.a_string}}''''''


        print(a_string)'
```
 and if we remove that parameter we get the correct yaml

```python
@script(inputs=[Parameter(name="an_int", value_from=ValueFrom( config_map_key_ref=ConfigMapKeySelector(
                    name="simple-parameters",
                    key="an_int_param"
                )))])
def echo_int():
    print("{{inputs.parameters.an_int}}")
```

```yaml
  - inputs:
      parameters:
      - name: an_int
        valueFrom:
          configMapKeyRef:
            key: an_int_param
            name: simple-parameters
    name: echo-int
    script:
      command:
      - python
      image: python:3.8
      source: 'import os

        import sys

        sys.path.append(os.getcwd())

        print(''{{inputs.parameters.an_int}}'')'
```

but the function cannot be tested locally.

dejamiko:
**Pull Request Checklist**
- [x] Fixes #704 
- [x] Tests added

Using annotations allows us to support configmaps for parameters in inputs of scripts. This can be done by:

```python
@script()
def echo(
    message: Annotated[
        str,
        Parameter(
            name="message",
            value_from=m.ValueFrom(
                config_map_key_ref=m.ConfigMapKeySelector(
                    name="simple-parameters",
                    key="msg",
                )
            ),
        ),
    ]
):
    print(message)
```

This also adds `value_from` for input parameters which is something we should support.

### DIFF
--- a/docs/examples/workflows/script_auto_infer.md
+++ b/docs/examples/workflows/script_auto_infer.md
@@ -79,7 +79,7 @@
           command:
           - python
           image: python:3.8
-          source: "import os\nimport sys\nsys.path.append(os.getcwd())\nimport json\n\n\
-            import pickle\nwith open('/tmp/i', 'rb') as f:\n    i = pickle.load(f)\nprint(i)"
+          source: "import os\nimport sys\nsys.path.append(os.getcwd())\nimport pickle\n\
+            with open('/tmp/i', 'rb') as f:\n    i = pickle.load(f)\nprint(i)"
     ```
 
--- a/src/hera/workflows/script.py
+++ b/src/hera/workflows/script.py
@@ -457,9 +457,10 @@ class InlineScriptConstructor(ScriptConstructor):
             # non-JSON encoded strings are returned, which fail the loads, but they can be used as plain strings
             # which is why this captures that in an except. This is only used for `InputFrom` cases as the extra
             # payload of the script is not necessary when regular input is set on the task via `func_params`
-            extract += f"""try: {param.name} = json.loads(r'''{{{{inputs.parameters.{param.name}}}}}''')\n"""
-            extract += f"""except: {param.name} = r'''{{{{inputs.parameters.{param.name}}}}}'''\n"""
-        return textwrap.dedent(extract)
+            if param.value_from is None:
+                extract += f"""try: {param.name} = json.loads(r'''{{{{inputs.parameters.{param.name}}}}}''')\n"""
+                extract += f"""except: {param.name} = r'''{{{{inputs.parameters.{param.name}}}}}'''\n"""
+        return textwrap.dedent(extract) if extract != "import json\n" else ""
 
     def generate_source(self, instance: Script) -> str:
         """Assembles and returns a script representation of the given function.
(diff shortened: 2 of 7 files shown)


docs/examples/workflows/script_auto_infer.md
```diff
@@ -79,7 +79,7 @@
           command:
           - python
           image: python:3.8
-          source: "import os\nimport sys\nsys.path.append(os.getcwd())\nimport json\n\n\
-            import pickle\nwith open('/tmp/i', 'rb') as f:\n    i = pickle.load(f)\nprint(i)"
+          source: "import os\nimport sys\nsys.path.append(os.getcwd())\nimport pickle\n\
```
dejamiko, line 82: Notice this changes one unrelated yaml because of the changes to `source` generation. It this a problem?

src/hera/workflows/script.py
```diff
@@ -407,9 +407,11 @@ def _get_param_script_portion(self, instance: Script) -> str:
...
             # which is why this captures that in an except. This is only used for `InputFrom` cases as the extra
             # payload of the script is not necessary when regular input is set on the task via `func_params`
-            extract += f"""try: {param.name} = json.loads(r'''{{{{inputs.parameters.{param.name}}}}}''')\n"""
-            extract += f"""except: {param.name} = r'''{{{{inputs.parameters.{param.name}}}}}'''\n"""
-        return textwrap.dedent(extract)
+
+            if param.value_from is None:
+                extract += f"""try: {param.name} = json.loads(r'''{{{{inputs.parameters.{param.name}}}}}''')\n"""
+                extract += f"""except: {param.name} = r'''{{{{inputs.parameters.{param.name}}}}}'''\n"""
+        return textwrap.dedent(extract) if extract != "import json\n" else ""
```
dejamiko, line 414: I decided to not append the `import json\n` when we don't actually have any params. @elliotgunton suggested some users might use the implicit import though. Is this a problem?

flaviuvadan: I think it's ok to skip `json` here. Hera assumes it is imported for Hera specifically, not for the user. If users do need `json` they are likely to explicitly import it without awareness of the implicit one

flaviuvadan (approved):

Issue #652, opened 2023-05-29
Integrate Argo Workflows into Hera's CI


flaviuvadan:
**Is your feature request related to a problem? Please describe.**
Right now there's no standard way to test Hera workflow examples and users *need* their own clusters for testing. 

**Describe the solution you'd like**
Integrate minikube/k3s into CI and add a job that runs at least several workflows in CI to ensure submission works

**Describe alternatives you've considered**
Not sure there are any that are specifically simpler than integrating k8s into CI but definitely open to any!

**Additional context**
None

PR #741, opened 2023-08-08
Script annotations output

### LINKED ISSUE #740: Script outputs annotations
We want to be able to specify the outputs with annotations for both parameters and artifacts to simplify writing scripts with outputs.

Hera offers this for a simple hello world example:
```python
@script(outputs=Artifact(name="hello-art", path="/tmp/hello_world.txt"))
def whalesay():
   with open("/tmp/hello_world.txt", "w") as f:
       f.write("hello world")
```

The new approach would allow us to avoid duplication of the path and result in a much more readable code.
```python
@script()
def func() -> Annotated[str, Artifact(name="hello-art", path="/tmp/hello_world.txt")]:
    return "hello world"
```

The idea is to save the returned value in a file according to this schema:
`/hera/outputs/parameters/<name>`
`/hera/outputs/artifacts/<name>`
for easy access in subsequent Steps. The output is also exposed in the `outputs` section of the resulting yaml. 

## The interface
The item returned from the function can be of any serialisable pydantic type. That arbitrary type needs to be `Annotated` with an `Artifact` or `Parameter` which we refer to with the “`OutputItem`” alias here. One of these is needed because we have to know where to save it to file at the end. The `OutputItem`’s `name` will be taken and used for creating the path for saving the item. If the annotation is `Artifact` and it contains a `path`, we use that `path` to save the output.

```python
@script()
def func(...) -> Annotated[arbitrary_pydantic_type, OutputItem]:
    return output
```

We also allow the return type to be a `Tuple` of arbitrary pydantic types with `OutputItem` annotations.
```python
@script()
def func(...) -> Tuple[Annotated[arbitrary_pydantic_type_a, OutputItem1], Annotated[arbitrary_pydantic_type_b, OutputItem2], ...]:
    return output_a, output_b
```

We also want to allow `OutputItem`s as part of the function signature, allowing users to access/write to the output path as if it were a `Path` object. They will require an additional field `output=True` to distinguish them from the input parameters but will otherwise behave in the same way.

```python
@script()
def func(..., output_param: Annotated[Path, Parameter(output=True, global_name="...", name="")]) -> Annotated[arbitrary_pydantic_type, OutputItem]:
    output_param.write_text("...")
    return output
```

dejamiko:
**Pull Request Checklist**
- [x] Fixes #740 
- [x] Tests added

This PR implements the output annotation behaviour described in #740. The case with output params in the function signature as `Path` is not tested and this can be done with #652. 

```python
def func(..., output_param: Annotated[Path, Parameter(output=True, global_name="...", name="")]) -> Annotated[arbitrary_pydantic_type, OutputItem]:
    output_param.write_text("...")
    return output
```


src/hera/workflows/script.py
```diff
@@ -156,31 +157,66 @@ def _build_inputs(self) -> Optional[ModelInputs]:
...
+            return outputs
+
+        out_parameters, out_artifacts = _get_outputs_from_callable(self.source)
+        func_parameters, func_artifacts = _get_outputs_from_callable_signature(self.source)
+        func_parameters.extend(out_parameters)
+        func_artifacts.extend(out_artifacts)
+
+        return self._process_io(outputs, func_parameters, func_artifacts, True)
+
+    def _process_io(
```
flaviuvadan, line 178: What do you think abut moving this logic to `hera.workflows._mixins.IOMixin`? I think the IO can be constructed there and then the `Script` remains free of that logic / maintains only the `build`ers

src/hera/workflows/runner.py
```diff
@@ -122,7 +123,53 @@ def _map_keys(function: Callable, kwargs: dict) -> dict:
     return mapped_kwargs
 
 
-def _runner(entrypoint: str, kwargs_list: Any) -> str:
+def _save_outputs(function_results, output_destinations):
```
flaviuvadan, line 126: Any chance this can be documented a bit more? Reading it is relatively straight forward, but some documentation, and types, would be very useful for readability

src/hera/workflows/runner.py
```diff
@@ -122,7 +123,53 @@ def _map_keys(function: Callable, kwargs: dict) -> dict:
...
+
+def _write_to_path(path: Path, file) -> None:
+    directory = path.parent
+    directory.mkdir(parents=True, exist_ok=True)
+    path.touch(exist_ok=True)
+    result = serialize(file)
+    path.write_text(result if result else "")
+
+
+def _runner(entrypoint: str, kwargs_list: Any) -> Any:
```
flaviuvadan, line 172: Small - given the documentation `The result of the function as a string or `None` if the outputs are to be saved.` should the return type by `Optional[str]`?

src/hera/workflows/script.py
```diff
@@ -156,31 +157,66 @@ def _build_inputs(self) -> Optional[ModelInputs]:
             else:
                 func_parameters = _get_parameters_from_callable(self.source)
 
-        if inputs is None and not func_parameters and not func_artifacts:
+        return self._process_io(inputs, func_parameters, func_artifacts, False)
+
+    def _build_outputs(self) -> Optional[ModelOutputs]:
```
flaviuvadan, line 162: Same comment on this as [here](https://github.com/argoproj-labs/hera/pull/741/files#r1287511036)

tests/script_annotations_outputs/script_annotations_output_param_and_artifact_in_func.py
```diff
@@ -0,0 +1,24 @@
...
+
+from hera.workflows import Artifact, Parameter, Workflow, script
+from hera.workflows.steps import Steps
+
+
+@script()
+def script_param_in_fun_sig(
+    a_number: Annotated[int, Parameter(name="a_number")],
+    successor: Annotated[Path, Parameter(name="successor", output=True)],
+    successor2: Annotated[Path, Artifact(name="successor2", output=True)],
```
flaviuvadan, line 16: From a UX perspective it might be beneficial to have the `Parameter.name` field, and `Artifact.name`, inferred based on the arg. Is that possible? Even if it means setting an arbitrary name on init then changing it, perhaps?

flaviuvadan:
```python
    a_number: Annotated[int, Parameter()],
    successor: Annotated[Path, Parameter(output=True)],
    successor2: Annotated[Path, Artifact(output=True)],
```
looks a bit nicer!

tests/script_annotations_outputs/script_annotations_output_param_and_artifact_mix.py
```diff
@@ -0,0 +1,29 @@
+try:
```
flaviuvadan, line 1: For my own understanding - why are these tests vs examples? What are these testing?

flaviuvadan: Oh, nvm, I looked up the name of the files and these files are actually used by the table style tests that follow! Maybe a small note at the top of each module might be helpful to clarify how these are tested?

flaviuvadan (commented):
🎖️ looks fantastic! I left a few questions / small suggestions

src/hera/workflows/runner.py
elliotgunton: Mikolaj actually pointed out the doc string seemed wrong but I guess he forgot to update it 😆 The return type should be `Any` because we're returning the value of a user function and only serializing it in `_run`. Not sure you gain much from `Optional[Any]` as it's technically covered by just `Any` but it's an option (pun intended 😎)

dejamiko: Indeed, the return type used to be `str` which I fixed but I forgot the docstring

src/hera/workflows/runner.py
dejamiko: Sure thing!

src/hera/workflows/script.py
dejamiko: I believe this logic is unique for scripts as it deals with extracting information from the function signature. If we ever want to reuse that we could possibly create a `FunctionIOMixin` containing that logic. What do you think?

tests/script_annotations_outputs/script_annotations_output_param_and_artifact_mix.py
dejamiko: I added a line to all the test files to explain what the purpose is

src/hera/workflows/_mixins.py
```diff
@@ -575,20 +575,23 @@ def _build_arguments(self) -> Optional[ModelArguments]:
...
                     [arg._build_artifact()] if result.artifacts is None else result.artifacts + [arg._build_artifact()]
                 )
-            elif isinstance(arg, ModelParameter):
-                result.parameters = [arg] if result.parameters is None else result.parameters + [arg]
             elif isinstance(arg, Parameter):
                 result.parameters = (
                     [arg.as_argument()] if result.parameters is None else result.parameters + [arg.as_argument()]
                 )
-
+            elif isinstance(arg, ModelParameter):
```
elliotgunton, line 593: Does `ModelArtifact` also need moving after `Artifact`?

src/hera/workflows/runner.py
```diff
@@ -122,7 +123,66 @@ def _map_keys(function: Callable, kwargs: dict) -> dict:
...
+                path = Path(dest[1].path)
+            else:
+                path = output_dir / f"artifacts/{cast(str, dest[1].name)}"
+        _write_to_path(path, res)
+
+
+def _extract_output_annotations(function: Callable) -> Optional[List]:
+    output = []
+    if get_origin(inspect.signature(function).return_annotation) is Annotated:
+        output.append(get_args(inspect.signature(function).return_annotation))
```
elliotgunton, line 169: Can we make variables like

`origin_type = get_origin(inspect.signature(function).return_annotation)`

and

`annotation_args = get_args(inspect.signature(function).return_annotation)`

to make this a bit easier to read?

src/hera/workflows/script.py
```diff
@@ -156,31 +157,66 @@ def _build_inputs(self) -> Optional[ModelInputs]:
...
+
+        return self._process_io(outputs, func_parameters, func_artifacts, True)
+
+    def _process_io(
+        self,
+        current_io: Optional[Union[ModelInputs, ModelOutputs]],
+        func_parameters: List[Parameter],
+        func_artifacts: List[Artifact],
+        output: bool,
+    ):
```
elliotgunton, line 184: Needs return type, `Optional[ModelOutputs]`?

src/hera/workflows/script.py
```diff
@@ -282,17 +318,72 @@ def _get_parameters_from_callable(source: Callable) -> List[Parameter]:
...
+    parameters = []
+    artifacts = []
+
+    for name, p in inspect.signature(source).parameters.items():
+        if get_origin(p.annotation) is not Annotated:
+            continue
+        annotation = get_args(p.annotation)[1]
+        if not hasattr(annotation, "output") or not annotation.output:
+            continue
+        mytype = type(annotation)
```
elliotgunton, line 357: Tiny name suggestion and then adding an assert helps tell us what that type should be and that calling `get_input_attributes` on it will be legal (side note, did we make that an abstract method or protocol somewhere?)

```suggestion
        output_type = type(annotation)
        assert isintance(output_type, (Artifact, Parameter))
```

src/hera/workflows/runner.py
```diff
@@ -122,7 +123,66 @@ def _map_keys(function: Callable, kwargs: dict) -> dict:
...
+    if not isinstance(function_results, tuple):
+        function_results = [function_results]
+    if len(function_results) != len(output_destinations):
+        raise ValueError("The number of outputs does not match the annotation")
+
+    output_dir = Path(
+        str(global_config.experimental_features["hera_output_directory"])
+        if global_config.experimental_features["hera_output_directory"]
+        else "hera/outputs"
+    )
```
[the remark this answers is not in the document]
flaviuvadan: Could we make another field on the global config vs use the experimental features + use the experimental features to check for allowed usage?

src/hera/workflows/script.py
flaviuvadan: Sounds good to me 👍🏻

src/hera/workflows/_mixins.py
dejamiko: Interestingly, no. That's because `Artifact` inherits from `BaseModel` not `ModelArtifact`

src/hera/workflows/runner.py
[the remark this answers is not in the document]
samj1912: I think we should use https://github.com/argoproj-labs/hera/blob/49140d07990cca8f2a22b72f1cf08aea8c225ca5/src/hera/shared/_global_config.py#L118 + an attribute in the script runner class.

samj1912:
@dejamiko you don't need to call _get_class_defaults, the hera mixin automatically handles it. See https://github.com/argoproj-labs/hera/blob/49140d07990cca8f2a22b72f1cf08aea8c225ca5/src/hera/shared/_global_config.py#L157

src/hera/workflows/runner.py
```diff
@@ -122,7 +123,62 @@ def _map_keys(function: Callable, kwargs: dict) -> dict:
...
+    The results are matched with the specified outputs and saved using the schema:
+
+    If the artifact path is specified, that is used instead.
+    """
+    if not isinstance(function_results, tuple):
+        function_results = [function_results]
+    if len(function_results) != len(output_destinations):
+        raise ValueError("The number of outputs does not match the annotation")
+
+    outputs_dir = Path(RunnerScriptConstructor().outputs_directory)
```
elliotgunton, line 140: I'm not sure this will work in the context of the script runner in a user's image - they would have to call `set_class_defaults` at some point in their code (as in, within the executing function as opposed to in a config somewhere), otherwise they'll get the default `hera/outputs`

tests/test_runner.py
```diff
@@ -36,7 +41,142 @@
...
+            "tests.script_annotations_outputs.script_annotations_output:script_two_artifacts",
+            [{"name": "a_number", "value": "3"}],
+            [
+                {"path": "hera/outputs/artifacts/successor", "value": "4"},
+                {"path": "hera/outputs/artifacts/successor2", "value": "5"},
+            ],
+        ),
+    ],
+)
+# @patch("hera.workflows.runner.importlib")
```
elliotgunton, line 99: Do we not need to mock the importlib?

dejamiko: Not anymore, since we're not using different `tmp_path`s for different tests

examples/workflows/script-annotations-outputs.yaml
```diff
@@ -0,0 +1,39 @@
...
+  templates:
+  - name: my_steps
+    steps:
+    - - arguments:
+          parameters:
+          - name: a_number
+            value: '3'
+        name: script-param-artifact-in-function-signature-and-return-type
+        template: script-param-artifact-in-function-signature-and-return-type
+  - inputs:
```
elliotgunton, line 16: This template doesn't mount volumes for the outputs to go in

examples/workflows/script-annotations-outputs.yaml
```diff
@@ -0,0 +1,47 @@
...
+      args:
+      - -m
+      - hera.workflows.runner
+      - -e
+      - examples.workflows.script_annotations_outputs:script_param_artifact_in_function_signature_and_return_type
+      command:
+      - python
+      env:
+      - name: hera__script_annotations
+        value: ''
```
dejamiko, line 37: Maybe for the yaml readability we should put some meaningful value here? Like maybe `true`

elliotgunton: The problem is there isn't a meaningful value here (because we're going by existence of the envvar only), and adding something like `true` suggests `false` is a valid alternative but it's not.

examples/workflows/script-annotations-outputs.yaml
```diff
@@ -0,0 +1,47 @@
...
+      - name: hera__outputs_directory
+        value: user/chosen/outputs
+      image: python:3.8
+      source: '{{inputs.parameters}}'
+      volumeMounts:
+      - mountPath: user/chosen/outputs
+        name: hera__outputs_directory
+    volumes:
+    - emptyDir: {}
+      name: hera__outputs_directory
```
elliotgunton, line 47: LGTM on the argo spec side - have you tried building an image with the function to try it out?

src/hera/workflows/artifact.py
```diff
@@ -74,6 +74,13 @@ class Artifact(BaseModel):
     loader: Optional[ArtifactLoader] = None
     """used in Artifact annotations for determining how to load the data"""
 
+    output: bool = False
+    """used in output annotations in function signatures"""
```
elliotgunton, line 78:
```suggestion
    """used to specify artifact as an output in function signature annotations"""
```

src/hera/workflows/parameter.py
```diff
@@ -22,6 +22,9 @@ class Parameter(_ModelParameter):
 
     name: Optional[str] = None  # type: ignore
 
+    output: Optional[bool] = False
+    """used in output annotations in function signatures"""
```
elliotgunton, line 26:
```suggestion
    """used to specify parameter as an output in function signature annotations"""
```

src/hera/workflows/runner.py
```diff
@@ -122,7 +123,57 @@ def _map_keys(function: Callable, kwargs: dict) -> dict:
...
+
+        path = _get_outputs_path(dest[1])
+        _write_to_path(path, res)
+
+
+def _get_outputs_path(destination: Union[Parameter, Artifact]) -> Path:
+    """Get the path from the destination annotation using the defined outputs directory."""
+    outputs_dir = Path(os.environ.get("hera__outputs_directory", "hera/outputs"))
+    if isinstance(destination, Parameter):
+        path = outputs_dir / f"parameters/{cast(str, destination.name)}"
```
elliotgunton, line 158: I guess the `cast` is because of linting issues because it's `Optional[str]`? I would just move the `if not dest[1].name:` to the top of this function, then the linter should be able to figure it out

src/hera/workflows/runner.py
```diff
@@ -122,7 +123,57 @@ def _map_keys(function: Callable, kwargs: dict) -> dict:
...
+            path = Path(destination.path)
+        else:
+            path = outputs_dir / f"artifacts/{cast(str, destination.name)}"
+    return path
+
+
+def _write_to_path(path: Path, file) -> None:
+    """Write the file contents to the provided path. Create the necessary parent directories."""
+    directory = path.parent
+    directory.mkdir(parents=True, exist_ok=True)
```
elliotgunton, line 170: Nitpicking, but because `directory` isn't used again
```suggestion
    path.parent.mkdir(parents=True, exist_ok=True)
```

src/hera/workflows/runner.py
```diff
@@ -122,7 +123,57 @@ def _map_keys(function: Callable, kwargs: dict) -> dict:
...
+        path = outputs_dir / f"parameters/{cast(str, destination.name)}"
+    elif isinstance(destination, Artifact):
+        if destination.path:
+            path = Path(destination.path)
+        else:
+            path = outputs_dir / f"artifacts/{cast(str, destination.name)}"
+    return path
+
+
+def _write_to_path(path: Path, file) -> None:
```
elliotgunton, line 167: Missing type on `file`

src/hera/workflows/script.py
```diff
@@ -146,44 +151,11 @@ def _constructor_validate(cls, values):
...
-                    inputs.artifacts = []
-                inputs.artifacts.append(artifact._build_artifact())
-
-        return inputs
-
     def _build_template(self) -> _ModelTemplate:
         assert isinstance(self.constructor, ScriptConstructor)
+        hera_outputs_used = self._check_hera_outputs_used()
+        if hera_outputs_used:
+            self._create_new_volume()
```
elliotgunton, line 158: These could be called within `_build_script` as long as we call it first here, i.e.
```suggestion
        script=self._build_script()
```

Then in `_build_script` check/call the `self._create_new_volume` before `env = self._build_env()`

src/hera/workflows/script.py
```diff
@@ -233,6 +205,8 @@ def _build_script(self) -> _ModelScriptTemplate:
             if not env:
                 env = []
             env.append(EnvVar(name="hera__script_annotations", value=""))
+            if hera_outputs_used:
```
elliotgunton, line 208: Seeing as we're doing `isinstance(self.constructor, RunnerScriptConstructor)`, I think it makes sense for this code (L201 to L209) to be in the RunnerScriptConstructor's `transform_script_template_post_build`? (i.e. we need to write and override the function as it doesn't exist in the subclass atm)

examples/workflows/script_annotations_outputs.py
```diff
@@ -0,0 +1,36 @@
...
+global_config.set_class_defaults(RunnerScriptConstructor, outputs_directory="user/chosen/outputs")
+
+
+@script(constructor="runner")
+def script_param_artifact_in_function_signature_and_return_type(
+    a_number: Annotated[int, Parameter(name="a_number")],
+    successor: Annotated[Path, Parameter(name="successor", output=True)],
+    successor2: Annotated[Path, Artifact(name="successor2", output=True)],
+) -> Tuple[
+    Annotated[int, Parameter(name="successor3", output=True)], Annotated[int, Artifact(name="successor4", output=True)]
```
elliotgunton, line 27: Why is `output=True` needed here when it's in the return annotation?

docs/getting-started/walk-through/advanced-hera-features.md
```diff
@@ -232,3 +234,62 @@ This feature can be enabled by setting the `experimental_feature` flag `script_a
...
+def func() -> Annotated[str, Artifact(name="hello-art", path="/tmp/hello_world.txt")]:
+    return "hello world"
+```
+
+The idea is to save the returned value in a file according to this schema:
+`/hera/outputs/parameters/<name>`
+`/hera/outputs/artifacts/<name>`
+for easy access in subsequent Steps. The output is also exposed in the `outputs` section of the resulting yaml. 
+
+The item returned from the function can be of any serialisable pydantic type. That arbitrary type needs to be
```
elliotgunton, line 261:
```suggestion
The item returned from the function can be of any serialisable Pydantic type. That arbitrary type needs to be 
```

docs/getting-started/walk-through/advanced-hera-features.md
```diff
@@ -232,3 +234,62 @@ This feature can be enabled by setting the `experimental_feature` flag `script_a
...
+    return "hello world"
+```
+
+The idea is to save the returned value in a file according to this schema:
+`/hera/outputs/parameters/<name>`
+`/hera/outputs/artifacts/<name>`
+for easy access in subsequent Steps. The output is also exposed in the `outputs` section of the resulting yaml. 
+
+The item returned from the function can be of any serialisable pydantic type. That arbitrary type needs to be 
+`Annotated` with an `Artifact` or `Parameter` which we refer to with the “`OutputItem`” alias here. One of
```
elliotgunton, line 262:
```suggestion
`Annotated` with an `Artifact` or `Parameter` which we refer to with the `OutputItem` alias here. One of 
```

docs/getting-started/walk-through/advanced-hera-features.md
```diff
@@ -232,3 +234,62 @@ This feature can be enabled by setting the `experimental_feature` flag `script_a
...
+```
+
+The idea is to save the returned value in a file according to this schema:
+`/hera/outputs/parameters/<name>`
+`/hera/outputs/artifacts/<name>`
+for easy access in subsequent Steps. The output is also exposed in the `outputs` section of the resulting yaml. 
+
+The item returned from the function can be of any serialisable pydantic type. That arbitrary type needs to be 
+`Annotated` with an `Artifact` or `Parameter` which we refer to with the “`OutputItem`” alias here. One of 
+these is needed because we have to know where to save it to file at the end. The `OutputItem`’s `name` will
```
elliotgunton, line 263: Special character
```suggestion
these is needed because we have to know where to save it to file at the end. The `OutputItem`'s `name` will 
```

docs/getting-started/walk-through/advanced-hera-features.md
```diff
@@ -232,3 +234,62 @@ This feature can be enabled by setting the `experimental_feature` flag `script_a
...
+be taken and used for creating the path for saving the item. If the annotation is `Artifact` and it contains a `path`, 
+we use that `path` to save the output.
+
+```python
+@script()
+def func(...) -> Annotated[arbitrary_pydantic_type, OutputItem]:
+    return output
+```
+
+We also allow the return type to be a `Tuple` of arbitrary pydantic types with `OutputItem` annotations.
```
elliotgunton, line 273:
```suggestion
We also allow the return type to be a `Tuple` of arbitrary Pydantic types with `OutputItem` annotations.
```

docs/getting-started/walk-through/advanced-hera-features.md
elliotgunton: Actually, reading it through I don't think the alias is helpful here, especially for `Tuple[Annotated[arbitrary_pydantic_type_a, OutputItem1], Annotated[arbitrary_pydantic_type_b, OutputItem2], ...]` - it starts to confuse what is the arbitrary type. Plus it's new terminology only used to introduce this feature. Can we replace "`OutputItem`" with "`Artifact`/`Parameter`" in the text, and choose one of the types for the examples (or show with both)?

docs/getting-started/walk-through/advanced-hera-features.md
```diff
@@ -232,3 +234,62 @@ This feature can be enabled by setting the `experimental_feature` flag `script_a
...
+def func(..., output_param: Annotated[Path, Parameter(output=True, global_name="...", name="")]) -> Annotated[arbitrary_pydantic_type, OutputItem]:
+    output_param.write_text("...")
+    return output
+```
+
+The parent outputs directory `/hera/outputs` can be overwritten by the user. This is done by adding 
+```python
+global_config.set_class_defaults(RunnerScriptConstructor, outputs_directory="user/chosen/outputs")
+```
+in the script using the outputs. Notice, this is only done for scripts using the runner constructor.
```
elliotgunton, line 295:
```suggestion
in the script using the outputs. Note, this is only done for scripts using the runner constructor.
```

src/hera/workflows/script.py
```diff
@@ -264,6 +238,103 @@ def _build_script(self) -> _ModelScriptTemplate:
...
+
+        if new_volume not in self.volumes:
+            self.volumes.append(new_volume)
+
+    def _get_outputs_directory(self) -> str:
+        """Get the outputs directory from the constructor, provide a default if not set."""
+        if isinstance(self.constructor, RunnerScriptConstructor):
+            if self.constructor.outputs_directory is not None:
+                return self.constructor.outputs_directory
+        return "/hera/outputs"
```
elliotgunton, line 336: This path has a leading `/` whereas in the Runner Constructor it's `outputs_directory: str = "hera/outputs"`. Could we confirm which is right? (I think with the leading?) Also, I don't think you can call this function without the constructor being a runner constructor, so you can change the `if isinstance` to `assert isinstance` to make sure linting passes, and also the outputs directory should not be able to be None, so I don't think `if self.constructor.outputs_directory is not None:` is necessary (just assert and return)

src/hera/workflows/script.py
```diff
@@ -264,6 +231,90 @@ def _build_script(self) -> _ModelScriptTemplate:
...
+                    artifacts=[a._build_artifact() for a in func_artifacts] or None,
+                )
+            else:
+                current_io = ModelInputs(
+                    parameters=[p.as_input() for p in func_parameters] or None,
+                    artifacts=[a._build_artifact() for a in func_artifacts] or None,
+                )
+
+        already_set_params = {p.name for p in current_io.parameters or []}
+        already_set_artifacts = {a.name for a in current_io.artifacts or []}
```
elliotgunton, line 286: `seen_params`/`seen_artifacts` might be a bit more conventional?

src/hera/workflows/script.py
```diff
@@ -330,6 +446,13 @@ def _get_parameters_and_artifacts_from_callable(
     return parameters, artifacts
 
 
+def _check_hera_outputs_used(source: Callable) -> bool:
+    """Check if hera outputs are used. This is needed to know if we need to create a volume."""
```
elliotgunton, line 450: This is only checking if output annotations are used - can we rename? Current name sounds like whether the default `hera/outputs` directory is used. Also don't need to say what this function is used for as that can change over time. Also, might be better to put the experimental features check here rather than at the call site in the other comment

```suggestion
def _output_annotations_used(source: Callable) -> bool:
    """Check if hera output annotations are used."""
```

docs/getting-started/walk-through/advanced-hera-features.md
```diff
@@ -232,3 +234,62 @@ This feature can be enabled by setting the `experimental_feature` flag `script_a
...
+```python
+@script()
+def func() -> Annotated[str, Artifact(name="hello-art", path="/tmp/hello_world.txt")]:
+    return "hello world"
+```
+
+The idea is to save the returned value in a file according to this schema:
+`/hera/outputs/parameters/<name>`
+`/hera/outputs/artifacts/<name>`
+for easy access in subsequent `Steps`/`Tasks``. The output is also exposed in the `outputs` section of the resulting yaml.
```
elliotgunton, line 259:
```suggestion
for easy access in subsequent `Steps`/`Tasks`. The output is also exposed in the `outputs` section of the resulting yaml. 
```

src/hera/workflows/protocol.py
```diff
@@ -67,3 +67,13 @@ class Steppable(Protocol):
...
     def _build_step(self) -> Any:
         ...
+
+
+@runtime_checkable
+class Inputable(Protocol):
+    """This runtime protocol indicates that an object supports being used as an input annotation."""
+
+    @classmethod
+    def get_input_attributes(cls) -> List[str]:
```
elliotgunton, line 77: Also this I think should be private

PR #755, opened 2023-08-31
Fix output Artifact/Parameter annotations

elliotgunton:
**Pull Request Checklist**
- [x] Fixes bugs in #741
- [x] Tests added

**Description of PR**
Currently, output Artifacts/Parameters contained in function return annotations or function parameter annotations where marked as output don't output their paths in the yaml. The runner also does not pre-load function input parameters marked as outputs as Path objects for the given kwarg.

This PR fixes output annotations for these use cases and makes the tests more robust. The "advanced-hera-features" doc was also edited.

I also fixed the codebase for use with VSCode's pytest debugger.

### DIFF
--- a/docs/getting-started/walk-through/advanced-hera-features.md
+++ b/docs/getting-started/walk-through/advanced-hera-features.md
@@ -177,7 +177,7 @@ def read_artifact():
         print(a_file.read())
 ```
 
-By using annotations we can avoid repeating the `path` of the file, and your function can use the variable directly as a
+By using annotations we can avoid repeating the `path` of the file, and the function can use the variable directly as a
 `Path` object, with its value already set to the given path:
 
 ```python
@@ -302,7 +302,7 @@ def func(...) -> Annotated[arbitrary_pydantic_type, Artifact(name="my-output")]:
 ```
 
 For multiple outputs, the return type should be a `Tuple` of arbitrary Pydantic types with individual
-`Parameter`/`Artifact` annotations, and you should return a tuple from the function matching these types:
+`Parameter`/`Artifact` annotations, and the function must return a tuple from the function matching these types:
 ```python
 @script()
 def func(...) -> Tuple[
@@ -314,7 +314,7 @@ def func(...) -> Tuple[
 ```
 
 Hera also allows output `Parameter`/`Artifact`s as part of the function signature when specified as a `Path` type,
-allowing users to write to the output path, without needing an explicit return. They require an additional field
+allowing users to write to the path as an output, without needing an explicit return. They require an additional field
 `output=True` to distinguish them from the input parameters and must have an underlying `Path` type (or another type
 that will write to disk). For `Parameters`, they will have the path of `/hera/outputs/parameters/<name>`, which can be
 changed by setting the parameter's `value_from` argument. For `Artifacts`, they will have a path of
--- a/docs/getting-started/walk-through/advanced-hera-features.md
+++ b/docs/getting-started/walk-through/advanced-hera-features.md
@@ -98,9 +98,9 @@ dictionary before using the feature:
 global_config.experimental_features["NAME_OF_FEATURE"] = True
 ```
 
-### Currently supported experimental features:
+## Currently supported experimental features:
 
-#### `RunnerScriptConstructor`
+### `RunnerScriptConstructor`
 The `RunnerScriptConstructor` found in `hera.workflows.script` and seen in the
 [callable script example](../../examples/workflows/callable_script.md) is a robust way to run Python functions on Argo.
 The image used by the script should be built from the source code package itself and its dependencies, so that the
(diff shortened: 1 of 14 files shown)


docs/getting-started/walk-through/advanced-hera-features.md
```diff
@@ -113,22 +113,23 @@ global_config.experimental_features["script_runner"] = True
...
 
-##### Parameters
+```py
+global_config.experimental_features["script_annotations"] = True
+```
 
-In Hera, we can specify inputs inside the `@script` decorator as follows:
+#### Parameters
+
+In Hera, we can currently specify inputs inside the `@script` decorator as follows:
```
dejamiko, line 132: I think "currently" is not the best word here, since both the "old" and the "new" are currently supported

docs/getting-started/walk-through/advanced-hera-features.md
```diff
@@ -160,85 +162,104 @@ def echo_all(
...
 The fields allowed in the `Parameter` annotations are: `name`, `default`, `enum`, and `description`.
 
-##### Artifacts
+#### Artifacts
+
+> Note: Artifact annotations are only supported when used with the `RunnerScriptConstructor`.
 
-The improvement is even more noticeable for `Artifact`s. In Hera we are able to specify `Artifact`s in `inputs`:
+
+The feature is even more powerful for `Artifact`s. In Hera we are currently able to specify `Artifact`s in `inputs`, but
```
dejamiko, line 170: As above, not sure if the "currently" is needed

docs/getting-started/walk-through/advanced-hera-features.md
```diff
@@ -160,85 +162,104 @@ def echo_all(
...
+have no way to programmatically link the given path to the code within the function:
```
dejamiko, line 171: We have no way as in we Hera developers? Maybe "the given path is not programmatically linked to the code" would be better?

docs/getting-started/walk-through/advanced-hera-features.md
```diff
@@ -160,85 +162,104 @@ def echo_all(
...
 ```python
 @script(inputs=Artifact(name="my-artifact", path="/tmp/file"))
 def read_artifact():
-    with open("/tmp/file") as a_file:
+    with open("/tmp/file") as a_file:  # Repeating "/tmp/file" is prone to human error!
         print(a_file.read())
 ```
 
-But by using annotations we can avoid repeating the `path` of the file, and access the artifact as if it's a regular Python argument of the given type:
+By using annotations we can avoid repeating the `path` of the file, and your function can use the variable directly as a
```
dejamiko, line 180: Maybe "the function" instead of "your function"

docs/getting-started/walk-through/advanced-hera-features.md
```diff
@@ -160,85 +162,104 @@ def echo_all(
...
-```
+> Note: Output annotations are only supported when used with the `RunnerScriptConstructor`.
 
-##### Outputs
-The annotations can also be used for outputs. An example can be seen [here](../../examples/workflows/script_annoations_outputs.md)
+Annotations can also be used for outputs. An example can be seen
+[here](../../examples/workflows/script_annotations_outputs).
 
-Hera offers this for a simple hello world example:
+For a simple hello world output artifact example we currently have:
```
dejamiko, line 262: As above, not sure if "currently" is a good word here

docs/getting-started/walk-through/advanced-hera-features.md
```diff
@@ -254,32 +275,50 @@ def func() -> Annotated[str, Artifact(name="hello-art", path="/tmp/hello_world.t
...
+
+```python
+@script()
+def func(...) -> Annotated[arbitrary_pydantic_type, Artifact(name="my-output")]:
+    return output  # will be saved to /hera/outputs/artifacts/my-output
 ```
 
-We also allow the return type to be a `Tuple` of arbitrary Pydantic types with `Parameter`/`Artifact` annotations.
+For multiple outputs, the return type should be a `Tuple` of arbitrary Pydantic types with individual
+`Parameter`/`Artifact` annotations, and you should return a tuple from the function matching these types:
```
dejamiko, line 305: Not sure if the direct "you should return" is congruent with the rest of the text. Maybe "and the function should return a tuple matching these types" would be better?

docs/getting-started/walk-through/advanced-hera-features.md
elliotgunton: "currently supported" as in, "doesn't require you to switch on the experimental flag" - not sure there's a better, more succinct word

dejamiko: Ah, I see. Maybe just drop it altogether? 
"In Hera, we can specify inputs inside the `@script` decorator as follows:"

dejamiko: That is a good point, it makes it easier to contrast them

samj1912 (approved):

PR #760, opened 2023-09-01
Examples roundtrip test

elliotgunton:
**Pull Request Checklist**
- [x] Another step towards #567 - we still need the actual Python examples
- [x] Tests added (but some failing)

**Description of PR**
Adds a fetcher for all upstream files, then performs a `Hera_class.from_yaml(file).to_yaml()` round trip to confirm the spec is the same.

Also fixed borked workflows-examples.md page

Also removed unused `Inputable` protocol that I thought I removed in #755 but I guess I forgot to commit


elliotgunton:
Confirmed table no longer borked 
image

flaviuvadan (approved):
This LGTM! One Q I have is whether it's possible that examples across different tagged version of AW could actually break on Hera's side because of a version mismatch? i.e. we generate models on 3.4.4 but there can be some example from 3.5 that might break on Hera's side

samj1912 (approved):

elliotgunton:
@flaviuvadan I think on Hera's side with the old models, it should ignore the extra fields when taking in the yaml in a `from_yaml` call, and I don't think new models will break old yamls - only if there is a non-optional field added in a new version which would be a backwards incompatible change so I'd hope that wouldn't happen in a minor bump!

PR #763, opened 2023-09-04
Callable syntax

### LINKED ISSUE #753: Callable syntax support
## Status quo

Currently, passing arguments to scripts is rather inconvenient. The general idea is passing a list to an `arguments` keyword argument. There is a simplified “dictionary” style for passing in parameters without instantiating the `Parameter` objects:
```python
@script()
def echo_all(a_number, a_bool, a_string):
    print(a_number)
    print(a_bool)
    print(a_string)


with Workflow(generate_name="example-callable-", entrypoint="my_steps") as w:
    with Steps(name="my_steps") as s:
        echo_all(arguments={"a_number": 3, "a_bool": True, "a_string": "abc"})

        # Which is a simplified syntax of:
        Step(name="echo_all", source=echo_all, arguments={...})
```

The same cannot be said about `Artifacts` though. They necessitate the “list” syntax which requires instantiating all the objects on the way.

```python
@script()
def echo_all(an_int, a_bool, a_string, an_artifact: Annotated[str, Artifact(name="my-artifact", path="tmp/file", loader=ArtifactLoader.file)]):
    print(an_int)
    print(a_bool)
    print(a_string)
    print(an_artifact)


with Workflow(generate_name="example-callable-", entrypoint="my_steps") as w:
    with Steps(name="my_steps") as s:
        echo_all(
            arguments=[
                Parameter(name="an_int", value=1),
                Parameter(name="a_bool", value=True),
                Parameter(name="a_string", value="abc"),
                Artifact(name="my-artifact", from_="somewhere"),
            ]
        )
```
In either case, the way arguments are passed is through the arguments field of the script. What would be the most Pythonic way of writing this which would emphasise the one-to-one of a Python function and a Hera template?

## Design
It is relatively easy to implement a similar solution that allows for args and kwargs (#647). There is however a problem with some corner cases. They are rather convoluted, but for instance for:
```python
@script()
def person(age=42, name="Joe"):
   print(age, name)


with Workflow(generate_name="script-default-params-", entrypoint="d") as w:
   with DAG(name="d"):
       person(name="Jane", name="my-argo-job") # <- we have a clash with name but we can't use positional syntax because we want to use age's default value
```

It’s not possible to use keyword arguments because of the clash with the Argo params. Positional arguments cannot be used in this case, since we want to use the default value of `age`. The user would have to resort to the old syntax:

```python
@script()
def person(age=42, name="Joe"):
   print(age, name)


with Workflow(generate_name="script-default-params-", entrypoint="d") as w:
   with DAG(name="d"):
       person(arguments={"name": "Jane"}, name="my-argo-job") # <- we have a clash with name but we can't use positional syntax because we want to use age's default value
```

One solution is to first pass in all the arguments to the script function which returns a Task/Step, depending on the context. Then, a second method call updates the object to contain the Task/Step specific attributes.

```python
@script()
def person(age=42, name="Joe"):
   print(age, name)


with Workflow(generate_name="script-default-params-", entrypoint="d") as w:
   with DAG(name="d"):
       person(name="Jane").with_(name="my-argo-job")
```

The `with_` method could be implemented as follows

```python
class Task:
    ...
    def with_atrributes(self, name: str, ...) -> Task:
        self.name = name
        ...
        return self 
```

Something similar to this can be used, possibly within a mixin to share that behaviour between `Task` and `Step`. 

I believe this approach is beneficial because it allows for passing in the arguments in a natural Pythonic way to the function and then setting the `Task`/`Step` attributes in one simple step. It allows for passing in a dictionary of arguments.

## How can we implement this alongside the old syntax?
We introduce this natural arguments passing as an experimental feature. To enable it, the user must use a kwarg in the script decorator 
```python
@script(directly_callable=True)
def echo_num(number: int = 42):
   print(number)


with Workflow(generate_name="script-default-params-", entrypoint="d") as w:
   with DAG(name="d"):
       echo_num().with_(name="my-echo", with_params=[1, 2, 3])
```
Then, in `TemplateInvocatorSubNodeMixin`, the `__call__` function can be updated and include a check for `self.directly_callable`. This can also be set in e.g. `Containers` to allow for callable syntax for those. `directly_callable` could maybe be an attribute of the `TemplateInvocatorSubNodeMixin` to allow it to be used in any of the classes that use it. With this check done, we can take the args and kwargs and assume all of them are actually function arguments. We can populate `kwargs["arguments"]` with those and return depending on the context. We can most likely do without introducing new classes for those uses, since the only difference is introducing one method.

This approach allows the user to decide on a granular level if they want to use the old or the new syntax. Calls to scripts with old syntax can coexist with the new syntax within the same workflow and file. The decision can be made for each callable. 

If the community does not like this feature, it is easy to disable/remove. The previous way of passing arguments is continually supported. 

If the community likes the feature, we have a choice of how to handle that. 

1. We can try to support both syntaxes forever. To make it easier to use the new syntax, we can move the `directly_callable` setting to `global_config` which reduces repetition. In this scenario, there is no sudden breaking point and the users can migrate if they want to or use a mix of both features. The new one will get new features but both will be maintained. 
2. We can also decide that breaking changes are a part of life and flip the switch with one of major releases. We would change the default from `directly_callable=False` to `directly_callable=True` which would break code using the previous syntax. We can still support the old syntax with manual `directly_callable=False` if the users would like that.

## Examples
This syntax for some currently existing examples looks as follows:

```python


@script(directly_callable=True)
def foo(a, b=42, c=None):
    print(a, b, c)


with Workflow(generate_name="script-default-params-", entrypoint="d") as w:
    with DAG(name="d"):
        foo(1).with_(name="b-unset-c-unset")
        foo(1, 2).with_(name="b-set-c-unset")
        foo(1, c=2).with_(name="b-unset-c-set")
        foo(1, 2, 3).with_(name="b-set-c-set")
```

```python
from hera.workflows import Container, Steps, Workflow

with Workflow(generate_name="container-on-exit-", entrypoint="step-template") as w:
    exit_container = Container(
        name="exitContainer",
        image="docker/whalesay",
        command=["cowsay"],
        args=["goodbye world"],
    )
    whalesay = Container(
        name="whalesay", image="docker/whalesay", command=["cowsay"], args=["hello world"], directly_callable=True
    )
    with Steps(name="step-template"):
        whalesay().with_(name="stepA", on_exit=exit_container)
        whalesay().with_(name="stepB", on_exit=exit_container)
```

```python
"""
This example showcases how to generate and parallelize generated sequences
"""
from hera.workflows.models import Sequence


@script(directly_callable=True)
def gen_num():
    print(3)


@script(directly_callable=True)
def say(message: str):
    print(message)


with Workflow(generate_name="with-sequence-example", entrypoint="d") as w:
    with DAG(name="d"):
        t1 = gen_num().with_(name="t1")
        t2 = say("{{item}}").with_(name="t2", with_sequence=Sequence(count=t1.result, start="0"))
        t3 = say("{{item}}").with_(
            name="t3",
            with_sequence=Sequence(start=t1.result, end="5", format="2020-05-%02X"),
        )
        t1 >> [t2, t3]
```

dejamiko:
**Pull Request Checklist**
- [x] Fixes #753 
- [x] Tests added

This PR is the first step of implementing the directly callable syntax. It introduces callable syntax for scripts and a `with_` function which allows for further modification of the created `Task`/`Step`


src/hera/workflows/_mixins.py
```diff
@@ -615,23 +615,24 @@ class CallableTemplateMixin(ArgumentsMixin):
     not duplicated.
     """
 
+    directly_callable: bool = False
```
elliotgunton, line 618: I'm not totally sold on this as an interface name that conveys the right message (as a user: "callable templates are already callable, what does "directly" mean?") Thinking about something like `use_with_builder`, or `use_builder`. Naming is hard...

src/hera/workflows/_mixins.py
```diff
@@ -881,6 +946,38 @@ class TemplateInvocatorSubNodeMixin(BaseMixin):
...
 
+    def with_(self, **kwargs: dict) -> TemplateInvocatorSubNodeMixin:
+        """Update the created Step/Task object with the attributes you want to pass.
+
+        Used for setting the attributes such as name, when, or with_param. Setting with_param
+        and with_items triggers a recalculation of arguments.
+        """
+        if (
+            self.template is None
+            or not hasattr(self.template, "directly_callable")
```
elliotgunton, line 957: Why not checking `isinstance(self.template, CallableTemplateMixin)` which is equiv to these two conditions

src/hera/workflows/_mixins.py
```diff
@@ -1128,6 +1225,26 @@ def _get_params_from_items(with_items: List[Any]) -> Optional[List[Parameter]]:
     return [Parameter(name=n, value=f"{{{{item.{n}}}}}") for n in with_items[0].keys()]
 
 
+def _process_params_and_artifacts(params, artifacts):
```
elliotgunton, line 1228: Missing types

src/hera/workflows/_mixins.py
```diff
@@ -1128,6 +1225,26 @@ def _get_params_from_items(with_items: List[Any]) -> Optional[List[Parameter]]:
...
+    if len(params) == 1:
+        params = [Parameter(name=p.name, value="{{item}}") for p in params]
```
elliotgunton, line 1230: The `for p in params` looks strange/implies more items, even though we have `len(params) == 1`

src/hera/workflows/_mixins.py
```diff
@@ -672,6 +673,64 @@ def __call__(self, *args, **kwargs) -> Union[None, Step, Task]:
...
+
+        if "source" in kwargs and "with_param" in kwargs:
+            arguments += self._get_deduped_params_from_source(parameter_names, artifact_names, kwargs["source"])
+        elif "source" in kwargs and "with_items" in kwargs:
+            arguments += self._get_deduped_params_from_items(parameter_names, kwargs["with_items"])
+
+        return arguments
+
+    def _extract_arguments_directly_callable(
+        self, args: tuple, kwargs: dict
```
elliotgunton, line 690: Needs cap types
```suggestion
        self, args: Tuple, kwargs: Dict
```

src/hera/workflows/_mixins.py
```diff
@@ -1128,6 +1225,26 @@ def _get_params_from_items(with_items: List[Any]) -> Optional[List[Parameter]]:
...
+def _process_params_and_artifacts(params, artifacts):
+    if len(params) == 1:
+        params = [Parameter(name=p.name, value="{{item}}") for p in params]
+    else:
+        params = [Parameter(name=p.name, value=f"{{{{item.{p.name}}}}}") for p in params]
+    params.extend(artifacts)
+    return params
+
+
+def _item_to_arg(item, name):
```
elliotgunton, line 1237: types

src/hera/workflows/_mixins.py
```diff
@@ -1128,6 +1225,26 @@ def _get_params_from_items(with_items: List[Any]) -> Optional[List[Parameter]]:
...
+    if isinstance(item, Artifact):
+        return item._build_artifact()
+    elif isinstance(item, ModelArtifact):
+        return item  # ?
```
elliotgunton, line 1241: This looks fine, any question?

src/hera/workflows/_mixins.py
```diff
@@ -1128,6 +1225,26 @@ def _get_params_from_items(with_items: List[Any]) -> Optional[List[Parameter]]:
...
+    else:
+        params = [Parameter(name=p.name, value=f"{{{{item.{p.name}}}}}") for p in params]
+    params.extend(artifacts)
+    return params
+
+
+def _item_to_arg(item, name):
+    if isinstance(item, Artifact):
+        return item._build_artifact()
+    elif isinstance(item, ModelArtifact):
```
elliotgunton, line 1240: You could just do `if`s, no need for `elif`s (pylint should actually tell you)

src/hera/workflows/_mixins.py
```diff
@@ -1128,6 +1225,26 @@ def _get_params_from_items(with_items: List[Any]) -> Optional[List[Parameter]]:
...
+        return item  # ?
+    elif isinstance(item, Parameter):
+        return item.as_argument()
+    else:
+        return Parameter(name=name, value=serialize(item))
```
elliotgunton, line 1245: Then `else` not needed

tests/arguments_passing_regression/any_success_all_fail_new.py
```diff
@@ -0,0 +1,30 @@
+from hera.workflows import DAG, Workflow, script
```
elliotgunton, line 1: Tbf the number of tests does help instill confidence, but I'm wondering if it would be easier/clearer to have a select few regression tests that use certain features, e.g. `with_param`. I feel like we're bloating the tests with regression tests on all the examples vs targeted tests

tests/test_script_callable_syntax.py
```diff
@@ -0,0 +1,115 @@
...
+
+    # don't care about file names for the runner
+    try:
+        templates_old = output_old["spec"]["templates"]
+        templates_new = output_new["spec"]["templates"]
+        for i, t in enumerate(templates_old):
+            if "script" in t:
+                templates_old[i]["script"]["args"] = templates_new[i]["script"]["args"]
+    except KeyError:
+        pass
```
elliotgunton, line 112: Do any modules still produce a `KeyError`?

src/hera/workflows/_mixins.py
```diff
@@ -672,6 +673,64 @@ def __call__(self, *args, **kwargs) -> Union[None, Step, Task]:
             f"Callable Template '{self.name}' is not under a Workflow, Steps, Parallel, or DAG context"  # type: ignore
         )
 
+    def _extract_arguments(self, kwargs: dict) -> List[Union[Parameter, Artifact, Dict]]:
```
elliotgunton, line 676:
```suggestion
    def _extract_arguments(self, kwargs: Dict) -> List[Union[Parameter, Artifact, Dict]]:
```

src/hera/workflows/_mixins.py
```diff
@@ -881,6 +946,38 @@ class TemplateInvocatorSubNodeMixin(BaseMixin):
     when: Optional[str] = None
     with_sequence: Optional[Sequence] = None
 
+    def with_(self, **kwargs: dict) -> TemplateInvocatorSubNodeMixin:
```
elliotgunton, line 949:
```suggestion
    def with_(self, **kwargs: Dict) -> TemplateInvocatorSubNodeMixin:
```

src/hera/workflows/_mixins.py
dejamiko: >There are only two hard things in Computer Science: cache invalidation and naming things.
>
>-- Phil Karlton
I am not a big fan of `builder` because it's not exactly a builder either. Maybe somehow related to "first-call-the-function"? I am open to suggestions.

src/hera/workflows/_mixins.py
dejamiko: True, I took it from somewhere else in the _mixins code

tests/test_script_callable_syntax.py
dejamiko: Yes, the ones that use the inline constructor

src/hera/workflows/_mixins.py
elliotgunton: More suggestions/ideas 😂
* `use_argument_kwargs`
* `use_function_kwargs`
* `use_inputs_as_kwargs` -> This then matches the idea for `Container`s?

tests/arguments_passing_regression/any_success_all_fail_new.py
dejamiko: Maybe the regression tests should be a part of a separate test suite that only runs on PRs, not on `make test`?

src/hera/workflows/_mixins.py
dejamiko: How about something like `function_params_only`? It is difficult to name 😢

elliotgunton: I think this is the clearest/most descriptive while being as succinct as possible 😂
```suggestion
    use_func_params_in_call: bool = False
```

src/hera/workflows/_mixins.py
```diff
@@ -616,23 +616,24 @@ class CallableTemplateMixin(ArgumentsMixin):
...
         # when the `source` is set via an `@script` decorator, it does not come in with the `kwargs` so we need to
         # set it here in order for the following logic to capture it
         if "source" not in kwargs and hasattr(self, "source"):
             kwargs["source"] = self.source  # type: ignore
 
-        if "source" in kwargs and "with_param" in kwargs:
-            arguments += self._get_deduped_params_from_source(parameter_names, artifact_names, kwargs["source"])
-        elif "source" in kwargs and "with_items" in kwargs:
-            arguments += self._get_deduped_params_from_items(parameter_names, kwargs["with_items"])
+        if hasattr(self, "directly_callable") and self.directly_callable:
```
elliotgunton, line 630: Why would a `self` of `CallableTemplateMixin` not have the attr?

src/hera/workflows/_mixins.py
```diff
@@ -616,23 +616,24 @@ class CallableTemplateMixin(ArgumentsMixin):
...
+            function_kwargs = {k: v for k, v in kwargs.items() if k not in ["source", "name"]}
+            arguments = self._extract_arguments_directly_callable(args, function_kwargs)
```
elliotgunton, line 632: We're not necessarily dealing with a function here - you have examples of `directly_callable` `Containers`

src/hera/workflows/_mixins.py
```diff
@@ -616,23 +616,24 @@ class CallableTemplateMixin(ArgumentsMixin):
...
         # set it here in order for the following logic to capture it
         if "source" not in kwargs and hasattr(self, "source"):
             kwargs["source"] = self.source  # type: ignore
 
-        if "source" in kwargs and "with_param" in kwargs:
-            arguments += self._get_deduped_params_from_source(parameter_names, artifact_names, kwargs["source"])
-        elif "source" in kwargs and "with_items" in kwargs:
-            arguments += self._get_deduped_params_from_items(parameter_names, kwargs["with_items"])
+        if hasattr(self, "directly_callable") and self.directly_callable:
+            function_kwargs = {k: v for k, v in kwargs.items() if k not in ["source", "name"]}
```
elliotgunton, line 631: We can make this code clearer and cleaner because in this scope we know all the args/kwargs that were passed in were for the function itself, so it would be good to create a new feature "if" branch on `if self.directly_callable` and build up the function `Parameter`/`Artifact` arguments there (via `_extract_arguments_directly_callable`), instead of manipulating the given `kwargs` dict (adding/deleting value) and passing to `return Step(*args, template=self, **kwargs)`.

src/hera/workflows/_mixins.py
```diff
@@ -616,28 +616,13 @@ class CallableTemplateMixin(ArgumentsMixin):
...
-            arguments += self._get_deduped_params_from_source(parameter_names, artifact_names, kwargs["source"])
-        elif "source" in kwargs and "with_items" in kwargs:
-            arguments += self._get_deduped_params_from_items(parameter_names, kwargs["with_items"])
-
-        # it is possible for the user to pass `arguments` via `kwargs` along with `with_param`. The `with_param`
-        # additional parameters are inferred and have to be added to the `kwargs['arguments']`, otherwise
-        # the step/task will miss adding them when building the final arguments
-        kwargs["arguments"] = arguments
+    def __call__(self, *args, **kwargs) -> Union[None, Step, Task]:
+        if self.use_func_params_in_call and global_config.experimental_features["use_func_params_in_call"]:
```
flaviuvadan, line 622: Should this raise an error when `use_func_params_in_call` is not `True` but `self.use_func_params_in_call` is True? Similar to how Hera raises an error if the script runner is not enabled but a client submits `constructor="runner"`

src/hera/workflows/_mixins.py
```diff
@@ -616,28 +616,13 @@ class CallableTemplateMixin(ArgumentsMixin):
...
+            args, kwargs = self._process_func_args_and_kwargs_for_callable(args, kwargs)
+        else:
+            kwargs = self._process_kwargs_for_callable(kwargs)
```
flaviuvadan, line 626: I think all of this should still happen, though. If users don't want to rely on callables (1) they should feel free to do so and (2) this is a breaking change

flaviuvadan: Oh nvm it's just moved

src/hera/workflows/_mixins.py
```diff
@@ -616,28 +616,13 @@ class CallableTemplateMixin(ArgumentsMixin):
...
     """
 
-    def __call__(self, *args, **kwargs) -> Union[None, Step, Task]:
-        if "name" not in kwargs:
-            kwargs["name"] = self.name  # type: ignore
-
-        arguments = self._get_arguments(**kwargs)
-        parameter_names = self._get_parameter_names(arguments)
-        artifact_names = self._get_artifact_names(arguments)
+    use_func_params_in_call: bool = False
```
flaviuvadan, line 619: @elliotgunton I wonder if this should be called something simpler like `callable_syntax`?

dejamiko: I agree, this was my favourite name from the beginning! https://github.com/argoproj-labs/hera/pull/763#discussion_r1316968243

flaviuvadan: We should use that!

samj1912:
Closing this for now. We can revisit this later.

PR #767, opened 2023-09-06
Fix Hera parameter inference

### LINKED ISSUE #766: Hera does not prioritize user-specified arguments in fanouts

- [x] No

- [ ] running Hera code and submitting to Argo

<h2>Bug report</h2>

**Describe the bug**

Fanouts with extra parameters in Python functions that are specified via `arguments` can result in duplicate parameter specifications in the arguments field.

**To Reproduce**
Full Hera code to reproduce the bug, taken from Hera example:
```py


@script()
def generate():
    import json
    import sys

    json.dump([i for i in range(10)], sys.stdout)


@script()
def consume(value: int, param1: str, param2: int = 42):
    print(value, param1, param2)


with Workflow(generate_name="dynamic-fanout-", entrypoint="d") as w:
    with DAG(name="d"):
        g = generate()
        c = consume(
            with_param=g.result,
            arguments={
                "value": "{{item}}",
                "param1": "hello",
                "param2": 42,
            },
        )
        g >> c

w.create()
```

These are the arguments that are set:
Screenshot 2023-09-04 at 7 38 58 PM

**Expected behavior**
No duplicate arguments. Hera should assume the user knows what they are doing and prioritize taking the user specified arguments rather than infer the `{{item.param}}` for every param

flaviuvadan:
**Pull Request Checklist**
- [x] Fixes #766
- [x] Tests added

**Description of PR**
See #766 for a description of the bug. 

This PR adjusts the way parameter names are inferred. When checking the intersection between user supplied parameters + Hera inferred parameters Hera will check _all_ parameter names, not just args, plus Hera will now also account for simple string keys in arguments not just `Parameter`/`ModelParameter`

### DIFF
--- a/src/hera/workflows/_mixins.py
+++ b/src/hera/workflows/_mixins.py
@@ -1,6 +1,7 @@
 """Core collection of Hera mixins that isolate shareable functionality between Hera objects."""
 from __future__ import annotations
 
+import functools
 import inspect
 from pathlib import Path
 
@@ -68,7 +69,7 @@ from hera.workflows.models import (
     VolumeDevice,
     VolumeMount,
 )
-from hera.workflows.parameter import MISSING, Parameter
+from hera.workflows.parameter import Parameter
 from hera.workflows.protocol import Templatable, TWorkflow
 from hera.workflows.resources import Resources
 from hera.workflows.user_container import UserContainer
@@ -687,13 +688,14 @@ class CallableTemplateMixin(ArgumentsMixin):
 
     def _get_parameter_names(self, arguments: List) -> Set[str]:
         """Returns the set of parameter names that are currently set on the mixin inheritor."""
-        parameters = [arg for arg in arguments if isinstance(arg, ModelParameter) or isinstance(arg, Parameter)]
-        return {p.name for p in parameters}
+        parameters = [arg for arg in arguments if isinstance(arg, (ModelParameter, Parameter))]
+        keys = [arg for arg in arguments if isinstance(arg, dict)]
+        return {p.name for p in parameters}.union(set(functools.reduce(lambda x, y: x + list(y.keys()), keys, [])))
 
     def _get_artifact_names(self, arguments: List) -> Set[str]:
         """Returns the set of artifact names that are currently set on the mixin inheritor."""
-        artifacts = [arg for arg in arguments if isinstance(arg, ModelArtifact) or isinstance(arg, Artifact)]
-        return {a.name for a in artifacts if a.name}
+        artifacts = [arg for arg in arguments if isinstance(arg, (ModelArtifact, Artifact))]
+        return {a if isinstance(a, str) else a.name for a in artifacts if a.name}
 
     def _get_deduped_params_from_source(
         self, parameter_names: Set[str], artifact_names: Set[str], source: Callable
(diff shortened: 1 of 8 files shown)


src/hera/workflows/_mixins.py
```diff
@@ -687,13 +688,14 @@ def _get_arguments(self, **kwargs) -> List:
 
     def _get_parameter_names(self, arguments: List) -> Set[str]:
         """Returns the set of parameter names that are currently set on the mixin inheritor."""
-        parameters = [arg for arg in arguments if isinstance(arg, ModelParameter) or isinstance(arg, Parameter)]
-        return {p.name for p in parameters}
+        parameters = [arg for arg in arguments if isinstance(arg, (ModelParameter, Parameter))]
+        keys = [arg for arg in arguments if isinstance(arg, dict)]
+        return {p.name for p in parameters}.union(set(functools.reduce(lambda x, y: x + list(y.keys()), keys, [])))
```
flaviuvadan, line 693: This essentially happens because Hera does not use `self._build_arguments()` [here](https://github.com/argoproj-labs/hera/blob/d956276f67f73a14ea72d43de494d156998b5f42/src/hera/workflows/_mixins.py#L676). When Hera uses `self.arguments` the arguments are _not_ built yet, hence it does not actually see the names of the other parameters _post_ processing from dict to `Parameter`s. An alternative solution to this PR is to call build first

src/hera/workflows/_mixins.py
```diff
@@ -1094,18 +1098,19 @@ def _get_params_from_source(source: Callable) -> Optional[List[Parameter]]:
...
     submitted to the Argo server.
     """
-    source_signature: Dict[str, Optional[object]] = {}
+    source_signature: List[str] = []
     for p in inspect.signature(source).parameters.values():
-        if p.default != inspect.Parameter.empty and p.kind == inspect.Parameter.POSITIONAL_OR_KEYWORD:
-            source_signature[p.name] = p.default
-        else:
-            source_signature[p.name] = MISSING
+        if p.default == inspect.Parameter.empty and p.kind == inspect.Parameter.POSITIONAL_OR_KEYWORD:
```
dejamiko, line 1103: With those changes, the docstring of the function needs to be updated

src/hera/workflows/_mixins.py
```diff
@@ -1085,27 +1089,31 @@ def get_parameter(self, name: str) -> Parameter:
...
+    """Returns an optional list of `Parameter` from the specified `source`.
 
-    Notes:
-    -----
-    If the value of an identified `Callable` keyword argument is found to be empty the value of
-    `hera.shared.serialization.MISSING` is used as a placeholder. This is later serialized as `None` -> `null` when
-    submitted to the Argo server.
+    This infers that each non-keyword, positional, argument of the given source is a parameter that stems from a
+    fanout. Therefore, each parameter value takes the form of `{{item}}` when there's a single argument or
+    `{{item.[argument name]}}` when there are other arguments.
```
elliotgunton, line 1097: Just because it looks like dictionary access
```suggestion
    `{{item.<argument name>}}` when there are other arguments.
```

src/hera/workflows/_mixins.py
```diff
@@ -1085,27 +1089,31 @@ def get_parameter(self, name: str) -> Parameter:
         return self._get_parameter(name=name, subtype=self._subtype)
 
 
-def _get_params_from_source(source: Callable) -> Optional[List[Parameter]]:
-    """Assembles an optional list of `Parameter` from the given `Callable` arguments.
+def _get_param_items_from_source(source: Callable) -> Optional[List[Parameter]]:
```
elliotgunton, line 1092: I'm not sure the return needs to be Optional? An empty list suffices for existence check, and you can shortcut straight to
```py
for p in new_parameters:
```
at the call site

src/hera/workflows/_mixins.py
```diff
@@ -1085,27 +1089,31 @@ def get_parameter(self, name: str) -> Parameter:
...
+            # kwargs. Otherwise, we assume that the user sets the value of the parameter via the `with_param` field
+            source_signature.append(p.name)
 
     if len(source_signature) == 0:
         return None
     elif len(source_signature) == 1:
-        return [Parameter(name=n, value="{{item}}") for n in source_signature.keys()]
-    return [Parameter(name=n, value=f"{{{{item.{n}}}}}") for n in source_signature.keys()]
+        return [Parameter(name=n, value="{{item}}") for n in source_signature]
+    return [Parameter(name=n, value=f"{{{{item.{n}}}}}") for n in source_signature]
```
elliotgunton, line 1116: Going off the `Optional` comment, this section can then be simplified just to

```py
if len(source_signature) == 1:
    return [Parameter(name=n, value="{{item}}") for n in source_signature]
return [Parameter(name=n, value=f"{{{{item.{n}}}}}") for n in source_signature]
```

src/hera/workflows/_mixins.py
```diff
@@ -1085,27 +1088,29 @@ def get_parameter(self, name: str) -> Parameter:
         return self._get_parameter(name=name, subtype=self._subtype)
 
 
-def _get_params_from_source(source: Callable) -> Optional[List[Parameter]]:
-    """Assembles an optional list of `Parameter` from the given `Callable` arguments.
+def _get_param_items_from_source(source: Callable) -> List[Parameter]:
+    """Returns an optional list of `Parameter` from the specified `source`.
```
elliotgunton, line 1092:
```suggestion
    """Returns a list (possibly empty) of `Parameter` from the specified `source`.
```

src/hera/workflows/_mixins.py
```diff
@@ -1085,27 +1088,29 @@ def get_parameter(self, name: str) -> Parameter:
...
-    `hera.shared.serialization.MISSING` is used as a placeholder. This is later serialized as `None` -> `null` when
-    submitted to the Argo server.
+    This infers that each non-keyword, positional, argument of the given source is a parameter that stems from a
+    fanout. Therefore, each parameter value takes the form of `{{item}}` when there's a single argument or
+    `{{item.<argument name>}}` when there are other arguments.
+
+    Returns:
+    -------
+    Optional[List[Parameter]]
+        An optional list of identified parameters or `None` when none are identified.
```
elliotgunton, line 1101:
```suggestion
    List[Parameter]
        A list of identified parameters (possibly empty).
```

elliotgunton (approved):
:shipit:

PR #778, opened 2023-09-13
Add support for better output references

### LINKED ISSUE #777: Output reference improvements
## How do we currently pass outputs between steps/tasks in Hera?
```python
with Workflow(generate_name="artifact-passing-", entrypoint="artifact-example") as w:
    with Steps(name="artifact-example") as s:
        whalesay(name="generate-artifact")
        print_message(
            name="consume-artifact",
            arguments=[Artifact(name="message", from_="{{steps.generate-artifact.outputs.artifacts.hello-art}}")],
        )
```

## How would we ideally pass outputs between steps/tasks?
```python
@script(outputs=Artifact(name="hello-art", path="/tmp/hello_world.txt"))
def whalesay():
   with open("/tmp/hello_world.txt", "w") as f:
       f.write("hello world")


@script(inputs=Artifact(name="message", path="/tmp/message"))
def print_message():
   with open("/tmp/message", "r") as f:
       message = f.readline()
   print(message)


with Workflow(generate_name="artifact-passing-", entrypoint="artifact-example") as w:
    with Steps(name="artifact-example") as s:
        step = whalesay(name="generate-artifact")
        print_message(
            message=step.outputs.artifacts.hello_art,
            # or without the callable syntax
            arguments=[Artifact(name="message", from_=step.outputs.artifacts.hello_art)],
        )
```

#### Part 1
`arguments=[Artifact(name="message", from_=step.outputs.artifacts.hello_art)]`

We need to dynamically generate the outputs of the step based on the outputs of the template it contains. We populate the outputs field with the dictionary `{output_name: output_path_string}`.

`output_name` - the name of the output. We can use the name of the `Artifact`/`Parameter` and its `type` to generate the output name. To use the dot notation, if the output is kebab-case, we can change it to snake_case. This can be seen above with `hello-art` changing into `hello_art`. This approach might have issues with clashing names, if we have the same name, differing by just the “-” and “_”. That problem might be solvable using Pydantic’s `field aliases` https://docs.pydantic.dev/latest/usage/fields/#field-aliases. Unfortunately, it seems like aliases are only a Pydantic 2 feature, so this feature will have to wait for #713. For now, the underscore mapping can be done and, in case of a clash, we can raise an Error and ask the user to rename the clashing output.

`output_path_string` - for the yaml to be correctly generated, the `step.outputs.hello_art` is not an `Artifact` object. It is the “path-within-the-workflow” to access the correct output. In the above example, the `generated_string` would be `"{{steps.generate-artifact.outputs.artifacts.hello-art}}"`. The string starts with the context; for `Steps` it is `steps` and for `DAG` it is `tasks`. Then, we access the `Step`/`Task` which generates the output we want to access by its `name`, then `outputs`, then depending on the type of the output we have `artifacts` or `parameters`, and finally the output `name`. 

#### Part 2
`message=step.outputs.artifacts.hello_art`

To achieve this, if the input item is an `Artifact` and we’re passing a string to it, we can generate the `Artifact(name="input-name", from_=passed_string)` in the `__call__` function. This depends on the Callable syntax https://github.com/argoproj-labs/hera/pull/763. 

## Implementation
To implement the above described feature, I suggest creating a mixin used by both `Step` and `Task` classes. This mixin has a method which creates and populates the `outputs` field. Then, at some point in the creation of the `Step`/`Task`, we need to call that method.

dejamiko:
**Pull Request Checklist**
- [x] Fixes #777 
- [x] Tests added

This PR introduces an `EasyOutputsMixin` which implements the design specified in #777. It does not use `field aliases` yet, so there might be problems with names differing only by "-" "_" clashing.

### DIFF
--- a/src/hera/workflows/_mixins.py
+++ b/src/hera/workflows/_mixins.py
@@ -1280,14 +1280,17 @@ class ExperimentalMixin(BaseMixin):
 
 
 class EasyOutputsMixin(BaseMixin):
+    """This mixin allows for the easy output access in subsequent `Step`s/`Task`s."""
+
     outputs: Optional[PyBaseModel] = None
 
     def __init__(self, *args, **kwargs):
+        """We initialize the inputs when this object is created."""
         super().__init__(*args, **kwargs)
         self._build_outputs()
 
-    # this needs to be called when the step is built
     def _build_outputs(self) -> None:
+        """We build the outputs for the Step/Task based on the outputs from `self.template`."""
         from hera.workflows import Step, Task
 
         if (
@@ -1329,7 +1332,7 @@ class EasyOutputsMixin(BaseMixin):
         self.outputs.parameters = dynamic_params  # type: ignore
 
     def _generate_output_string(self, output: Union[Artifact, ModelArtifact, Parameter, ModelParameter]) -> str:
-        """Generate the new output by generating the output_name and the output_path_string."""
+        """Generate the new output_string."""
         if not hasattr(self, "name") or not self.name:
             raise ValueError(f"{self} has no name.")
         from hera.workflows import Step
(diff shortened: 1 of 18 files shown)


src/hera/workflows/_mixins.py
```diff
@@ -1272,3 +1277,93 @@ def _check_enabled(cls, values):
         if not global_config.experimental_features[cls._flag]:
             raise ValueError(cls._experimental_warning_message.format(cls, cls._flag))
         return values
+
+
+class EasyOutputsMixin(BaseMixin):
```
flaviuvadan, line 1282: I suggest adding a `subtype` field to this, so that child classes such as Step and Task can set it. That way this does not have to import `Step` or `Task` for type checking, and can rely on the `subtype` field to sub in the right string the same way `TemplateInvocatorSubNodeMixin` does it

flaviuvadan (commented):
LGTM! Just a single comment on use of task and step

samj1912:
Closing this for now. We can revisit this later.

PR #780, opened 2023-09-18
Fixup volumes

elliotgunton:
* Still use /tmp for script outputs but allow mounting own volume with `volume_for_outputs` passed to Runner constructor
* Fix examples using volumes incorrectly and simplify where possible


src/hera/workflows/_mixins.py
```diff
@@ -1061,7 +1061,7 @@ def _get_artifact(self, name: str, subtype: str) -> Artifact:
 
         obj = next((output for output in artifacts if output.name == name), None)
         if obj is not None:
-            return Artifact(name=name, path=obj.path, from_=f"{{{{{subtype}.{self.name}.outputs.artifacts.{name}}}}}")
```
dejamiko, line 1064: This return object looks strange to me. It has a given `name` and the `from_` has the same name in the path? We don't return the Artifact but rather create a new one with a populated `from_` string. If we are doing that, for clarity, we might want to change the function slightly to combine the `get_artifact` with `as_name` and use the provided "new name" for the return object. We can make the `new_name` equal to `old_name` for convenience but I feel like this would be clearer to the user

elliotgunton: I agree it looks strange, but it makes a bit more sense from the user's perpective e.g. for parameters we have `with_name` which kind of matches our new callable syntax, and is used like:

```py
t1 = Task(name="a", template=out)
t2 = Task(name="b", template=in_, arguments=t1.get_parameter("x").with_name("a"))
```

And for artifacts we have `as_name` (which we should probably deprecate and use `with_name` as well):

```py
out = output_artifact(arguments={"a_number": 3})
use_artifact(arguments=[out.get_artifact("successor_out").as_name("successor_in")])
```

Your suggestion would be a small change to e.g.
```py
out = output_artifact(arguments={"a_number": 3})
use_artifact(arguments=[out.get_artifact("successor_out", with_name="successor_in")])
```

Which I'm open to, but I'd rather nail that change down in the new callable syntax PR, this line change was just to remove the `path` which isn't used for artifacts-as-arguments.

samj1912 (approved):

PR #782, opened 2023-09-19
Add k8s example to auth docs

flaviuvadan:
**Pull Request Checklist**
- [ ] Fixes #
- [ ] Tests added

Vikram pointed out in the [CNCF #hera channel](https://cloud-native.slack.com/archives/C03NRMD9KPY/p1695030320402139) that we no longer have a K8s auth example. This PR adds back the example to the docs

### DIFF
--- a/docs/walk-through/authentication.md
+++ b/docs/walk-through/authentication.md
@@ -1,8 +1,7 @@
 # Authentication
 
-There are multiple ways you can authenticate with Argo Workflows. The way you authenticate generally depends on your
-unique organization setup. If you submit workflows through Hera directly you have multiple ways to authenticate with
-the Argo server.
+The way you authenticate generally depends on your unique organization setup. If you submit workflows through Hera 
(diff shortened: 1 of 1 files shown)


docs/walk-through/authentication.md
```diff
@@ -100,6 +101,55 @@ class MyTokenGenerator(TokenGenerator):
...
+
+##### Generating a K8s token via `TokenGenerator`
+
+```python
+import base64
+import errno
+import os
+from typing import Optional
+
+from kubernetes import client, config
```
elliotgunton, line 124: This would require users to pip install kubernetes right? Could we make a note of that

elliotgunton (commented):
LGTM but we should add a note about the new dependency that a user would need to install

docs/walk-through/authentication.md
flaviuvadan: Sure thing!

elliotgunton (approved):
LGTM!

PR #783, opened 2023-09-19
[WIP] Add failing test for Hera runner output serialization

flaviuvadan:
**Pull Request Checklist**
- [ ] Fixes #
- [x] Tests added

**Description of PR**
@lancour discovered that typing a field as a union such as `str | int | float` + using the Hera runner results in `int` + `float` being serialized as `str`. This is likely caused by the way serialization occurs during runner validation here. My guess is that it happens somewhere around https://github.com/argoproj-labs/hera/blob/main/src/hera/workflows/runner.py#L234

This PR only adds a failing test to illustrate how an expected integer is actually a string, hence it fails. Still working through the code

### DIFF
--- a/examples/workflows/callable_script.py
+++ b/examples/workflows/callable_script.py
@@ -1,4 +1,4 @@
-from typing import List
+from typing import List, Union
 
 try:
     from typing import Annotated  # type: ignore
@@ -32,6 +32,7 @@ global_config.experimental_features["script_annotations"] = True
 class Input(BaseModel):
     a: int
     b: str = "foo"
+    c: Union[str, int, float]
 
 
 # An optional pydantic output type
@@ -73,8 +74,9 @@ def str_function(input: str) -> Output:
 def function_kebab(
     a_but_kebab: Annotated[int, Parameter(name="a-but-kebab")] = 2,
     b_but_kebab: Annotated[str, Parameter(name="b-but-kebab")] = "foo",
+    c_but_kebab: Annotated[float, Parameter(name="c-but-kebab")] = 42.0,
 ) -> Output:
-    return Output(output=[Input(a=a_but_kebab, b=b_but_kebab)])
+    return Output(output=[Input(a=a_but_kebab, b=b_but_kebab, c=c_but_kebab)])
 
 
 @script()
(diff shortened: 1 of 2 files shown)


elliotgunton:
I would say this is WAD/the issue stems from Pydantic - the order of the Union types dictates how the class is serialised. The basic idea is that the stricter types should come first, as it will be the first type can coerce the value that is used. Or, you can set the config to use `smart_union` - this is  explained [here](https://docs.pydantic.dev/1.10/usage/model_config/#smart-union) in pydantic 1.10. Not sure about 2.0+

Therefore it's a user issue to set their types correctly - I'm happy to add the test as an example and use the `smart_union` config to get it to pass?

src/hera/workflows/runner.py
```diff
@@ -265,7 +265,10 @@ def _runner(entrypoint: str, kwargs_list: List) -> Any:
     else:
         kwargs = _map_argo_inputs_to_function(function, kwargs)
 
-    function = validate_arguments(function)
+    # using smart union by default just in case clients do not rely on it. This means that if a function uses a union
+    # type for any of its inputs, then this will at least try to map those types correctly if the input object is
+    # not a pydantic model with smart_union enabled
+    function = validate_arguments(function, config=dict(smart_union=True))
```
flaviuvadan, line 271: @elliotgunton also added the smart union config here, see comment

flaviuvadan:
@elliotgunton if you have a chance and want to approve I think this is ready to ship

src/hera/workflows/runner.py
elliotgunton: I'm not sure about this change in terms of it being a behaviour change for users. Also I think we can do something to let users set the whole pydantic config themselves - some ideas:

* what about a `config` kwarg into the script decorator?
* Or a separate decorator e.g. `script_runner_config` that will let users set these (and more) runtime configs?
* Or moving it into the script runner constructor class and fetching it out of the script decorator?

flaviuvadan: I'd argue this is currently limiting users, though, since the current setup can lead to all sorts of confusing scenarios such as "why am I getting a string from my union when this thing is definitely an integer" and am not sure whether the configuration would be _enabling_, so to say 🤔

elliotgunton (approved):

PR #792, opened 2023-09-30
Make path optional for input Artifact annotations; deprecate as_name

### LINKED ISSUE #787: Script runner with input annotation artifacts, Path could be optional
When creating a script function with an input Artifact in the annotations, the path can be whatever you like as it is basically ignored during runtime except to load the file into the data type, in this case an `int`:

```py
@script(constructor="runner")
def use_artifact(
    successor_in: Annotated[
        int,
        Artifact(name="successor_in", path="/wherever", loader=ArtifactLoader.json),
    ]
):
    print(successor_in)
```

It should be optional to omit the path and set it based on the function parameter name. Maybe a default inputs directory would be used like `/hera/inputs/artifacts` so the actual path used in this example would be `/hera/inputs/artifacts/successor_in`
```py
@script(constructor="runner")
def use_artifact(
    successor_in: Annotated[
        int,
        Artifact(name="successor_in", loader=ArtifactLoader.json),
    ]
):
    print(successor_in)
```

We add the path manually to create valid yaml, but from a user perspective it's ignored anyway.

**Additional context**

elliotgunton:
**Pull Request Checklist**
- [x] Fixes #787 
- [x] Tests added

**Description of PR**
See #787.

Also deprecated `as_name` in favour of `with_name` for artifacts, as the former builds the artifact rather than making a copy. Also, it now mirrors the naming and functionality of the `with_name` function from `Parameter`.


src/hera/workflows/artifact.py
```diff
@@ -112,11 +118,21 @@ def _build_artifact_paths(self) -> _ModelArtifactPaths:
...
+        logger.warning("'as_name' is deprecated, use 'with_name'")
         artifact = self._build_artifact()
         artifact.name = name
         return artifact
 
+    def with_name(self, name: str) -> Artifact:
+        """Returns a copy of the current artifact, renamed using the specified `name`."""
+        artifact = self.copy(deep=True)
+        artifact.name = name
+        return artifact
```
flaviuvadan, line 134: Small request - can we add a `with_path` as well? Heard some feedback from the community that it would be very helpful, it was part of v4

src/hera/workflows/script.py
```diff
@@ -62,6 +62,9 @@
     from typing_extensions import Annotated  # type: ignore
 
 
+_DEFAULT_ARTIFACT_INPUT_DIRECTORY = "/tmp/hera/inputs/artifacts/"
```
flaviuvadan, line 65: Is this folder created by Hera when annotations are used? I think writing to it might fail if it does not exist

src/hera/workflows/artifact.py
elliotgunton: The path is dictated by the template though, that only worked in v4 because the template/tasks were linked together

PR #794, opened 2023-10-01
Replace ruff excludes with pydocstyle ignores

JacobHayes:
**Pull Request Checklist**
- [-] Fixes #
- [-] Tests added
- [-] Documentation/examples added

**Description of PR**
Previously we excluded ruff checking and formatting for all of the examples, tests, and generated models in order to ignore the unneeded (for these modules) docstring checks. Aside from the unneeded docstrings checks though, that disabled import formatting/pruning and a couple other small things.

Instead of completely excluding those modules, this PR ignores the docstring checks specifically for those modules in order to keep the other ruff checks/formatting enabled (and my editor happy when it tries to save-and-format these files - yes I should probably fix that :grin:).

---

I separate the config change and fixes in separate commits, so it's likely easier to review commit-by-commit:
- 1: the config change (the main change)
- 2: removes some unused imports
- 3: adds a missing parameter in a task function
- 4: the result of `ruff --fix .` over the remaining `Import block is un-sorted or un-formatted` errors
- 5: update docs with the template import changes


elliotgunton (approved):
Another very helpful QoL fix! Thank you!

PR #795, opened 2023-10-01
Add support for pydantic v2

### LINKED ISSUE #713: Support Pydantic v2
Pydantic v2 was [released recently](https://docs.pydantic.dev/latest/blog/pydantic-v2-final/) with a nice [migration guide](https://docs.pydantic.dev/latest/migration/) and even a helpful [code rewriter](https://docs.pydantic.dev/latest/migration/#code-transformation-tool). Even more useful for libraries to provide a gradual transition by supporting v1 and v2, there's a [backwards compatible `from pydantic.v1 import BaseModel` import](https://docs.pydantic.dev/latest/migration/#continue-using-pydantic-v1-features).

As `hera` is a library, support for both v1 and v2 would be helpful to ease upgrades for users who may have other v1 models. Perhaps to start the migration and support both versions with very little other rewrites, `hera` could relax the dep pin to support v2 and then add an import try/except like:

```python
try:
except ImportError:
    from pydantic.v1 import BaseModel
```

and then cut a release that supports both v1 and v2.

Eventually, the `hera` models should change to use the new v2 `pydantic.BaseModel`, at which point supporting both v1 and v2 may be a bit harder (depending on `class Config` vs `model_config`, etc) and v1 support could be dropped.

**Additional context**
[I posted in Slack too](https://cloud-native.slack.com/archives/C03NRMD9KPY/p1689628321669929)

JacobHayes:
**Pull Request Checklist**
- [x] Fixes #713
- [ ] Tests added

**Description of PR**
WIP, but wanted to get the bulk of it up for any feedback while I:
- add a test run in CI with pydantic v1 installed (to ensure existing users are covered - they passed manually)
- add a couple more tests and try to support serializing v1 *and* v2 user-defined models (not just the v1 models from either version - I think this should be possible).
- update the model generation to update the Field import

---

This PR updates hera to support use with either pydantic v1 or v2 installed. All hera internal code now imports pydantic objects from `hera.shared._pydantic` (extended and renamed from `_base_model`). This module ensures that we're always using v1 compatible objects, which allows us to support codebases installing either pydantic v1 or v2 (although the models passed to Hera must be v1 for pydantic v1 and v2 for pydantic v2).


JacobHayes:
Sorry for the large PR w/o much prior discussion - I think a lot of it is pretty rote transformations though (updating imports).

matty-rose:
hey @JacobHayes bit of a driveby comment but i'd done a little digging as well into introducing pydantic v2 compatibility here and ran into the exact issues with the `codegen` stuff seen in the CI failures here - the commit from this PR - https://github.com/JacobHayes/hera/pull/1 i think should help fix those. hope this helps a little, super keen to see this PR land!

JacobHayes:
I'd still like to add a test run that overrides the `poetry install` in CI to use pydantic v1 to make sure that does (and continues to) work.

Do you all have any suggestions? Perhaps adding a simplified `test-pydantic-v1` job to `.github/workflows/cicd.yaml` that overrides the installed version (that doesn't use a matrix, upload coverage, etc)?

flaviuvadan:
> I'd still like to add a test run that overrides the `poetry install` in CI to use pydantic v1 to make sure that does (and continues to) work.
> 
> Do you all have any suggestions? Perhaps adding a simplified `test-pydantic-v1` job to `.github/workflows/cicd.yaml` that overrides the installed version (that doesn't use a matrix, upload coverage, etc)?

This sounds great to me. Could it be included in the matrix build maybe?

JacobHayes:
Are you thinking a new `pydantic_major_version: [1, 2]` arg and just always do something like `pip install 'pydantic==$pydantic_major_version` (w/ whatever tweaks pip needs to "take latest for that major version") after the `poetry install`? It does add a lot of new jobs but I guess the parallelism takes care of that haha.

One thing I'm not certain about is whether `mypy` will work across both versions (which is one reason for the test in the first place :grin:) - hopefully we don't need to tweak the job any if there are issues there.

flaviuvadan:
> It does add a lot of new jobs but I guess the parallelism takes care of that haha

Yea, that's what I was thinking! And also this would be temporary. Perhaps we can remove it in a few months.

> One thing I'm not certain about is whether mypy will work across both versions (which is one reason for the test in the first place 😁) - hopefully we don't need to tweak the job any if there are issues there.

Hopeful that we don't have to change it :)

flaviuvadan:
@samj1912 and @elliotgunton this PR + approach LGTM. Any chance I can trouble you to review? Also weigh in on the Q of [pydantic V1 vs V2 testing](https://github.com/argoproj-labs/hera/pull/795#issuecomment-1758099872)

JacobHayes:
Going to work on adding pydantic versions to build matrix here in a bit

.github/workflows/cicd.yaml
```diff
@@ -53,6 +53,39 @@ jobs:
         with:
           path: dist/*.whl
 
+  # Run tests against pydantic v1 until we drop support.
+  test-pydantic-v1:
```
JacobHayes, line 57: It was easier to just duplicate with a separate job so we can:
- remove quickly after dropping support for pydantic v1
- skip the build+upload steps (which would otherwise write duplicate wheels or require some more complex conditional logic)
- test the `poetry` locked version in the main job w/o a separate job to look it up first (or gross logic in "Install dependencies")

flaviuvadan:
@elliotgunton and @samj1912 nudge to review this. Again, LGTM, and I am a fan of the second job approach to verify pydantic v1 backwards compatibility

flaviuvadan:
@JacobHayes what do you think about appending 
```toml
[mypy-pydantic.*]
ignore_missing_imports = True
```
to `mypy.ini` prior to running pydantic v1 tests in CI?

src/hera/shared/_pydantic.py
```diff
@@ -1,8 +1,39 @@
...
+# users across both versions.
+try:
+    from pydantic.v1 import (
+        BaseModel as PydanticBaseModel,
+        Field,
+        ValidationError,
+        root_validator,
+        validate_arguments,
+        validator,
+    )
```
elliotgunton, line 13: Might be worth talking about how we, as maintainers of Hera, should be using Pydantic v2 - eventually we'll need to use v2 classes and properly migrate. Is there any way we can do that gradually/make it easier than an eventual big bang major version upgrade? cc @flaviuvadan

.github/workflows/cicd.yaml
```diff
@@ -53,6 +53,39 @@ jobs:
...
+      - name: setup python ${{ matrix.python-version }}
+        uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python-version }}
+          cache: "poetry"
+
+      - name: Install dependencies
+        run: |
+          poetry install
+          poetry run pip install "pydantic<2"
```
elliotgunton, line 84: I suppose something like
```
echo "\n[mypy-pydantic.*]\nignore_missing_imports = True\n" >> mypy.ini
```

would go here [to address @flaviuvadan's idea](https://github.com/argoproj-labs/hera/pull/795#issuecomment-1786175455)?

elliotgunton (commented):
Awesome stuff @JacobHayes! 🚀 

On this:
> add a couple more tests and try to support serializing v1 and v2 user-defined models (not just the v1 models from either version - I think this should be possible).

Have you been able to add v2 models? I think a couple of tests using v2 models mirroring the `callable_script.py` example would suffice!

src/hera/shared/_pydantic.py
flaviuvadan: Perhaps we decide on which breaking changes should go into v6, accumulate a few of those, this one included, and then migrate to V2? 🤷🏻

flaviuvadan: Might be a good time to also move Hera to Argo 3.5

JacobHayes:
Thanks for the suggestions, circling back to this today. 🚀

I think we should be able to get v1 and v2 _user_ models serialized, but I think it'd be best (or at least easiest 😁) for the Hera internal models to be "all v1" or "all v2" until we move to v2 only.

JacobHayes:
rebased, but will follow up on the rest tomorrow

samj1912:
@JacobHayes thank you so much for the PR. I added a couple of commits to ensure that we can support pydantic v2 models from users (rather than pydantic.v1) for our hera runner.

@elliotgunton/ @flaviuvadan given that pydantic v2 support is a new "feature" I would want to start off with supporting v2 user models right off the bat, rather than asking users to use v1 models and then making a "breaking" change to support v2 models.

examples/workflows/scripts/callable_script.py
```diff
@@ -89,7 +91,7 @@ def function_kebab_object(annotated_input_value: Annotated[Input, Parameter(name
 with Workflow(name="my-workflow") as w:
     with Steps(name="my-steps") as s:
         my_function(arguments={"input": Input(a=2, b="bar", c=42)})
-        str_function(arguments={"input": Input(a=2, b="bar", c=42).json()})
+        str_function(arguments={"input": serialize(Input(a=2, b="bar", c=42))})
```
samj1912, line 94: This needs to be changed to `serialize` function in hera, which correctly handles dumping both v1 and v2 objects.

tests/test_runner.py
```diff
@@ -78,18 +78,6 @@
...
-            [{"name": "my_json_str", "value": json.dumps({"my": "dict"})}],
-            {"my": "dict"},
-            id="str-subclass-json-param-as-dict",
-        ),
-        pytest.param(
-            "tests.script_runner.parameter_inputs:str_subclass_annotated_parameter_expects_jsonstr_dict",
-            [{"name": "my_json_str", "value": json.dumps({"my": "dict"})}],
-            {"my": "dict"},
-            id="str-subclass-json-annotated-param-as-dict",
-        ),
```
samj1912, line 92: Had to remove these tests as `MyStr` is considered an arbitrary type in pydantic v2 and we don't have arbitrary types enabled even for v1. Made more sense to just get rid of this example (which we had added of our own volition in a recent PR by @elliotgunton )

samj1912 (approved):

src/hera/workflows/runner.py
```diff
@@ -64,30 +63,34 @@ def _parse(value, key, f):
...
+    return origin_type
+
 
-    return issubclass(origin_type, str)
+def _is_str_kwarg_of(key: str, f: Callable):
+    """Check if param `key` of function `f` has a type annotation of a subclass of str."""
+    type_ = _get_type(key, f)
+    if type_ is None:
+        return False
+    return issubclass(type_, str)
```
elliotgunton, line 93: With the arbitrary model problem, is it even possible for `type_` to be a subclass of `str` without being `str` itself? Don't think any code will change, just want to check my understanding

src/hera/workflows/runner.py
```diff
@@ -64,30 +63,34 @@ def _parse(value, key, f):
...
+        return type_
     origin_type = cast(type, get_origin(type_))
     if origin_type is Annotated:
-        return issubclass(get_args(type_)[0], str)
+        return get_args(type_)[0]
+    return origin_type
+
 
-    return issubclass(origin_type, str)
+def _is_str_kwarg_of(key: str, f: Callable):
```
elliotgunton, line 88: Noticed we're a bit thin on function types in this file 😅 
```suggestion
def _is_str_kwarg_of(key: str, f: Callable) -> bool:
```

tests/test_examples.py
```diff
@@ -27,7 +27,7 @@
 HERA_REGENERATE = os.environ.get("HERA_REGENERATE")
 CI_MODE = os.environ.get("CI")
 
-LOWEST_SUPPORTED_PY_VERSION = (3, 8)
+LOWEST_SUPPORTED_PY_VERSION = (3, 9)
```
elliotgunton, line 30: Why the bump? 👀

elliotgunton (commented):
Nothing blocking, looking good 🚀

JacobHayes:
This would have (or could 😁) simplify mypy things across versions: https://github.com/pydantic/pydantic/pull/9042

PR #798, opened 2023-10-03
Add missing build to mixin and QOL image pull policy builder

### LINKED ISSUE #797: image_pull_policy doesn't work in UserContainers

- [x] No

<h2>Bug report</h2>

**Describe the bug**
Using image_pull_policy in a UserContainer doesn't work. If you try, the workflow submits fine but imagePullPolicy doesn't show up in the submitted manifest. When using to_file() to see the YAML, the key shows up as "image_pull_policy" instead of imagePullPolicy and the workflow cannot be submitted with error `Failed to parse workflow: json: unknown field "image_pull_policy"`. When changing the YAML to imagePullPolicy and submitting manually, everything works as expected.

**To Reproduce**
```py
from hera.workflows import Container, Workflow, UserContainer
from hera.workflows.models import ImagePullPolicy

with Workflow(
    generate_name="sidecar-hera5-",
    entrypoint="echo",
) as w:
    echo = Container(
        name="echo",
        image="appropriate/curl",
        command=["sh", "-c"],
        args=[
            "until `curl -G 'http://127.0.0.1/' >& /tmp/out`; do echo sleep && sleep 1; done && cat /tmp/out"
        ],
        sidecars=[
            UserContainer(
                name="nginx",
                image="nginx:1.13",
                command=["nginx", "-g", "daemon off;"],
                image_pull_policy=ImagePullPolicy.always,
            )
        ],
    )
w.to_file()
w.create()

```

Result of to_yaml():
```
kind: Workflow
metadata:
  generateName: sidecar-hera5-
spec:
  entrypoint: echo
  templates:
  - container:
      args:
      - until `curl -G 'http://127.0.0.1/' >& /tmp/out`; do echo sleep && sleep 1;
        done && cat /tmp/out
      command:
      - sh
      - -c
      image: appropriate/curl
    name: echo
    sidecars:
    - command:
      - nginx
      - -g
      - daemon off;
      image: nginx:1.13
      image_pull_policy: Always
      name: nginx
```

**Expected behavior**
I expected image_pull_policy in a UserContainer to become imagePullPolicy in the YAML and show up in the final submitted manifest. Also ideally I wish Hera would throw an error or warning if the manifest contains invalid YAML so elements would not just disappear between Hera and Argo.

**Environment**
 - OS: Windows 10
 - Browser: Firefox
 - Version of Hera: 5.7.3
 - Version of Argo: 3.2.4

flaviuvadan:
**Pull Request Checklist**
- [x] Fixes #797 
- [ ] Tests added

The build call is missing when building the user container. This PR adds it


src/hera/workflows/user_container.py
```diff
@@ -23,6 +23,31 @@ class UserContainer(_ModelUserContainer):
     resources: Optional[Union[Resources, ResourceRequirements]] = None  # type: ignore[assignment]
     volumes: Optional[List[_BaseVolume]] = None
 
+    def _build_image_pull_policy(self) -> Optional[str]:
```
flaviuvadan, line 26: There's another one of these. This is not shared because it returns a `str` whereas the other returns an `ImagePullPolicy`. The type difference comes from upstream. Maybe we should edit the OpenAPI spec on this one too and only use strings?

samj1912:
@flaviuvadan can we add a test to confirm this works?

flaviuvadan:
@samj1912 added, ready for review whenever

samj1912 (approved):

PR #801, opened 2023-10-06
Fix inherited_members

### LINKED ISSUE #774: Make properties of Hera classes more discoverable
> Hey @fazilhero! You can definitely set affinity and tolerations on [containers](https://hera.readthedocs.io/en/stable/api/workflows/hera/#hera.workflows.Container). The `Container` implementation [inherits](https://github.com/argoproj-labs/hera/blob/main/src/hera/workflows/container.py#L29) from the `Template`, which allows you to set [affinity](https://github.com/argoproj-labs/hera/blob/main/src/hera/workflows/_mixins.py#L422) and [tolerations](https://github.com/argoproj-labs/hera/blob/main/src/hera/workflows/_mixins.py#L448).

```python
from hera.workflows import Container, models as m

c = Container(affinity=m.Affinity(...), tolerations=[m.Toleration(...)])
```

> @elliotgunton and @samj1912 I wonder if we can make `mkdocs` display the inherited fields on child classes as well, which would make this more clear in the docs + not require users to understand the inheritance chain of Hera objects?

_Originally posted by @flaviuvadan in https://github.com/argoproj-labs/hera/issues/773#issuecomment-1712570411_

elliotgunton:
* show_inherited_members is not a valid option
* Merge options and selection. from mkdocstrings.github.io: "Since mkdocstrings 0.19, the YAML selection key is merged into the options key."
  * We are using `mkdocstrings-0.23` in rtd builds

**Pull Request Checklist**
- [x] Fixes #774 

**Description of PR**
Currently, subclasses don't show inherited members. This option was intended to be turned on but the name was wrong.

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -118,11 +118,10 @@ plugins:
           options:
             # show_category_heading: true
             show_if_no_docstring: true
-            show_inherited_members: true
+            inherited_members: true
             filters: ["!^_.*$"]
             show_bases: false
             separate_signature: true
-          selection:
             docstring_style: google
           rendering:
             show_source: true


elliotgunton:
The good news, it works!
image

The bad news, the api reference page is now _huge_. Will see if I can split them up in the mkdocs.yml file

flaviuvadan (approved):
Great find! 🎖️

PR #805, opened 2023-10-14
Add example of how to run computation against a Dask cluster + Spark

flaviuvadan:
**Pull Request Checklist**
- [ ] Fixes #
- [ ] Tests added

**Description of PR**
I received feedback that some users would like some suggestions on how to integrate with external products such as Dask, Spark, etc. This PR introduces a first such example of using Dask in Hera. 

**Question**
Should we have an `integrations` folder to make it clear what this type of example is about? I think it's ok for now to leave it as a Hera example but perhaps in the future it can be organized in its own namespace

### DIFF
--- a/examples/workflows/dask.py
+++ b/examples/workflows/dask.py
@@ -1,0 +1,44 @@
+"""This example showcases how to run Dask within a Hera submitted Argo workflow."""
+from hera.workflows import (
+    Steps,
+    Workflow,
+    script,
+)
+
+
+@script(image="ghcr.io/dask/dask:latest")
+def dask_computation(namespace: str = "default", n_workers: int = 1) -> None:
+    import subprocess
+
+    # this is required for otherwise the dask distributed and kubernetes clients packages are not included by default
+    # ideally, you'd have a package in your organization that takes care of the following cluster details :)
+    subprocess.run(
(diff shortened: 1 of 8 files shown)


examples/workflows/dask.py
```diff
@@ -0,0 +1,44 @@
...
+)
+
+
+@script(image="ghcr.io/dask/dask:latest")
+def dask_computation(namespace: str = "default", n_workers: int = 1) -> None:
+    import subprocess
+
+    # this is required for otherwise the dask distributed and kubernetes clients packages are not included by default
+    # ideally, you'd have a package in your organization that takes care of the following cluster details :)
+    subprocess.run(
```
vikramsg, line 15: This does not look great. Would it be possible to do something like what you propose here?
https://github.com/argoproj-labs/hera/discussions/665#discussioncomment-6220596

elliotgunton:
> Should we have an integrations folder

We should rearrange the examples folder entirely IMO 😆 For Hera examples especially there's lots of features we could show off with better folder structure. Happy to leave the upstream examples as-is

flaviuvadan:
@elliotgunton I def agree. I actually tried that a bit in this PR then realized it's better suited to another PR

elliotgunton (approved):

PR #820, opened 2023-10-29
Add Pydantic `Extra.forbid` to shared base model config 

### LINKED ISSUE #819: Hera should error out when users set kwargs that are invalid
It is incredibly hard to debug the usage of a kwarg that's semantically not allowed but syntactically allowed. For example, I can use the incorrect `@script(input=...)` vs the correct `@script(inputs=...)` and Hera does not tell me I am wrong. This happens because of Hera's use of Pydantic. Hera allows extra fields on the models for very good reasons, but Hera should also protect users from this dynamic aspect unless it provides value in some way.

Upon initialization of a model Hera should check the specified kwargs vs allowed fields and raise an error when the user specification is incorrect.

None, going with status quo

flaviuvadan:
…examples

**Pull Request Checklist**
- [x] Fixes #819
- [x] Tests exists

**Description of PR**
This PR adds the Pydantic `Extra.forbid` configuration to the shared base model. This means users will no longer be faced with silent semantic errors as outlined in https://github.com/argoproj-labs/hera/issues/819. The core change is the base model. The rest of  the changes are related to fixes after introducing the new config. For instance, there were Hera examples that used `subpath` vs `sub_path`, which is exactly the sort of problem users have!

### DIFF
--- a/docs/examples/workflows-examples.md
+++ b/docs/examples/workflows-examples.md
@@ -26,7 +26,6 @@ Explore the examples through the side bar!
 | [artifacts-workflowtemplate](https://github.com/argoproj/argo-workflows/blob/master/examples/artifacts-workflowtemplate.yaml) |
 | [buildkit-template](https://github.com/argoproj/argo-workflows/blob/master/examples/buildkit-template.yaml) |
 | [ci](https://github.com/argoproj/argo-workflows/blob/master/examples/ci.yaml) |
-| [ci-workflowtemplate](https://github.com/argoproj/argo-workflows/blob/master/examples/ci-workflowtemplate.yaml) |
 | [cluster-workflow-template/clustertemplates](https://github.com/argoproj/argo-workflows/blob/master/examples/cluster-workflow-template/clustertemplates.yaml) |
 | [conditional-parameters](https://github.com/argoproj/argo-workflows/blob/master/examples/conditional-parameters.yaml) |
 | [conditionals](https://github.com/argoproj/argo-workflows/blob/master/examples/conditionals.yaml) |
--- a/docs/examples/workflows-examples.md
+++ b/docs/examples/workflows-examples.md
@@ -107,6 +107,7 @@ Explore the examples through the side bar!
 | [template-defaults](https://github.com/argoproj/argo-workflows/blob/master/examples/template-defaults.yaml) |
 | [testvolume](https://github.com/argoproj/argo-workflows/blob/master/examples/testvolume.yaml) |
 | [timeouts-step](https://github.com/argoproj/argo-workflows/blob/master/examples/timeouts-step.yaml) |
+| [webhdfs-input-output-artifacts](https://github.com/argoproj/argo-workflows/blob/master/examples/webhdfs-input-output-artifacts.yaml) |
 | [work-avoidance](https://github.com/argoproj/argo-workflows/blob/master/examples/work-avoidance.yaml) |
 | [workflow-count-resourcequota](https://github.com/argoproj/argo-workflows/blob/master/examples/workflow-count-resourcequota.yaml) |
 | [workflow-event-binding/event-consumer-workfloweventbinding](https://github.com/argoproj/argo-workflows/blob/master/examples/workflow-event-binding/event-consumer-workfloweventbinding.yaml) |
--- a/src/hera/shared/_base_model.py
+++ b/src/hera/shared/_base_model.py
@@ -28,4 +28,4 @@ class BaseModel(PyBaseModel):
         """uses smart union for matching a field's specified value to the underlying type that's part of a union."""
 
         extra = Extra.forbid
-        """forbids extra fields from being added to Hera objects. This protects users from types."""
+        """forbids extra fields from being added to Hera objects. This protects users from using the wrong kwargs."""
--- a/src/hera/shared/_base_model.py
+++ b/src/hera/shared/_base_model.py
@@ -1,5 +1,8 @@
 """Module that holds the underlying base Pydantic models for Hera objects."""
-from pydantic import BaseModel as PyBaseModel
+from pydantic import (
+    BaseModel as PyBaseModel,
+    Extra,
+)
 
 
 class BaseModel(PyBaseModel):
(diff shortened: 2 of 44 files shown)


docs/examples/workflows-examples.md
```diff
@@ -107,6 +107,7 @@ Explore the examples through the side bar!
 | [template-defaults](https://github.com/argoproj/argo-workflows/blob/master/examples/template-defaults.yaml) |
 | [testvolume](https://github.com/argoproj/argo-workflows/blob/master/examples/testvolume.yaml) |
 | [timeouts-step](https://github.com/argoproj/argo-workflows/blob/master/examples/timeouts-step.yaml) |
+| [webhdfs-input-output-artifacts](https://github.com/argoproj/argo-workflows/blob/master/examples/webhdfs-input-output-artifacts.yaml) |
```
flaviuvadan, line 110: FYI upstream example contains an extra field that is not accepted, so moved to hera examples. I looked at the OpenAPI spec on main and it does not have the field so moved until upstream is fixed

src/hera/workflows/_mixins.py
```diff
@@ -664,10 +664,12 @@ def __call__(self, *args, **kwargs) -> Union[None, Step, Task]:
...
                 )
             if isinstance(_context.pieces[-1], (Steps, Parallel)):
-                return Step(*args, template=self, **kwargs)
+                allowed_step_kwargs = Step.__fields__.keys()
+                return Step(template=self, **{k: v for k, v in kwargs.items() if k in allowed_step_kwargs})
 
             if isinstance(_context.pieces[-1], DAG):
-                return Task(*args, template=self, **kwargs)
+                allowed_task_kwargs = Task.__fields__.keys()
+                return Task(template=self, **{k: v for k, v in kwargs.items() if k in allowed_task_kwargs})
```
flaviuvadan, line 672: Required for otherwise there are extra fields that are not accepeted

src/hera/workflows/artifact.py
```diff
@@ -73,7 +73,7 @@ class Artifact(BaseModel):
     path: Optional[str] = None
     """path where the artifact should be placed/loaded from"""
 
-    recurse_mode: Optional[str] = None
+    recurse_mode: Optional[bool] = None
```
flaviuvadan, line 76: The generated object is a bool not str

src/hera/workflows/user_container.py
```diff
@@ -13,16 +13,82 @@
 from hera.workflows.resources import Resources
 from hera.workflows.volume import _BaseVolume
 
+# TODO: Mixins imports user container, so this avoids a circular import. Fix in a future release
```
flaviuvadan, line 16: mixins import the user container unfortunately. Duplicating here means we don't introduce breaking changes

src/hera/workflows/_mixins.py
elliotgunton: When would there be extra fields? Do we have them in our examples? Considering

> raise an error when the user specification is incorrect

from the issue, this won't raise an error afaict, and will silently ignore extra kwargs

src/hera/workflows/user_container.py
elliotgunton: Looking at [the template mixin](https://github.com/argoproj-labs/hera/blob/961963ba6c2866006ebd850e3c128d81a596ad17/src/hera/workflows/_mixins.py#L413-L448), I'm not sure initially how to even solve this as it's used for attribute types 🤔

src/hera/workflows/user_container.py
```diff
@@ -13,16 +13,82 @@
...
+"""
+
 
 class UserContainer(_ModelUserContainer):
     """`UserContainer` is a container type that is specifically used as a side container."""
 
-    env: Optional[List[Union[_BaseEnv, EnvVar]]] = None  # type: ignore[assignment]
-    env_from: Optional[List[Union[_BaseEnvFrom, EnvFromSource]]] = None  # type: ignore[assignment]
+    env: EnvT = None  # type: ignore[assignment]
+    env_from: EnvFromT = None  # type: ignore[assignment]
```
elliotgunton, line 44: Why does the linter have an issue with assignment here? 🤔 Are the "ignores" actually still needed?

tests/script_annotations_inputs_regression/script_annotations_artifacts_oss_old.py
```diff
@@ -1,22 +1,26 @@
...
             name="my_artifact",
             path="/tmp/file",
-            access_key_secret={"name": "my-oss-credentials", "key": "secretKey"},
+            access_key_secret=m.SecretKeySelector(name="my-oss-credentials", key="secretKey"),
+            secret_key_secret=m.SecretKeySelector(name="my-oss-credentials", key="secretKey"),
             bucket="test-bucket-name",
             create_bucket_if_not_present=True,
             endpoint="http://oss-cn-hangzhou-zmf.aliyuncs.com",
             key="test/mydirectory/",
-            lifecycle_rule={"name": "my-oss-rule", "key": "ruleKey"},
```
elliotgunton, line 18: Is it worth adding an actual lifecycle rule that matches the spec? https://github.com/argoproj-labs/hera/blob/961963ba6c2866006ebd850e3c128d81a596ad17/src/hera/workflows/models/io/argoproj/workflow/v1alpha1.py#L279-L292

elliotgunton (commented):
Looking good 💯 Only thing I want to confirm is whether errors will raise on unknown kwargs - could we add a unit test for it?

src/hera/workflows/user_container.py
flaviuvadan: > Why does the linter have an issue with assignment here? 🤔 Are the "ignores" actually still needed?

Copy / pasta, sorry, will change

src/hera/workflows/user_container.py
flaviuvadan: Naively, we could declare UserContainer as a private attr of mixins then export it in user_container as a public class

src/hera/workflows/user_container.py
flaviuvadan: Parent class has these fields that are overwritten Screenshot 2023-11-01 at 07 15 29

flaviuvadan: And you get this nonsense `src/hera/workflows/user_container.py:43: error: Incompatible types in assignment (expression has type "Union[_BaseEnv, EnvVar, List[Union[_BaseEnv, EnvVar, Dict[str, Any]]], Dict[str, Any], None]", base class "UserContainer" defined the type as "Optional[List[EnvVar]]")  [assignment]`

src/hera/workflows/_mixins.py
flaviuvadan: > When would there be extra fields? Do we have them in our examples? Considering

The **kwargs of scripts used to go into the step / task, so that contains extra fields like `image` relative to the fields of step / task. First thing I did after adding the `Extra.forbid` was to run tests. Almost _all_ of the tests failed. 

> from the issue, this won't raise an error afaict, and will silently ignore extra kwargs

Correct but it does with the addition of `Extra.forbid`

src/hera/shared/_base_model.py
```diff
@@ -10,16 +13,19 @@ class Config:
...
         arbitrary_types_allowed = True
-        """supports specifying arbitrary types for any field to support Hera object fields processing"""
+        """supports specifying arbitrary types for any field to support Hera object fields processing."""
 
         smart_union = True
-        """uses smart union for matching a field's specified value to the underlying type that's part of a union"""
+        """uses smart union for matching a field's specified value to the underlying type that's part of a union."""
+
+        extra = Extra.forbid
+        """forbids extra fields from being added to Hera objects. This protects users from types."""
```
elliotgunton, line 31:
```suggestion
        """forbids extra fields from being added to Hera objects. This protects users from using the wrong kwargs."""
```

src/hera/workflows/_mixins.py
elliotgunton: Ah I see, TIL!

elliotgunton (commented):
LGTM, tiny word change first though

elliotgunton (approved):
:shipit:

PR #825, opened 2023-10-31
Unify imports of hera.workflows and hera.workflows.models

### LINKED ISSUE #800: Challenging to tell when to use `models` vs not. Can we unify the objects into a single init?
It is currently challenging to know when to import the `models` module and use objects from that with Hera. Some users might expect that `import hera.workflows.*` gives them everything they need. However, imagine using `ttl_strategy`, which is easily available on a container but the GC strategy import must come from `models`, which is not as obvious. We should consider unifying the imports somehow

flaviuvadan:
**Pull Request Checklist**
- [x] Fixes #800
- [ ] Tests added

It's challenging to know when to use `from hera.workflows` or `from hera.workflows.models`. This PR unifies the imports of the two into `hera.workflows` so that users can import both.

## Questions for reviewers

Is this the right approach?


samj1912:
+1 to everything Elliot said. I'm not sold on this idea either and Hera classes behave very differently to normal generated pydantic classes. I feel like this will put us into a hole wrt backwards incompatible changes.

hermesribeiro:
I do agree that mixing pydantic models with `hera.workflows` classes can cause some issues , such as compatibility breaks down the road in case a new hera class with the same name is implemented in the `workflows` folder.

However the problem still remains that there is no direct way to know which Argo object is available through a hera tailored class and which is only available as a pydantic model (or even just as an Enum). Since hera's proposal is to be an intuitive and easy to use yaml manifest replacement, ideally I'd like to be able to import and use any object in the yaml syntax in one unified way.

>... create actual Hera classes for all the models that we think should have our syntactic sugar/helper functions (which could be all of them, could be some of them).

Even if there is no syntactic sugar or helper to add, could it be for all of them? An idea here, maybe you could extend the current PR to generate dummy classes for the `autogen_modules`.

flaviuvadan:
Closing for now, let's discuss in the maintainers meeting. Maybe there's a world in which we autogen Hera classes "prepared" for syntactic sugar

PR #859, opened 2023-11-10
Fix _is_str_kwarg_of for annotated parameters

### LINKED ISSUE #855: Script annotations + runner for annotated strings breaks for json strs
Discussed with @samj1912 - the runner currently does some implicit json loading of inputs that we'd like to review/remove some of and move the control of loading parameters (e.g. given a json str or yaml str, choose whether to load it via `json.loads` or `yaml.safe_load`). Marked the PR as draft, and will make the change in another PR on top of this one.

Also, this code doesn't deal well with `Annotated` strings, resulting in different behaviour for `my_str: str` and `my_str: Annotated[str, Parameter(...)]` in script parameters https://github.com/argoproj-labs/hera/blob/04eca946cd3ff83beeca536c50c40db8d23ff193/src/hera/workflows/runner.py#L77

_Originally posted by @elliotgunton in https://github.com/argoproj-labs/hera/issues/850#issuecomment-1803751319_

elliotgunton:
**Pull Request Checklist**
- [x] Fixes #855 
- [x] Tests added

**Description of PR**
Currently, annotations break the `_is_str_kwarg_of` function. Adds logic to check whether it's a built-in `str`, an `Annotated` str or a simple subclass of `str`. It also fixes/makes explicit the logic behind doing a json parse of the value if it is untyped.

### DIFF
--- a/src/hera/workflows/runner.py
+++ b/src/hera/workflows/runner.py
@@ -78,7 +78,14 @@ def _is_str_kwarg_of(key: str, f: Callable):
     if not type_:
         return True
     try:
-        return issubclass(type_, str)
+        if get_origin(type_) is None:
+            return issubclass(type_, str)
+
+        origin_type = cast(type, get_origin(type_))
+        if origin_type is Annotated:
+            return issubclass(get_args(type_)[0], str)
+
+        return issubclass(origin_type, str)
     except TypeError:
         # If this happens then it means that the annotation is a more complex type annotation
         # and may be interpretable by the Hera runner
--- a/src/hera/workflows/runner.py
+++ b/src/hera/workflows/runner.py
@@ -75,21 +75,19 @@ def _parse(value, key, f):
 def _is_str_kwarg_of(key: str, f: Callable):
     """Check if param `key` of function `f` has a type annotation of a subclass of str."""
     type_ = inspect.signature(f).parameters[key].annotation
-    if not type_:
-        return True
-    try:
-        if get_origin(type_) is None:
-            return issubclass(type_, str)
 
-        origin_type = cast(type, get_origin(type_))
-        if origin_type is Annotated:
-            return issubclass(get_args(type_)[0], str)
-
-        return issubclass(origin_type, str)
-    except TypeError:
-        # If this happens then it means that the annotation is a more complex type annotation
-        # and may be interpretable by the Hera runner
+    if type_ is inspect.Parameter.empty:
+        # Untyped args are interpreted according to json spec
+        # ie. we will try to load it via json.loads in _parse
         return False
+    if get_origin(type_) is None:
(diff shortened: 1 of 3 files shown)


src/hera/workflows/runner.py
```diff
@@ -78,7 +78,14 @@ def _is_str_kwarg_of(key: str, f: Callable):
...
-        return issubclass(type_, str)
+        if get_origin(type_) is None:
+            return issubclass(type_, str)
+
+        origin_type = cast(type, get_origin(type_))
+        if origin_type is Annotated:
+            return issubclass(get_args(type_)[0], str)
+
+        return issubclass(origin_type, str)
     except TypeError:
```
elliotgunton, line 89: I'm not sure this is needed anymore?

src/hera/workflows/runner.py
```diff
@@ -75,14 +75,19 @@ def _parse(value, key, f):
 def _is_str_kwarg_of(key: str, f: Callable):
     """Check if param `key` of function `f` has a type annotation of a subclass of str."""
     type_ = inspect.signature(f).parameters[key].annotation
-    if not type_:
-        return True
-    try:
-        return issubclass(type_, str)
-    except TypeError:
```
elliotgunton, line 82: Confirmed this error is only raised if the type annotation is an object rather than a class

```py
@script()
def object_type_parameter(my_anything: object()) -> Any:
    """Trying to pass an object as the type instead of class to raise TypeError"""
    return my_anything
```

I don't think this really makes sense as Python code so my vote is to remove the TypeError code. Otherwise I can add a function like above and test it

src/hera/workflows/runner.py
```diff
@@ -75,14 +75,19 @@ def _parse(value, key, f):
 def _is_str_kwarg_of(key: str, f: Callable):
     """Check if param `key` of function `f` has a type annotation of a subclass of str."""
     type_ = inspect.signature(f).parameters[key].annotation
-    if not type_:
-        return True
```
elliotgunton, line 79: This code was impossible to hit afaict

src/hera/workflows/runner.py
```diff
@@ -75,14 +75,19 @@ def _parse(value, key, f):
...
-    try:
-        return issubclass(type_, str)
-    except TypeError:
-        # If this happens then it means that the annotation is a more complex type annotation
-        # and may be interpretable by the Hera runner
+
+    if type_ is inspect.Parameter.empty:
+        # Untyped args are interpreted according to json spec
+        # ie. we will try to load it via json.loads in _parse
         return False
```
samj1912, line 82: I believe untyped args should be left as is, since Argo params are all strings

flaviuvadan (approved):
LGTM and @elliotgunton the reasoning you outlined sounds great

PR #874, opened 2023-11-23
Add support for modifying field properties prior to model generation

### LINKED ISSUE #869: HTTPGetAction "port: int" will output as string in yaml

- [x] No

<h2>Bug report</h2>

**Describe the bug**
When defining Probe inside of `UserContainer`

```
startup_probe=Probe(
    http_get=HTTPGetAction(path="/startup", port=9801, scheme="HTTP"),
    timeoutSeconds=10,
    periodSeconds=1,
    successThreshold=1,
    failureThreshold=60
)
```

The output of startupProbe will look like this

```
  startupProbe:
    httpGet:
      path: /startup
      port: '9801'
      scheme: HTTP
    timeoutSeconds: 10
    periodSeconds: 1
    successThreshold: 1
    failureThreshold: 60
 ```  
 
 Error message when running the workflow
 
 ```
 Pod "rg-custom-drill-daily-s4gj9-pod-179047205" is invalid: [spec.containers[2].lifecycle.preStop.httpGet.port: Invalid value: "9091": must contain at least one letter (a-z), spec.containers[2].livenessProbe.httpGet.port: Invalid value: "9801": must contain at least one letter (a-z), spec.containers[2].startupProbe.httpGet.port: Invalid value: "9801": must contain at least one letter (a-z)]
 ```

**Expected behavior**
Expected the port is not string

```
  startupProbe:
    httpGet:
      path: /startup
      port: 9801
      scheme: HTTP
    timeoutSeconds: 10
    periodSeconds: 1
    successThreshold: 1
    failureThreshold: 60
 ```  

**Environment**
- Hera Version: 5.9.0
- Python: 3.9
- Argo Workflow: 3.5.0-rc1

flaviuvadan:
**Pull Request Checklist**
- [x] Fixes #869
- [x] Tests added (we already have tests / examples that compare upstream and Hera YAMLs that include HTTPGetAction)

#869 reports how the field `IntOrString` incorrectly renders the port specification into a string that's not accepted by K8s. This PR adds a spec modification strategy that takes the remote Argo OpenAPI JSON specification and modifies the field from `IntOrString` to `integer`. This way the correct model is generated, and users know an integer is supposed to be used. Ideally, we'd fix these upstream but this at least unblocks current Hera users.

### DIFF
--- a/src/hera/workflows/models/io/k8s/api/core/v1.py
+++ b/src/hera/workflows/models/io/k8s/api/core/v1.py
@@ -1526,7 +1526,7 @@ class HTTPGetAction(BaseModel):
         description=("Custom headers to set in the request. HTTP allows repeated headers."),
     )
     path: Optional[str] = Field(default=None, description="Path to access on the HTTP server.")
-    port: intstr.IntOrString = Field(
+    port: int = Field(
         ...,
         description=(
             "Name or number of the port to access on the container. Number must be in"
--- a/docs/examples/workflows/upstream/daemon_nginx.md
+++ b/docs/examples/workflows/upstream/daemon_nginx.md
@@ -66,7 +66,7 @@ The upstream example can be [found here](https://github.com/argoproj/argo-workfl
           readinessProbe:
             httpGet:
               path: /
-              port: '80'
+              port: 80
             initialDelaySeconds: 2
             timeoutSeconds: 1
         daemon: true
--- a/examples/workflows/upstream/daemon-nginx.yaml
+++ b/examples/workflows/upstream/daemon-nginx.yaml
@@ -10,7 +10,7 @@ spec:
       readinessProbe:
         httpGet:
           path: /
-          port: '80'
+          port: 80
         initialDelaySeconds: 2
         timeoutSeconds: 1
     daemon: true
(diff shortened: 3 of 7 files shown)


src/hera/workflows/models/io/k8s/api/core/v1.py
```diff
@@ -1526,7 +1526,7 @@ class HTTPGetAction(BaseModel):
         description=("Custom headers to set in the request. HTTP allows repeated headers."),
     )
     path: Optional[str] = Field(default=None, description="Path to access on the HTTP server.")
-    port: intstr.IntOrString = Field(
+    port: int = Field(
         ...,
         description=(
             "Name or number of the port to access on the container. Number must be in"
```
elliotgunton, line 1532: Interesting that the description is "name or number" 🤔

elliotgunton (approved):
I'm conscious of `scripts/spec.py` being a bit hacky at this point, but it's to address problems upstream so 🤷‍♂️  Also, the fix in the underlying generated code is probably a smart union to try to use an int if it looks like an int. But that's for another PR

src/hera/workflows/models/io/k8s/api/core/v1.py
flaviuvadan: Yea.. and all other ports I looked at, on other objects, are integers. Perhaps it's a backwards compatibility thing that upstream tried to address. I lack the context

PR #875, opened 2023-11-23
Fix incorrect aggregation of volume mount and dynamic volumes created via Hera

flaviuvadan:
**Pull Request Checklist**
- [x] Fixes #
- [x] Tests added

**Description of PR**
Hera supports building dynamic volumes via the `Volume` object. That creates PVCs + PVs dynamically. Users have the liberty to mount these volumes wherever they want. However, if users have a volume mount _already_, then the dynamic volumes are not combined with mounts correctly. That happens because of the incorrect conditions inside of `build_volume_mounts`. This PR fixes that


src/hera/workflows/_mixins.py
```diff
@@ -539,25 +539,25 @@ def _build_volume_mounts(self) -> Optional[List[VolumeMount]]:
...
         if result is None and self.volume_mounts is None:
             return None
         elif result is None and self.volume_mounts is not None:
             return self.volume_mounts
+        elif result is not None and self.volume_mounts is None:
+            return result
 
-        mounts = cast(List[VolumeMount], self.volume_mounts) or [] + cast(List[VolumeMount], result) or []
-        return mounts or None
+        return self.volume_mounts + result
```
elliotgunton, line 560: Think this still needs a cast to pass lint checks

src/hera/workflows/_mixins.py
```diff
@@ -539,25 +539,25 @@ def _build_volume_mounts(self) -> Optional[List[VolumeMount]]:
...
-        else:
+        if self.volumes is None:
             volumes = []
+        else:
+            volumes = self.volumes if isinstance(self.volumes, list) else [self.volumes]
 
         result = (
             None
-            if self.volumes is None
+            if volumes is []
```
elliotgunton, line 549: Using `is` here is ringing alarm bells as it's only for `None` really (should use `==` or directly check on `volumes`). We could use a `not volumes` or swap the if/else around to use `if volumes`?
```suggestion
            if not volumes
```

flaviuvadan:
@elliotgunton applied your suggestions

elliotgunton (approved):
:shipit:

PR #876, opened 2023-11-23
Implement get_parameter API on IO mixin and add tests

### LINKED ISSUE #816: get_parameter/artifact for template invocators (Steps/DAG)
Passing Steps/DAG inputs directly through to substeps/Tasks requires the expression syntax string or `expr`

DAGs/Steps to expose `get_parameter`/`get_artifact` for use in child nodes

Currently, for an example, see [callable_dag.py](https://github.com/argoproj-labs/hera/blob/b13811ab8548bd4a169c330c8dd777ce0794e90e/examples/workflows/callable_dag.py#L13-L15
), but instead passing the value through directly:

```py
    with DAG(name="my-dag", inputs=Parameter(name="my-dag-input")) as my_dag:
        hello(name="hello", arguments={"name": "{{inputs.parameters.my-dag-input}}"})
```

Preferred would be:

```py
    with DAG(name="my-dag", inputs=Parameter(name="my-dag-input")) as my_dag:
        hello(name="hello", arguments=[my_dag.get_parameter("my-dag-input").with_name("name")])
```

This will help us get away from requiring expr knowledge, but can be used for now.

**Additional context**

We have some work in https://github.com/argoproj-labs/hera/pull/778 for output references, should think about how these can work together, also the callable syntax in #763

flaviuvadan:
**Pull Request Checklist**
- [x] Fixes #816
- [x] Tests added

Implements the `get_parameter` API on the `IOMixin`. This allows clients that need parameters in `arguments` fields to get the parameter object from DAGs, Steps, etc. This helps avoid the need to write `{{inputs.parameters.whatever}}` explicitly. `get_parameter` _assumes_ that clients want to use a parameter as input, so the value field is set accordingly

### DIFF
--- a/examples/workflows/callable_dag_with_param_get.py
+++ b/examples/workflows/callable_dag_with_param_get.py
@@ -1,0 +1,31 @@
+from typing_extensions import Annotated
+
+from hera.workflows import DAG, Parameter, Workflow, script
+
+
+@script(constructor="runner")
+def hello_with_output(name: str) -> Annotated[str, Parameter(name="output-message")]:
+    return "Hello, {name}!".format(name=name)
+
+
+with Workflow(
+    generate_name="callable-dag-",
+    entrypoint="calling-dag",
+) as w:
+    with DAG(
+        name="my-dag-with-outputs",
+        inputs=Parameter(name="my-dag-input"),
+        outputs=Parameter(
+            name="my-dag-output",
+            value_from={
+                "parameter": "{{hello.outputs.parameters.output-message}}"
+            },  # Don't think we can improve this?
+        ),
+    ) as my_dag:
+        hello_with_output(name="hello", arguments={"name": f"hello {my_dag.get_parameter('my-dag-input')}"})
+
+    with DAG(name="calling-dag") as d:
+        t1 = my_dag(name="call-1", arguments={"my-dag-input": "call-1"})
+        # Here, t1 is a Task from the called dag, so get_parameter is called on the Task to get the output parameter! 🚀
+        t2 = my_dag(name="call-2", arguments=t1.get_parameter("my-dag-output").with_name("my-dag-input"))
+        t1 >> t2
--- a/examples/workflows/callable_dag_with_param_get.py
+++ b/examples/workflows/callable_dag_with_param_get.py
@@ -22,6 +22,7 @@ with Workflow(
             },
         ),
     ) as my_dag:
+        # Here, get_parameter searches through the *inputs* of my_dag
         hello_with_output(name="hello", arguments={"name": f"hello {my_dag.get_parameter('my-dag-input')}"})
 
     with DAG(name="calling-dag") as d:
--- a/examples/workflows/callable_dag_with_param_get.py
+++ b/examples/workflows/callable_dag_with_param_get.py
@@ -17,9 +17,7 @@ with Workflow(
         inputs=Parameter(name="my-dag-input"),
         outputs=Parameter(
             name="my-dag-output",
-            value_from={
-                "parameter": "{{hello.outputs.parameters.output-message}}"
-            },
+            value_from={"parameter": "{{hello.outputs.parameters.output-message}}"},
         ),
     ) as my_dag:
         # Here, get_parameter searches through the *inputs* of my_dag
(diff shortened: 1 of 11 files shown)


examples/workflows/upstream/parallelism_nested.py
```diff
@@ -23,7 +23,7 @@
             one_job(
                 name="seq-step",
                 arguments=[
-                    Parameter(name="parallel-id", value="{{inputs.parameters.parallel-id}}"),
+                    seq_worker.get_parameter("parallel-id"),
                     Parameter(name="seq-id", value="{{item}}"),
                 ],
                 with_param="{{inputs.parameters.seq-list}}",
```
elliotgunton, line 29: Can these `with_param`s change as well?

src/hera/workflows/_mixins.py
```diff
@@ -274,6 +274,32 @@ class IOMixin(BaseMixin):
     inputs: InputsT = None
     outputs: OutputsT = None
 
+    def get_parameter(self, name: str) -> Parameter:
+        """Finds and returns the parameter with the supplied name.
```
elliotgunton, line 278: I was writing a comment about output parameters from DAGs to confirm the logic, and ended up finding it works really well! Leaving a code example here - maybe we can add it as an example alongside callable_dag.py? Maybe `callable_dag_with_outputs.py`?

```py
@script(constructor="runner")
def hello_with_output(name: str) -> Annotated[str, Parameter(name="output-message")]:
    return "Hello, {name}!".format(name=name)

with Workflow(
    generate_name="callable-dag-",
    entrypoint="calling-dag",
) as w:
    with DAG(
        name="my-dag-with-outputs",
        inputs=Parameter(name="my-dag-input"),
        outputs=Parameter(
            name="my-dag-output",
            value_from={"parameter": "{{hello.outputs.parameters.output-message}}"},  # Don't think we can improve this?
        ),
    ) as my_dag:
        # I believe this works within the f-string - get_parameter is called on the IOMixin to get the input parameter
        hello_with_output(name="hello", arguments={"name": f"hello {my_dag.get_parameter('my-dag-input')}"})

    with DAG(name="calling-dag") as d:
        t1 = my_dag(name="call-1", arguments={"my-dag-input": "call-1"})
        # Here, t1 is a Task from the called dag, so get_parameter is called on the Task to get the output parameter! 🚀 
        t2 = my_dag(name="call-2", arguments={"my-dag-input": t1.get_parameter("my-dag-output")})
        t1 >> t2
```

elliotgunton (commented):
Looks awesome! Wondering if we can add the dual uses of `get_parameter` as an example? Not approving just yet as auto merge is on 😆

examples/workflows/callable_dag_with_param_get.py
```diff
@@ -0,0 +1,31 @@
...
+    entrypoint="calling-dag",
+) as w:
+    with DAG(
+        name="my-dag-with-outputs",
+        inputs=Parameter(name="my-dag-input"),
+        outputs=Parameter(
+            name="my-dag-output",
+            value_from={
+                "parameter": "{{hello.outputs.parameters.output-message}}"
+            },  # Don't think we can improve this?
```
elliotgunton, line 22: This comment snuck in
```suggestion
            },
```

examples/workflows/callable_dag_with_param_get.py
```diff
@@ -0,0 +1,31 @@
...
+        ),
+    ) as my_dag:
+        hello_with_output(name="hello", arguments={"name": f"hello {my_dag.get_parameter('my-dag-input')}"})
```
elliotgunton, line 25: Depending on if we're keeping the explanation comments of get_parameter, would suggest:

```suggestion
        # Here, get_parameter searches through the *inputs* of my_dag
        hello_with_output(name="hello", arguments={"name": f"hello {my_dag.get_parameter('my-dag-input')}"})
```

examples/workflows/callable_dag_with_param_get.py
```diff
@@ -0,0 +1,31 @@
...
+
+    with DAG(name="calling-dag") as d:
+        t1 = my_dag(name="call-1", arguments={"my-dag-input": "call-1"})
+        # Here, t1 is a Task from the called dag, so get_parameter is called on the Task to get the output parameter! 🚀
```
elliotgunton, line 29: Do we want the comment?

elliotgunton (commented):
Just need a cleanup on the comments (and regen code to pass ci)

examples/workflows/callable_dag_with_param_get.py
elliotgunton: I'm going with yes lol

elliotgunton (approved):

PR #877, opened 2023-11-23
Remap `image_pull_policy` field on `Container` from `ImagePullPolicy` enum to `str`

### LINKED ISSUE #784: Validation error when using Stored Template but job works fine

- [x] No

- [ ] running Hera code without submitting to Argo (e.g. when exporting to YAML)

<h2>Bug report</h2>

**Describe the bug**

_Error log if applicable_:
```
  File "/usr/lib/python3.8/runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/home/vikramsg/Projects/pipeline/hera_sr.py", line 299, in <module>
    create_workflow()
  File "/home/vikramsg/Projects/pipeline/hera_sr.py", line 295, in create_workflow
    w.create()
  File "/home/vikramsg/Projects/pipeline/pipe-src/.venv/lib/python3.8/site-packages/hera/workflows/workflow.py", line 385, in create
    wf = self.workflows_service.create_workflow(
  File "/home/vikramsg/Projects/pipeline/pipe-src/.venv/lib/python3.8/site-packages/hera/workflows/service.py", line 847, in create_workflow
    return Workflow(**resp.json())
  File "pydantic/main.py", line 341, in pydantic.main.BaseModel.__init__
pydantic.error_wrappers.ValidationError: 1 validation error for Workflow
status -> storedTemplates -> namespaced/s3-copy-templates-v6/copy-processing-data -> container -> imagePullPolicy
  value is not a valid enumeration member; permitted: 'Always', 'IfNotPresent', 'Never' (type=type_error.enum; enum_values=[<ImagePullPolicy.always: 'Always'>, <ImagePullPolicy.if_not_present: 'IfNotPresent'>, <ImagePullPolicy.never: 'Never'>])

```

**To Reproduce**
```py
def _s3_download_processing_data() -> Step:
    return Step(
        name="s3-download-processing-data",
        template_ref=m.TemplateRef(
            name="s3-copy-templates-v6",
            template="copy-processing-data",
        ),
    )
```

**Environment**
 - OS: [Kubuntu]
 - Browser: [firefox]
 - Version of Hera: [5.7.1]
 - Version of Argo: [3.4.4]

**Additional context**
I am trying to submit a job using Hera that references a stored template on Argo. The issue is that Hera seems to complain while linting. Remarkably, the job still submits fine, and right now seems to work as intended. 

It is pretty difficult to debug however, since the problem occurs at `Workflow(**resp.json())` which is a huge `json` argument. I am assuming it is ingesting the template and validating it, but finds an issue with it?

Anyway to fix this? Alternatively, anyway to debug this to get to why this is happening?

flaviuvadan:
**Pull Request Checklist**
- [x] Fixes #784
- [x] Tests added

**Description of PR**
When an autogenerated object is created through the service the service returns auto generated objects. This means the services relies on autogenerated code _only_. #784 documents how creating an autogen `Workflow` object fails because the value of `Always` does not fit the `Template.container.imagePullPolicy` field. This happens because the returned value is of type `str` (from the Argo Server) while the Python value is an enum of type `ImagePullPolicy`. This is caused by an inconsistency in the OpenAPI specification. Some fields use `str` for image pull policy (look at Events, User Container, etc) while Container uses `ImagePullPolicy`. This PR changes the field in the OpenAPI schema specification so that this Container field is also a `str`. This should solve #784. Our Pydantic configuration is set to use enum values, which means that users who rely on the autogen Container object and pass the ImagePullPolicy object in will have Pydantic automatically extract the `.value` field of that enum. The Hera objects already accommodate both str and ImagePullPolicy, so no breaking change there.

### DIFF
--- a/scripts/spec.py
+++ b/scripts/spec.py
@@ -3,7 +3,7 @@
 import json
 import logging
 import sys
-from typing import Dict, List, Set, Tuple
+from typing import Dict, List, Optional, Set, Tuple, Union
 
 import requests
 
@@ -56,17 +56,26 @@ for definition, optional_fields in DEFINITION_TO_OPTIONAL_FIELDS.items():
 # these are specifications of objects with fields that are marked to have a union type of IntOrString. However, K8s
 # only accepts one or the other, unfortunately. Here, we remap those fields from their respective `$ref`s, which
 # contain a specific type, to another type. The mapping is from the `$ref` to a tuple of the existing type and the
-# new type. The dictionary model is:
+# new type. While the first piece of the mapping tuple must be defined the second one is optional - if the second
+# part of the tuple is None then the field will be removed. The dictionary model is:
 # { object name: { field name: ( ( existing field, existing value ) , ( new field, new value ) ) } }
-INT_OR_STRING_FIELD_REMAPPING: Dict[str, Dict[str, Tuple[Tuple[str, str], Tuple[str, str]]]] = {
+FIELD_REMAPPINGS: Dict[
+    str, Dict[str, Tuple[Tuple[str, Union[str, List[str]]], Optional[Tuple[str, Union[str, List[str]]]]]]
+] = {
     "io.k8s.api.core.v1.HTTPGetAction": {
         "port": (
             ("$ref", "#/definitions/io.k8s.apimachinery.pkg.util.intstr.IntOrString"),
             ("type", "integer"),
         ),
     },
+    "io.k8s.api.core.v1.Container": {
+        "imagePullPolicy": (
+            ("enum", None),
+            None,
+        )
+    },
 }
-for obj_name, field in INT_OR_STRING_FIELD_REMAPPING.items():
+for obj_name, field in FIELD_REMAPPINGS.items():
     try:
         curr_field = spec["definitions"][obj_name]
     except KeyError as e:
@@ -94,7 +103,12 @@ for obj_name, field in INT_OR_STRING_FIELD_REMAPPING.items():
 
         # get the tuple of the existing field and value, and the new field and value
         existing_field, existing_value = field[property_to_change][0]
-        new_field, new_value = field[property_to_change][1]
+        if field[property_to_change][1] is None:
+            # if the second one is absent it means we want to delete the existing field
+            del curr_property[existing_field]
+            continue
+        else:
+            new_field, new_value = field[property_to_change][1]
 
         # check that the existing field and value are the same as the current field and value
         assert curr_property[existing_field] == existing_value, (
--- a/scripts/spec.py
+++ b/scripts/spec.py
@@ -122,6 +122,21 @@ for obj_name, field in FIELD_REMAPPINGS.items():
         if existing_field != new_field:
             del curr_property[existing_field]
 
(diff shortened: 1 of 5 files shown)


flaviuvadan:
Turns out autogen removes the `ImagePullPolicy` definition altogether, which is not the intended result..

flaviuvadan:
> Turns out autogen removes the `ImagePullPolicy` definition altogether, which is not the intended result..

Fixed

flaviuvadan:
@vikramsg FYI ^

scripts/spec.py
```diff
@@ -56,17 +56,26 @@
...
     "io.k8s.api.core.v1.HTTPGetAction": {
         "port": (
             ("$ref", "#/definitions/io.k8s.apimachinery.pkg.util.intstr.IntOrString"),
             ("type", "integer"),
         ),
     },
+    "io.k8s.api.core.v1.Container": {
+        "imagePullPolicy": (
+            ("enum", None),
+            None,
```
elliotgunton, line 74: What's this `None` doing here? 🤔

scripts/spec.py
```diff
@@ -56,17 +56,26 @@
 # these are specifications of objects with fields that are marked to have a union type of IntOrString. However, K8s
 # only accepts one or the other, unfortunately. Here, we remap those fields from their respective `$ref`s, which
 # contain a specific type, to another type. The mapping is from the `$ref` to a tuple of the existing type and the
-# new type. The dictionary model is:
+# new type. While the first piece of the mapping tuple must be defined the second one is optional - if the second
+# part of the tuple is None then the field will be removed. The dictionary model is:
 # { object name: { field name: ( ( existing field, existing value ) , ( new field, new value ) ) } }
-INT_OR_STRING_FIELD_REMAPPING: Dict[str, Dict[str, Tuple[Tuple[str, str], Tuple[str, str]]]] = {
+FIELD_REMAPPINGS: Dict[
+    str, Dict[str, Tuple[Tuple[str, Union[str, List[str]]], Optional[Tuple[str, Union[str, List[str]]]]]]
```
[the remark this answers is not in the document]
flaviuvadan: Yea... I was very reluctant to add this type, especially as I realized it's growing. Yea, we'll have to refactor this a bit in the future.

scripts/spec.py
flaviuvadan: First part of the tuple says "original field, original field value" second one says "new field, new field value". New field value of None means delete the original field. So, this is removing the enum from Container, allowing codegen to put a string in place of the Container image pull policy enum, and stop generating the ImagePullPolicy enum altogether.

Later in the script you see we have to add a specification for IPP manually for otherwise it's not fully backwards compatible, even if it's not used by Container!

elliotgunton: Ah of course, I was misreading the outer tuple. Makes sense (almost!), though I would suggest a refactor to helper classes if you have time/want to in a separate PR

elliotgunton (approved):
Maybe we add an issue for us to revisit/refactor `spec.py`? For now: :shipit:

PR #881, opened 2023-11-28
Revert "Revert "Add Pydantic `Extra.forbid` to shared base model config ""

flaviuvadan:
Reverts argoproj-labs/hera#880


src/hera/shared/_base_model.py
```diff
@@ -10,16 +13,19 @@ class Config:
...
 
         arbitrary_types_allowed = True
-        """supports specifying arbitrary types for any field to support Hera object fields processing"""
+        """supports specifying arbitrary types for any field to support Hera object fields processing."""
 
         smart_union = True
-        """uses smart union for matching a field's specified value to the underlying type that's part of a union"""
+        """uses smart union for matching a field's specified value to the underlying type that's part of a union."""
+
+        extra = Extra.forbid
```
elliotgunton, line 30: Maybe we can turn it off when interacting with the server to solve the underlying issue?

flaviuvadan: Yea, probably.. I am surprised that the fields on the return payload _may_ not match the fields on the autogen code though :(

PR #886, opened 2023-12-06
Add hera CLI and implement `generate yaml`.

DanCardin:
**Pull Request Checklist**
- [x] Tests added

**Description of PR**
Per design outlined in https://github.com/argoproj-labs/hera/issues/670 and https://github.com/argoproj-labs/hera/issues/711

My personal usecase ideally requires generating for a whole folder at a time, but I only need output to stdout (Argo sidecar cmp plugin). Anything not in that critical path, I'm happy to remove if it's more challenging to agree on everything all at once.

* Adds the `hera` cli console script
  * optional dependency, informs the user of the required extra if they use it
* implements `generate yaml`

Notes:
* Notable difference from the design outlined in the above issues: rather than `--from` as a required "option", I opted for it to be a positional argument for now. I can certainly change this back to `--from` if you feel strongly against it
* I will certainly add tests once there's an indication that this looks roughly like what you'd want/accept
* All of the help text/error messaging is just a first draft, feel free to suggest different prose.
* I've implemented both `file.py -> stdout`, `folder/ to stdout`, `file.py -> foo.yaml`, and `folder/ -> to_folder/`
  * I think folder to folder is perhaps the most opinionated, in that it makes the decision that all source files get converted with the same name but a yaml suffix. It seems logical to me, but I could just as easily disallow this option.
* I've chosen to use Typer here, because it was what https://github.com/argoproj-labs/hera/pull/674 originally chose to use. I have a personal (very biased) preference towards [Cappa](https://cappa.readthedocs.io/), and I think it matches the style of Hera itself really well. With that said, Typer is the obviously more established option, so I'm not at all expecting to get buy-in to use Cappa here (just putting it out there 🤣).


src/hera/_cli/main.py
```diff
@@ -0,0 +1,25 @@
...
+
+def main():
+    try:
+        import typer
+    except ImportError:
+        raise ImportError("Use of the `hera` CLI tool requires installing the 'cli' extra, `pip install hera[cli]`.")
+
+    app = typer.Typer()
+
+    generate_app = typer.Typer()
```
DanCardin, line 15: this is admittedly kind of weird. Typer, because of its click origins, (imo) makes testing more challenging because it takes over your function when using their normal decoration pattern. I did this so that `generate_yaml` could be directly called in tests for the majority of the testing, rather than needing to either use their test runner or define extra functions to enable the same thing.

I can certainly undo this and use the more natural decorator style, although i anticipate that blowing up the numbers of boilerplate files and functions required to achieve the same thing

flaviuvadan:
Thanks for the contribution @DanCardin! Also, first time contributor 🥳 Will review the PR this week

flaviuvadan (commented):
This looks great @DanCardin!

I don't have any major meaningful feedback + saw Sam's feedback on Cappa so I will let Sam + Elliot provide the major input here

src/hera/_cli/__init__.py
```diff
@@ -0,0 +1 @@
+"""Hera's command line interface."""
```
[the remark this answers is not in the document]
DanCardin: I was just copying the base package @elliotgunton used in the original closed PR. At least in the short term, it would make sense to me to define it as a private package at least until it's more set in stone. I'm otherwise very ambivalent about it though.

DanCardin:
Added note about the CLI being experimental. Also here's some screenshots of what it looks like currently:

Screenshot 2023-12-15 at 1 09 07 PM
Screenshot 2023-12-15 at 1 09 16 PM
Screenshot 2023-12-15 at 1 09 25 PM

src/hera/__main__.py
```diff
@@ -0,0 +1,23 @@
...
+        import cappa
+    except ImportError:
+        raise ImportError("Use of the `hera` CLI tool requires installing the 'cli' extra, `pip install hera[cli]`.")
+
+    import rich
+
+    from hera._cli.base import Hera
+
+    rich.print(
+        "[yellow bold]warning: The `hera` CLI is a currently work-in-progress, subject to change at any time![/yellow bold]"
```
elliotgunton, line 16:
```suggestion
        "[yellow bold]warning: The `hera` CLI is a work-in-progress, subject to change at any time![/yellow bold]"
```

src/hera/_cli/base.py
```diff
@@ -0,0 +1,50 @@
...
+    help="Generate yaml from python Workflow definitions.",
+    invoke="hera._cli.generate.yaml.generate_yaml",
+)
+class GenerateYaml:
+    from_: Annotated[
+        Path,
+        Arg(
+            value_name="from",
+            help=(
+                "The path from which the yaml is generated. This can be a file, "
```
elliotgunton, line 32:
```suggestion
                "The path from which the yaml is generated. This can be a file "
```

src/hera/_cli/base.py
```diff
@@ -0,0 +1,50 @@
...
+                "or a folder. When a folder is provided, all python files in the "
```
elliotgunton, line 33:
```suggestion
                "or a folder. When a folder is provided, all Python files in the "
```

src/hera/_cli/generate/yaml.py
```diff
@@ -0,0 +1,102 @@
...
+import sys
+from pathlib import Path
+from typing import Generator
+
+from hera._cli.base import GenerateYaml
+from hera.workflows.workflow import Workflow
+
+
+def generate_yaml(options: GenerateYaml):
+    """Generate yaml from python Workflow definitions.
```
elliotgunton, line 15:
```suggestion
    """Generate yaml from Python Workflow definitions.
```

src/hera/_cli/generate/yaml.py
```diff
@@ -0,0 +1,102 @@
...
+
+    If the provided path is folder, generates yaml for all python files containing `Workflow`s
```
elliotgunton, line 17:
```suggestion
    If the provided path is a folder, generates yaml for all Python files containing `Workflow`s
```

tests/cli/examples/single_workflow.py
```diff
@@ -0,0 +1,3 @@
+from hera.workflows import Workflow
+
+workflow = Workflow(name="single")
```
elliotgunton, line 3: We also need tests/examples for `WorkflowTemplate` and `ClusterWorkflowTemplates`

elliotgunton:
@DanCardin awesome stuff, thank you! We're happy to trial out cappa, given it's a small CLI for now and looks like we can change over to typer without too much trouble if we hit too many walls. The testable functions is a big bonus!

The PR in general looks good (just went through with some nitpicks!), and happy with the choices - having the separate CLI installable and experimental warning when running looks like enough to cover our bases in terms of angry customers! 😆

src/hera/_cli/base.py
```diff
@@ -0,0 +1,50 @@
...
+from cappa import Arg, Subcommands, command
+from typing_extensions import Annotated
+
+
+@dataclass
+class Hera:
+    subcommand: Subcommands[Generate]
+
+
+@command(help="Subcommands for converting to/from hera Workflows.")
```
elliotgunton, line 15: We have more generic intentions (see #670 examples) for this subcommand:

```suggestion
@command(help="Subcommands for generating yaml, code, and docs from Hera Workflows.")
```

poetry.lock
```diff
@@ -662,48 +686,49 @@ files = [
 
 [[package]]
 name = "mypy"
-version = "1.0.1"
+version = "1.7.1"
```
elliotgunton, line 689: I expect this is the cause of the CI errors!

tests/cli/examples/single_workflow.py
DanCardin: Given that they both inherit from `Workflow`, I wouldn't think it would matter? So long as they both have tested implementations of `to_yaml`, it seems like they'd exhibit identical behaviors.

I can certainly add tests for them, or else parametrize the tests on each of the class types. I just dont know what would be the differentiating factor of the tests, if they existed otherwise.

---

For reference, the `examples/` folder thing is solely so that I could fully exercise the importlib shenans used for loading the python files against proper modules.

poetry.lock
DanCardin: Updated PR to constrain the mypy more. Wasn't sure if you'd prefer an exact version constraint or not.

tests/cli/examples/single_workflow.py
elliotgunton: True, but it's more so that users can see it's not just for Workflows - helps avoid repeating simple questions!

.github/workflows/cicd.yaml
```diff
@@ -91,6 +91,42 @@ jobs:
         with:
           file: ./coverage.xml
 
+  test-cli:
+    name: test py${{ matrix.python-version }} (pydantic v1) on ${{ matrix.os }}
```
elliotgunton, line 95:
```suggestion
    name: test cli with py${{ matrix.python-version }} (pydantic v1) on ${{ matrix.os }}
```

.github/workflows/cicd.yaml
```diff
@@ -91,6 +91,42 @@ jobs:
...
+        uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python-version }}
+          cache: "poetry"
+
+      - name: Install dependencies
+        run: |
+          poetry install -E cli
+
+      - name: run ci checks
```
elliotgunton, line 122:
```suggestion
      - name: run cli tests
```

Makefile
```diff
@@ -44,7 +44,11 @@ lint:  ## Run a `lint` process on Hera and report problems
 
 .PHONY: test
 test:  ## Run tests for Hera
-	@poetry run python -m pytest --cov-report=term-missing -m "not on_cluster"
+	@poetry run python -m pytest --cov-report=term-missing -m "not on_cluster and not cli"
+
+.PHONY: test-cli
+test-cli:  ## Run tests for Hera
+	@poetry run python -m pytest --cov-report=term-missing -m 'cli'
```
elliotgunton, line 51: Wouldn't recommend a coverage report for a subset of tests. Might be worth a new `test-all` make target for regular tests and cli
```suggestion
	@poetry run python -m pytest -m 'cli'
```

poetry.lock
flaviuvadan: I think it's ok to pin mypy on 1.0.1 because:
1. It's a dev dependency so it does not install through Hera on client side
2. Anecdotally any time I've updated mypy there was _some_ type error, somewhere. Pinning to 1.0.1 will help us decouple the mypy update from this CLI development and not make type changes in this PR

DanCardin:
mmm. Given how you're linting now, i guess it might perhaps be easier to add `cappa` as an unconditional dev dependency, run all tests normally without this separate only-cli-tests job, and then add a test that patches away the dependency and calls `__main__.main()`

samj1912 (approved):

samj1912:
LGTM! Thanks for this! Let's follow up on subsequent PRs with additional features/changes.

PR #887, opened 2023-12-08
Add on-cluster workflow test to CI

### LINKED ISSUE #652: Integrate Argo Workflows into Hera's CI

elliotgunton:
**Pull Request Checklist**
- [x] Fixes #652
- [x] Tests added

**Description of PR**
Introduce local Argo cluster workflow tests to CI, and `make` targets to set up a local cluster for testing (this worked very well on GitHub codespaces, where I could even get a web UI to the cluster).

Added a single test to confirm the workflow runs via `create` and reaches the `succeeded` `status.phase`. We can think about creating a more flexible/extensible framework for adding `on_cluster` tests -
* I'd like to use a subset of our examples
* But we want to assert on more than the `status.phase`, e.g. output values, running order, skipped nodes etc, which would require bespoke tests for each
* It would also be good to have a wrapper for fetching these values, e.g. getting a node requires iterating through the `status.nodes` dictionary, which is keyed by hashes, so you have to check `displayName` from the values.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -126,3 +126,41 @@ regenerate-test-data: install-3.8
 	find examples -name "*.yaml" -type f -delete
 	HERA_REGENERATE=1 make test examples
 	@poetry run python -m pytest -k test_for_missing_examples --runxfail
+
+.PHONY: install-minikube
+install-minikube: ## Install minikube client
+	curl -LO https://storage.googleapis.com/minikube/releases/latest/minikube-linux-amd64
+	sudo install minikube-linux-amd64 /usr/local/bin/minikube
+
+.PHONY: install-argo
+install-argo:  ## Install argo client
+	# Download the binary
+	curl -sLO https://github.com/argoproj/argo-workflows/releases/download/v$(ARGO_WORKFLOWS_VERSION)/argo-linux-amd64.gz
+
+	# Unzip
+	gunzip argo-linux-amd64.gz
+
+	# Make binary executable
+	chmod +x argo-linux-amd64
+
+	# Move binary to path
+	sudo mv ./argo-linux-amd64 /usr/local/bin/argo
+
+	# Test installation
+	argo version
+
+.PHONY: run-argo
+run-argo: ## Start the argo server
+	minikube start --ports 2746
+	-kubectl create namespace argo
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -126,3 +126,43 @@ regenerate-test-data: install-3.8
 	find examples -name "*.yaml" -type f -delete
 	HERA_REGENERATE=1 make test examples
 	@poetry run python -m pytest -k test_for_missing_examples --runxfail
+
+.PHONY: install-minikube
```
samj1912, line 130: We should use k3d instead. It is much quicker to spin up.

Makefile
```diff
@@ -44,7 +44,7 @@ lint:  ## Run a `lint` process on Hera and report problems
 
 .PHONY: test
 test:  ## Run tests for Hera
-	@poetry run python -m pytest --cov-report=term-missing
+	@poetry run python -m pytest --cov-report=term-missing -m "not workflow"
```
flaviuvadan, line 47: picky comment - would this flag be more intuitive if it were called "remote" or "remote_in_ci"? A bunch of tests test workflows

elliotgunton: Yeah lol I didn't want to get hung up on the naming, STC! 😆 Will change

Makefile
elliotgunton: Just for posterity (and the curious) - minikube took 41s [here](https://github.com/argoproj-labs/hera/actions/runs/7144548456/job/19458392050?pr=887), and k3d took 49s [here](https://github.com/argoproj-labs/hera/actions/runs/7170762330/job/19524240213?pr=887)

elliotgunton: Benchmarks from https://www.blueshoe.io/blog/minikube-vs-k3d-vs-kind-vs-getdeck-beiboot/ suggest k3d is faster and more efficient.

flaviuvadan (approved):
This LGTM! The next iteration can address some of the other concerns you mentioned in the description. Thx again for adding this <3

PR #904, opened 2023-12-21
feat: Add --include and --exclude to the hera generate yaml command.

DanCardin:
**Pull Request Checklist**
- [ ] Fixes #
- [x] Tests added

**Description of PR**
Screenshot 2023-12-21 at 8 50 23 AM

Used to fine tune what gets picked up by the CLI, in particular when the FROM is a folder

### DIFF
--- a/src/hera/_cli/base.py
+++ b/src/hera/_cli/base.py
@@ -1,8 +1,8 @@
 from __future__ import annotations
 
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from pathlib import Path
-from typing import Union
+from typing import List, Union
 
 from cappa import Arg, Subcommands, command
 from typing_extensions import Annotated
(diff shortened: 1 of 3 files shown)


samj1912:
Could we also additionally have it support variable names?

Potentially using colon. So `**:workflow` for eg would include all files (by the `**` syntax and the latter would only filter it down to workflow variables defined in the files.

DanCardin:
hrm. I was anticipating this filter to be purely about determining the set of input files, and other options might exist to further specialize what gets selected within those files.

Personally, I'd opt for something more like `--kind workflow`, but which would obviously only work for the whole command. Do you anticipate per-file slash per-glob filtering like that to be important to the above request?

samj1912 (approved):

samj1912:
@DanCardin could you please rebase?

PR #905, opened 2023-12-21
Refactor "one or many" typed Hera values

### LINKED ISSUE #653: Standardize the way single item vs list[item] is processed on Hera object fields
One of the QOL features Hera provides is the ability to provide either single items or lists of items for fields that take exclusively lists. Hera performs the translation of single Hera, or generated model item, for the user. However, repeating the code that handles this does not add value. 

It would be amazing to centralize the processing of single Hera/model item vs lists[Hera/model item] into a single helper function with a signature similar to 
```python
def _hera_objects_to_model_object(
    objects: Union[hera.shared._base_model.BaseModel, List[hera.shared._base_model.BaseModel]],
    from_hera_type: hera.shared._base_model.BaseModel, 
    to_model_type: hera.shared._base_model.BaseModel,
) -> Optional[List[hera.shared._base_model.BaseModel]]: ...
```

1. Do nothing and maintain the current dev approach. This is ok as duplication is not _inherently_ bad but in this case it does not add much value. However, it's ok to go forward like this
2. In the `BaseModel` add a root validator that iterates over provided fields, performs a `getattr(self, f"_build_{field}")` and calls `build_{field}` if it exists. This way the root validator takes care of converting all items using the custom `_build_{field}` methods that are currently constructed on independent Hera objects

**Additional context**
None

elliotgunton:
**Pull Request Checklist**
- [x] Fixes #653 
- [x] Tests added

**Description of PR**
Currently, typing a value as a single value of a union or a list of the same union is verbose and becomes less readable. This PR adds an explicit `OneOrMany` type for us to reuse throughout the mixins, as well as validators that can normalize the value to a list (or as an alternative valid single-value).

The issue also discusses moving the `_build_*` logic into the root validator - we could do something like that instead of the `normalize_to_list` functions, but chose to get to this halfway point first.


src/hera/workflows/_mixins.py
```diff
@@ -85,11 +97,36 @@
...
     _yaml = None
 
+
+T = TypeVar("T")
+OneOrMany = Union[T, List[T]]
+"""OneOrMany is provided as a convenience to allow Hera models to accept single values or lists of
+values, and so that our code is more readable. It is used by the 'normalize' validators below."""
+
+
+def normalize_to_list(v: Optional[OneOrMany]):
```
flaviuvadan, line 107:
```suggestion
def normalize_to_list(v: Optional[OneOrMany]) -> Optional[List[T]]:
```

src/hera/workflows/_mixins.py
```diff
@@ -85,11 +97,36 @@
...
+    """Normalize given value to a list if not None."""
+    if v is None or isinstance(v, list):
+        return v
+    return [v]
+
+
+def normalize_to_list_or(*valid_types: Type):
```
flaviuvadan, line 114: I think? 

```suggestion
def normalize_to_list_or(*valid_types: Type) -> Callable[[Optional[OneOrMany]], Optional[List[T]]]:
```

src/hera/workflows/_mixins.py
```diff
@@ -85,11 +97,36 @@
 except ImportError:
     _yaml = None
 
+
+T = TypeVar("T")
+OneOrMany = Union[T, List[T]]
+"""OneOrMany is provided as a convenience to allow Hera models to accept single values or lists of
```
flaviuvadan, line 103:
```suggestion
"""OneOrMany is provided as a convenience type to allow Hera models to accept single values or lists of
```

flaviuvadan (commented):
🎖️

src/hera/workflows/_mixins.py
elliotgunton: Haha yeah lemme fix these types

flaviuvadan (approved):
🚀

PR #906, opened 2023-12-21
feat: Perform dynamic import such that local relative imports functio…

DanCardin:
…n correctly.

**Pull Request Checklist**
- [ ] Fixes #
- [x] Tests added

**Description of PR**

Given a folder structure like:

```python
folder/
  foo.py
  bar.py
```

Currently, files containing workflows cannot reference other python modules, local to the point of import.

i.e. `from .foo import ...` from inside `bar.py`.

This PR updates the way the dynamic import occurs such that local imports **do** function correctly.

What does **not** yet appear to work correctly is if the referenced code can be imported where absolute imports to a root package would function correctly. I suspect that for that to work, the FROM argument would need to reference a source `module.path` rather than a path to a file.

### DIFF
--- a/src/hera/_cli/generate/yaml.py
+++ b/src/hera/_cli/generate/yaml.py
@@ -82,10 +82,12 @@ def load_workflows_from_module(path: Path) -> list[Workflow]:
     Returns:
         A list containing all `Workflow` objects defined within that module.
     """
-    spec = importlib.util.spec_from_file_location(path.stem, path)
+    module_name = path.stem
+    spec = importlib.util.spec_from_file_location(module_name, path, submodule_search_locations=[str(path.parent)])
     assert spec
 
     module = importlib.util.module_from_spec(spec)
+    sys.modules[module_name] = module
 
     assert spec.loader
     spec.loader.exec_module(module)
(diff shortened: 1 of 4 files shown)


DanCardin:
tl;dr of the usecase, as perhaps evidenced by the example test, is to factor out common config/definitions to a module. 

---

I guess arguably this allows one to write code that doesn't "work" correctly when invoked in certain ways. like `python folder/foo.py` wouldn't work because of the relative import, whereas `python -m folder.foo` would. I'm not sure whether that's good or bad.

At least the `sys.modules` bit is commonly recommended alongside code that performs the find_spec/exec_module workflow. It's the `submodule_search_locations` bit that's...perhaps(?) more opinionated, although the docstring for `spec_from_file_location` does recommend using it to indicate the loaded file is part of a package.

Admittedly, python's semantics for directly invoking python files is generally wild anyways....

DanCardin:
After doing some more testing, I think this seems to be the right way.

* If you're executing from inside an "installed" project, then you can still import absolute paths without issue
* if you have local relative imports, it should work. as the module is imported relative to how it was referenced as though through `python -m module.path.to.module`.
  * **currently** this my strategy, to avoid needing to `poetry install` inside an argo CMP plugin
* an module with relative imports wont be directly executable through `python foo.py`, but that's just a vanilla python issue; and it's more of an author choice at that point.

samj1912 (approved):

PR #917, opened 2024-01-05
Add "buildable" protocol.

DanCardin:
**Pull Request Checklist**
- [ ] Addresss https://github.com/argoproj-labs/hera/discussions/914
- [x] Tests added

**Description of PR**
Per the above linked discussion, the goal is to enable authoring non-workflow objects that would still benefit from being declaratively described alongside hera objects...and have the hera CLI be able to generate them also. I suppose in an ideal world, they might also be `create`-able also, but that's outside my current interest scope.

---

So I'm not certain whether this is the best way to go about factoring these methods or not. It **seems** like there's perhaps some more methods (to_file, from_*, etc) and attributes (name, generate_name) that might correspond to this sort of "top-level-object that can be produced, whether it's a Workflow or not"

I had originally intended to use ModelMapperMixin, since it seems like **it** would be the right place for many of these things to live, but it's in a private module, and I wasn't sure if it made sense to subclass it outside hera itself.

Whereas as-implemented, it doesn't need to directly depend on hera objects in order to be supported. With that said, I dont know that that's an explicit goal here, I'm just unfamiliar with the design intent of some of these workings.

### DIFF
--- a/src/hera/workflows/protocol.py
+++ b/src/hera/workflows/protocol.py
@@ -4,6 +4,7 @@ from typing import Any, List, Optional, Union
 
 from typing_extensions import Protocol, runtime_checkable
 
+from hera.shared._pydantic import BaseModel
 from hera.workflows.models import (
     ClusterWorkflowTemplate,
     ContainerSetTemplate,
@@ -35,6 +36,20 @@ TWorkflow = Union[
     WorkflowTemplate,
 ]
 """`TWorkflow` is a union type collection of all the Hera workflow type objects that can manage contexts"""
+
+
+@runtime_checkable
+class Buildable(Protocol):
+    """This runtime protocol indicates that an object can built into a standalone serializable representation."""
+
+    def build(self) -> BaseModel:
+        """Produce a serializable model object."""
+
+    def to_dict(self) -> Any:
+        """Produce a raw python dict structure."""
+
+    def to_yaml(self, *args, **kwargs) -> str:
+        """Produce a yaml string of the `to_dict` representation."""
 
 
 @runtime_checkable
--- a/docs/examples/workflows-examples.md
+++ b/docs/examples/workflows-examples.md
@@ -107,6 +107,7 @@ Explore the examples through the side bar!
 | [template-defaults](https://github.com/argoproj/argo-workflows/blob/main/examples/template-defaults.yaml) |
 | [testvolume](https://github.com/argoproj/argo-workflows/blob/main/examples/testvolume.yaml) |
 | [timeouts-step](https://github.com/argoproj/argo-workflows/blob/main/examples/timeouts-step.yaml) |
+| [title-and-descriptin-with-markdown](https://github.com/argoproj/argo-workflows/blob/main/examples/title-and-descriptin-with-markdown.yaml) |
 | [work-avoidance](https://github.com/argoproj/argo-workflows/blob/main/examples/work-avoidance.yaml) |
 | [workflow-count-resourcequota](https://github.com/argoproj/argo-workflows/blob/main/examples/workflow-count-resourcequota.yaml) |
 | [workflow-event-binding/event-consumer-workfloweventbinding](https://github.com/argoproj/argo-workflows/blob/main/examples/workflow-event-binding/event-consumer-workfloweventbinding.yaml) |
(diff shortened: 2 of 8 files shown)


src/hera/workflows/protocol.py
```diff
@@ -37,6 +38,20 @@
...
+class Buildable(Protocol):
+    """This runtime protocol indicates that an object can built into a standalone serializable representation."""
+
+    def build(self) -> BaseModel:
+        """Produce a serializable model object."""
+
+    def to_dict(self) -> Any:
+        """Produce a raw python dict structure."""
+
+    def to_yaml(self, *args, **kwargs) -> str:
```
samj1912, line 51: I don't think we need to_yaml to be part of the protocol

src/hera/workflows/protocol.py
```diff
@@ -37,6 +38,20 @@
...
 
 
+@runtime_checkable
+class Buildable(Protocol):
+    """This runtime protocol indicates that an object can built into a standalone serializable representation."""
+
+    def build(self) -> BaseModel:
+        """Produce a serializable model object."""
+
+    def to_dict(self) -> Any:
```
samj1912, line 48: I think we can also skip this if build is guaranteed to return a basemodel

src/hera/workflows/protocol.py
DanCardin: So my usecase is wholly based off being able to have the CLI be able to infer the ability to `to_yaml` something. whether it's a Workflow, some other theoretical hera-native top-level object that's not a worfklow subclass, or some externally defined resource.

I think `to_yaml` is the **minimum** protocol function I require. I included `to_dict` and `build` primarily because it seems likely that any top-level object will have this `build` step, and the current Workflow function chain path is `to_yaml -> to_dict -> build`.

DanCardin:
...it maybe seems to me like Buildable and BuildableMixin as concepts, are more narrowly scoped to my usecase, and something like "ResourceBase" would be a more widely applicable concept? Namely because there are a bunch of other functions/fields seem to be very generic and widely applicable to anything that can be serialized standalone as a top-level output.

perhaps something more like:

```python
class ResourceBase(BaseModel):  # dunno whether this is really a mixin, so much as the expected base class...
    kind: ClassVar[str] # ???

    name: str | None = None
    generate_name: str | None = None # unsure about this one...but used by to_file currently, so...
    api_version: str | None = None
    annotations: ...

    def build(self) -> BaseModel: ...

    def from_model(cls, ...): ...

    def from_dict(cls, ...): ...
    def to_dict(self, ...): ...

    def from_yaml(cls, ...): ...
    def to_yaml(self, ...): ...

    def from_file(cls, ...): ...
    def to_file(self, ...): ...
```
...Which starts to look a lot like `ModelMapperMixin`...which is what i started with. I just found myself changing a lot more code with this direction, and I wasn't entirely sure how the `ModelMapper` and `_WorkflowModelMapper` and whatnot all fit together enough to warrant fiddling with everything.

samj1912:
@DanCardin trying to circle back here, would it instead make sense to support the following - 

class must contain api_version and kind properties.

if `to_yaml` is implemented proceed
if `to_dict` is implemented, call and then convert to yaml and proceed
if `build` is implemented, call and then convert to dict -> yaml and proceed?

So minimally we need two properties and any of the above 3 methods to exist.

DanCardin:
@samj1912, the thing is that `to_dict` and `to_yaml` dont really need implementing, they're inferable from the the fact that `build` produces a model. which I guess is what you were saying originally.

I think perhaps the protocol here is kind of unnecessary. I had originally included it to serve as a contract that external implementers could use that does not involve subclassing hera base classes themselves, but i think perhaps that goal isn't particularly useful in and of itself.

---

Coincidentally, I started looking at this again this morning. I think perhaps `ModelMapperMixin` could:

* (relocated and) renamed to something like `ResourceBase`
* relocate `to_dict` and `to_yaml` onto it
* define a default (raise NotImplementedError) `build` on it
* perhaps something custom with api_version and kind, which seem like they'd be intrinsically tied to the class, rather than being normal user-input attributes

At which point one could externally author something like:

```
class ServiceAccount(ResourceBase):
    mapped_model: ClassVar[BaseModel] = ServiceAccountModel
    api_version: ClassVar[str] = "v1"
    kind: ClassVar[str] = "ServiceAccount"

    annotations: Annotated[Optional[Dict[str, str]], ModelMapper("metadata.annotations")] = None
```

Which would enable the CLI to check for `isinstance(x, ResourceBase)`, which would then allow anyone to author a CRD, of which Workflow, CronWorkflow, etc would all just be examples of.

PR #920, opened 2024-01-10
Add Pydantic V1 IO models for use with Hera Runner

### LINKED ISSUE #858: Improve Hera I/O models

Currently hera i/o with annotated params can become extremely verbose. The output syntax is especially error-prone.

Introduce a HeraInput and HeraOutput class which derives from pydantic basemodel. When this is present, hera script runner will expand out the top level class attributes to params.

```python
from hera.workflows.io import Input, Output


class AnotherModel(BaseModel):
	...

class MyInput(InputModel):
	my_param: int = 1
	another_param: Annotated[int, Parameter(name="another-param") = 2
	another_artifact: Annotated[AnotherModel, Artifact(...)]


class MyOutput(OutputModel):
	my_output: str = ""
	another_output: Annotated[Path, Parameter("test"]]

def my_function(my_input: MyInput, another_param_inline: int) -> MyOutput:
	...
	outputs = MyOutput(exit_code=10)
	outputs.my_output = "foo"
	outputs.another_output.write("foo")
	outputs.result = "foo"
	return outputs
```

**Additional context**

elliotgunton:
**Pull Request Checklist**
- [x] Part of #858 
- [x] Tests added
- [ ] ~Documentation/examples added~ See #939 

**Description of PR**
Currently hera i/o with annotated params can become extremely verbose. The output syntax is especially error-prone.

This PR introduces custom Input/Output BaseModels for users to subclass, which allow a cleaner arrangement of inputs and outputs for functions. These are available under the `script_pydantic_io` experimental feature flag.

With these Pydantic input/output models, the following should be noted:
* duplicated param names (for normal Parameters as well as the new models) are now detected in Hera rather than when linted by Argo (as well as duplicated artifact names). Parameters and Artifacts having the same name is legal in the Argo spec as they exist in different scopes e.g.
```yaml
...
      inputs:
        parameters:
          - name: my-name
            default: test
        artifacts:
          - name: my-name
            path: /tmp
            optional: true
...
```
* `exit_code` and `result` are reserved attributes for the `RunnerOutput`. A user trying to use their own parameters with these names would have to be specified with an annotated parameter e.g. `my_exit_code: Annotated[int, Parameter(name="exit_code")]` (TBC with a test)
* Scripts cannot have a return tuple containing any `RunnerOutput` to avoid multiple `exit_code`s being specified. @samj1912 / @flaviuvadan this is up for debate but I think would encourage better practices to discourage tuples and have a single script template outputting a single `RunnerOutput` subclass, and it keeps the logic clearer from the Hera side. Users can still use inline output parameters alongside the `RunnerOutput` return annotation
* Multiple input parameters when using a `RunnerInput` in the function params **is not legal**
* A `RunnerInput`'s `__fields__` as defined by pydantic are used to "explode" the input class into constituent parameters for the Argo spec. i.e. using the following class as an input param to a script function:
```py
class MyInput(RunnerInput):
     my_input_str: str
     my_input_int: int

@script(constructor="runner")
def my_func(my_input: MyInput):
    ...
```
will create the script template `my_func` in yaml with Parameters `my_input_str` and `my_input_int`, NOT `my_input`, [see the example](https://github.com/argoproj-labs/hera/blob/92f11d341eb29d2501b9ee5be57a703160b35e24/docs/examples/workflows/experimental/script_pydantic_io.md)

### DIFF
--- a/src/hera/shared/serialization.py
+++ b/src/hera/shared/serialization.py
@@ -30,10 +30,18 @@ class PydanticEncoder(JSONEncoder):
         # Note that these are slightly different outputs b/w v1 and v2
         # v1 will give the actual python object whereas v2 will serialize it into
         # a json compatible format.
+        from hera.workflows.io import RunnerOutput
+
         if _PYDANTIC_VERSION == 1:
+            if isinstance(o, RunnerOutput):
+                obj = o.dict(by_alias=True)
+                return o.replace_keys(obj)
             if isinstance(o, BaseModel):
                 return o.dict(by_alias=True)
         else:
+            if isinstance(o, RunnerOutput):
+                obj = o.dict(by_alias=True)
+                return o.replace_keys(obj)
             if isinstance(o, BaseModel):
                 return o.model_dump(by_alias=True, mode="json")  # type: ignore
             if isinstance(o, V1BaseModel):
--- a/src/hera/shared/serialization.py
+++ b/src/hera/shared/serialization.py
@@ -30,18 +30,10 @@ class PydanticEncoder(JSONEncoder):
         # Note that these are slightly different outputs b/w v1 and v2
         # v1 will give the actual python object whereas v2 will serialize it into
         # a json compatible format.
-        from hera.workflows.io import RunnerOutput
-
         if _PYDANTIC_VERSION == 1:
-            if isinstance(o, RunnerOutput):
-                obj = o.dict(by_alias=True)
-                return o.replace_keys(obj)
             if isinstance(o, BaseModel):
                 return o.dict(by_alias=True)
         else:
-            if isinstance(o, RunnerOutput):
-                obj = o.dict(by_alias=True)
-                return o.replace_keys(obj)
             if isinstance(o, BaseModel):
                 return o.model_dump(by_alias=True, mode="json")  # type: ignore
             if isinstance(o, V1BaseModel):
--- a/src/hera/workflows/io.py
+++ b/src/hera/workflows/io.py
@@ -3,6 +3,7 @@ from collections import ChainMap
 from typing import Any, List, Union
 
 from hera.shared._pydantic import BaseModel
+from hera.shared.serialization import serialize
 from hera.workflows.artifact import Artifact
 from hera.workflows.parameter import Parameter
 
@@ -30,7 +31,8 @@ class RunnerInput(BaseModel):
                 if isinstance(get_args(annotations[field])[1], Parameter):
                     param = get_args(annotations[field])[1]
                     if cls.__fields__[field].default:
-                        param.default = cls.__fields__[field].default
+                        # Serialize the value (usually done in Parameter's validator)
+                        param.default = serialize(cls.__fields__[field].default)
                     parameters.append(param)
             else:
                 # Create a Parameter from basic type annotations
--- a/src/hera/workflows/io.py
+++ b/src/hera/workflows/io.py
@@ -1,4 +1,4 @@
-"""Input/output models for the Hera runner. TODO move to hera.workflows.runner package."""
+"""Input/output models for the Hera runner."""
 from collections import ChainMap
 from typing import Any, List, Union
 
--- a/src/hera/workflows/io.py
+++ b/src/hera/workflows/io.py
@@ -1,5 +1,5 @@
 """Input/output models for the Hera runner. TODO move to hera.workflows.runner package."""
-from typing import Any, List, Union
+from typing import Any, Dict, List, Tuple, Type, Union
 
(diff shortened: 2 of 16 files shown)


samj1912:
@elliotgunton can you provide a brief examples/details in the PR description of the functionality that is currently enabled?

Specifically I am curious how we handle things when a user - 
- has duplicate param names between a non annotated param/annotated param/normal param
- has duplicate names between params/artifacts (this should be valid?)
- has a parameter/artifact called output/exit code
- has multiple output return types
- has multiple input types
- how is the input object constructed from argo params

Having that in the PR description would help future readers understand what is happening.

src/hera/shared/serialization.py
```diff
@@ -30,10 +30,18 @@ def default(self, o: Any):
         # Note that these are slightly different outputs b/w v1 and v2
         # v1 will give the actual python object whereas v2 will serialize it into
         # a json compatible format.
+        from hera.workflows.io import RunnerOutput
+
         if _PYDANTIC_VERSION == 1:
+            if isinstance(o, RunnerOutput):
+                obj = o.dict(by_alias=True)
+                return o.replace_keys(obj)
```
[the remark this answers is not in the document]
elliotgunton: I don't think this is needed in the end, was solved by using the attribute's annotation directly in `_save_annotated_return_outputs`

src/hera/workflows/io.py
```diff
@@ -0,0 +1,102 @@
...
+    def _get_parameters(cls) -> List[Parameter]:
+        parameters = []
+        annotations = {k: v for k, v in ChainMap(*(get_annotations(c) for c in cls.__mro__)).items()}
+
+        for field in cls.__fields__:
+            if get_origin(annotations[field]) is Annotated:
+                if isinstance(get_args(annotations[field])[1], Parameter):
+                    param = get_args(annotations[field])[1]
+                    if cls.__fields__[field].default:
+                        param.default = cls.__fields__[field].default
```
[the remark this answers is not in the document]
elliotgunton: Good point, thought it would be serialized from the `root_validator` but that wouldn't happen if we're doing a late init. The problem doesn't exist for normal func inputs because we create a new Parameter with the default here https://github.com/argoproj-labs/hera/blob/7982f90c604d1c1c7c87ce120d6f7fbe10f36398/src/hera/workflows/script.py#L337

src/hera/workflows/io.py
```diff
@@ -0,0 +1,86 @@
+"""Input/output models for the Hera runner. TODO move to hera.workflows.runner package."""
```
elliotgunton, line 1: @samj1912 / @flaviuvadan any preference on where these classes should live? Considering they're only usable in the runner, thought a rearrangement of that module would work and move these classes into it?

elliotgunton:
>I see we added support for defaults as well (is it also added back to the normal params apart from hera io models?)

Just answering this one separately as it's existing behaviour for normal func params - for script annotations see https://github.com/argoproj-labs/hera/blob/7982f90c604d1c1c7c87ce120d6f7fbe10f36398/src/hera/workflows/script.py#L416-L427

or for the non-experimental code path, see https://github.com/argoproj-labs/hera/blob/7982f90c604d1c1c7c87ce120d6f7fbe10f36398/src/hera/workflows/script.py#L331-L338

so we already serialize the python default for functions

src/hera/workflows/io.py
samj1912: workflows/io is fine

src/hera/workflows/script.py
```diff
@@ -358,11 +359,30 @@ def append_annotation(annotation: Union[Artifact, Parameter]):
...
-        append_annotation(get_args(inspect.signature(source).return_annotation)[1])
-    elif get_origin(inspect.signature(source).return_annotation) is tuple:
-        for annotation in get_args(inspect.signature(source).return_annotation):
+    return_annotation = inspect.signature(source).return_annotation
+    if get_origin(return_annotation) is Annotated:
+        append_annotation(get_args(return_annotation)[1])
+    elif get_origin(return_annotation) is tuple:
+        for annotation in get_args(return_annotation):
+            if isinstance(annotation, type) and issubclass(annotation, RunnerOutput):
+                raise ValueError("RunnerOutput cannot be part of a tuple output")
```
elliotgunton, line 368: Only one `RunnerOutput` is allowed as the function return type. Note that we also haven't dealt with nested `Annotated` types above.

elliotgunton:
@samj1912 ready for re-review! Self-reviewed to double check this PR stands on its own to implement for Pydantic V1 and signposted the single input/output we discussed 🚀

samj1912 (approved):

PR #935, opened 2024-01-18
fix: Handle multiline strings in yaml serialization.

DanCardin:
**Pull Request Checklist**
- [ ] Fixes #
- [x] Tests added

**Description of PR**
Currently, multiline strings serialize to yaml in a way that is not true to the original content.

For example, copying [argo workflows' exit-handler example](https://github.com/argoproj/argo-workflows/blob/main/examples/exit-handler-slack.yaml)

excerpt:
```yaml
      args: [
        "curl -X POST --data-urlencode 'payload={
          \"text\": \"{{workflow.name}} finished\",
          \"blocks\": [
            {
              \"type\": \"section\",
              \"text\": {
                \"type\": \"mrkdwn\",
                \"text\": \"Workflow {{workflow.name}} {{workflow.status}}\",
              }
            }
          ]
        }'
        YOUR_WEBHOOK_URL_HERE"
      ]
```

currently produces:
```yaml
        args:
        - "curl -X POST --data-urlencode 'payload={\n          \"text\": \"{{workflow.name}}\
          \ finished\",\n          \"blocks\": [\n            {\n              \"\
          type\": \"section\",\n              \"text\": {\n                \"type\"\
          : \"mrkdwn\",\n                \"text\": \"Workflow {{workflow.name}} {{workflow.status}}\"\
          ,\n              }\n            }\n          ]\n        }'\n        YOUR_WEBHOOK_URL_HERE\"\
          \n        "
```

whereas this PR produces:
```yaml
        args:
        - |-
          curl -X POST --data-urlencode 'payload={
                    "text": "{{workflow.name}} finished",
                    "blocks": [
                      {
                        "type": "section",
                        "text": {
                          "type": "mrkdwn",
                          "text": "Workflow {{workflow.name}} {{workflow.status}}",
                        }
                      }
                    ]
                  }'
                  YOUR_WEBHOOK_URL_HERE";
```

---

Note that this does not fix https://github.com/yaml/pyyaml/issues/121#issuecomment-1018117110. That is, the following example will revert back to original formatting due to the trailing `\n` at the end of the string.
```python
args="""
two
"""
```

I would suggest switching `ruamel.yaml` over pyyaml just generally, which would fix this issue, but that's probably going to be considered a breaking change.

In any case is probably more tangential to the root issue above (which affects both libraries in the same way).

### DIFF
--- a/src/hera/_yaml.py
+++ b/src/hera/_yaml.py
@@ -1,0 +1,31 @@
+from types import ModuleType
+from typing import Optional
+
+_yaml: Optional[ModuleType] = None
+try:
+    import yaml
+
+    _yaml = yaml
+except ImportError:
+    _yaml = None
+else:
+
+    def str_presenter(dumper, data):
+        """Configures yaml for dumping multiline strings.
+
+        Ref: https://github.com/yaml/pyyaml/issues/240
+        """
+        if data.count("\n") > 0:  # check for multiline string
+            return dumper.represent_scalar("tag:yaml.org,2002:str", data, style="|")
+        return dumper.represent_scalar("tag:yaml.org,2002:str", data)
+
+    _yaml.add_representer(str, str_presenter)
+    _yaml.representer.SafeRepresenter.add_representer(str, str_presenter)
+
+
+def dump(*args, default_flow_style: bool = False, sort_keys: bool = False, **kwargs) -> str:
+    """Builds the Workflow as an Argo schema Workflow object and returns it as yaml string."""
+    if not _yaml:
+        raise ImportError("`PyYAML` is not installed. Install `hera[yaml]` to bring in the extra dependency")
+
+    return _yaml.dump(*args, default_flow_style=default_flow_style, sort_keys=sort_keys, **kwargs)
--- a/src/hera/_yaml.py
+++ b/src/hera/_yaml.py
@@ -23,9 +23,12 @@ else:
     _yaml.representer.SafeRepresenter.add_representer(str, str_presenter)
 
 
-def dump(*args, default_flow_style: bool = False, sort_keys: bool = False, **kwargs) -> str:
+def dump(*args, **kwargs) -> str:
     """Builds the Workflow as an Argo schema Workflow object and returns it as yaml string."""
     if not _yaml:
         raise ImportError("`PyYAML` is not installed. Install `hera[yaml]` to bring in the extra dependency")
 
-    return _yaml.dump(*args, default_flow_style=default_flow_style, sort_keys=sort_keys, **kwargs)
+    # Set some default options if not provided by the user
+    kwargs.setdefault("default_flow_style", False)
+    kwargs.setdefault("sort_keys", False)
+    return _yaml.dump(*args, **kwargs)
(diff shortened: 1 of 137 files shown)


DanCardin:
😱  neat! An even better way of exemplifying the benefit than my test, lol.

Just to check, are the literal examples that are being rewritten tested somehow to make sure this hasn't secretly broken the output somehow, even if it looks prettier?

flaviuvadan:
> 😱 neat! An even better way of exemplifying the benefit than my test, lol. Added as a 2nd commit, separate from the actual code change.
> 
> Just to check, are the literal examples that are being rewritten tested somehow to make sure this hasn't secretly broken the output somehow, even if it looks prettier?

Yea, they're all tested [here](https://github.com/argoproj-labs/hera/blob/main/tests/test_examples.py)! It essentially compares the YAML output but without formatting concerns. As long as the content is valid, the tests pass

flaviuvadan (approved):
All those examples look _sooo_ much better now 🎖️

flaviuvadan:
We can 🚢 once @elliotgunton or @samj1912 approve too. Their stake in the CLI + YAML is larger than mine, so I want to be mindful of their input

src/hera/_yaml.py
```diff
@@ -0,0 +1,31 @@
...
+    _yaml.add_representer(str, str_presenter)
+    _yaml.representer.SafeRepresenter.add_representer(str, str_presenter)
+
+
+def dump(*args, default_flow_style: bool = False, sort_keys: bool = False, **kwargs) -> str:
+    """Builds the Workflow as an Argo schema Workflow object and returns it as yaml string."""
+    if not _yaml:
+        raise ImportError("`PyYAML` is not installed. Install `hera[yaml]` to bring in the extra dependency")
+
+    return _yaml.dump(*args, default_flow_style=default_flow_style, sort_keys=sort_keys, **kwargs)
```
elliotgunton, line 31: I'm not sure it's worth having `default_flow_style: bool = False, sort_keys: bool = False` in the function signature as we're passing args/kwargs through to the actual function. We can do something with decorators/`ParamSpec` to help intellisense to get the args/kwargs of the real `dump` function but that can be done later (and might have an issue if you have a named arg that would be expected in the kwargs, so would prefer to remove them for now).

```suggestion
def dump(*args, **kwargs) -> str:
    """Builds the Workflow as an Argo schema Workflow object and returns it as yaml string."""
    if not _yaml:
        raise ImportError("`PyYAML` is not installed. Install `hera[yaml]` to bring in the extra dependency")

    # Set some default options if not provided by the user
    kwargs.setdefault("default_flow_style", False)
    kwargs.setdefault("sort_keys", False)
    return _yaml.dump(*args, **kwargs)
```

DanCardin: I dont think the native yaml dump function in either of the yaml libraries have real kwargs typings, just **kwargs, fwiw.

I'm pretty sure there's no functional difference between the two versions in any way (except that you get slightly more signature information). As for it being a named arg or not, if the underlying library changed params, given that the **kwargs are getting passed through from your public API, it'd represent a user visible behavior change. In that way, you're probably better off having explicit options here rather than even passing through the library's own kwargs.

With that said, it's an internal module now, and these named kwargs are not getting passed through from the root `to_yaml` call, so it's mostly irrelevant. I dont feel super strongly about putting back to how it was.

elliotgunton (approved):
Awesome stuff! 🚀  The dodgy source dumping has been bugging me forever! 😆

PR #937, opened 2024-01-18
Support passing Hera Parameter/ModelParameter as dictionary input to tasks

### LINKED ISSUE #923: Remove the need to write `.value` on `Parameter`s passed to `arguments`
If users don't write `.value` on `arguments={'a': other.get_parameter('a').value}` then Hera passes a Pydantic dictionary representation of the `a` parameter object. So, the problem is that the code works syntactically but it breaks semantically

Hera automatically performs `.value` internally if any object passed to the `arguments` mapping is a `Parameter`

CC: @ljyanesm

flaviuvadan:
**Pull Request Checklist**
- [x] Fixes #923 
- [ ] Tests added

See #923

### DIFF
--- a/src/hera/workflows/_mixins.py
+++ b/src/hera/workflows/_mixins.py
@@ -622,9 +622,10 @@ class ArgumentsMixin(BaseMixin):
             if isinstance(arg, dict):
                 for k, v in arg.items():
                     if isinstance(v, Parameter):
-                        value = v.as_argument()
+                        value = v.with_name(k).as_argument()
                     elif isinstance(v, ModelParameter):
                         value = Parameter.from_model(v).as_argument()
+                        value.name = k
                     else:
                         value = Parameter(name=k, value=v).as_argument()
 
--- a/src/hera/workflows/_mixins.py
+++ b/src/hera/workflows/_mixins.py
@@ -626,12 +626,12 @@ class ArgumentsMixin(BaseMixin):
                     elif isinstance(v, ModelParameter):
                         value = Parameter.from_model(v).as_argument()
                     else:
-                        value = v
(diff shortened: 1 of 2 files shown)


src/hera/workflows/_mixins.py
```diff
@@ -621,12 +621,17 @@ def _build_arguments(self) -> Optional[ModelArguments]:
...
-                        [value.as_argument()]
-                        if result.parameters is None
-                        else result.parameters + [value.as_argument()]
-                    )
+                    if isinstance(v, Parameter):
+                        value = v.as_argument()
+                    elif isinstance(v, ModelParameter):
+                        value = Parameter.from_model(v).as_argument()
+                    else:
+                        value = v
```
elliotgunton, line 629: We would still need to create a parameter from the given dict value `v` as `result.parameter` is a list of argument parameters (this is what broke CI lol).

```suggestion
                        value = Parameter(name=k, value=v).as_argument()
```

We could call `as_argument` after setting `value` in each of the three `if` branches, not sure if it's more readable but it's the common action after acquiring a Hera `Parameter`.

src/hera/workflows/_mixins.py
```diff
@@ -621,12 +621,17 @@ def _build_arguments(self) -> Optional[ModelArguments]:
...
+
+                    if result.parameters is None:
+                        result.parameters = [value]
+                    else:
+                        result.parameters.extend([value])
```
elliotgunton, line 634: Can just append!
```suggestion
                        result.parameters.append(value)
```

elliotgunton (changes requested):

elliotgunton:
May I suggest a test or two? 👀

elliotgunton (approved):

PR #938, opened 2024-01-19
Add RunnerInput/Output Pydantic V2 classes

elliotgunton:
Part of #858, follow up to #920  
* Enables users to use Pydantic V2 objects in their scripts while maintaining V1 usage internally for Hera
* RunnerInput/Output classes are created in hera.workflows.io depending on the value of _PYDANTIC_VERSION - users will automatically get classes using their (possibly pinned) version of Pydantic
* I have not yet managed to get an explicit test that uses the automatic import of V1 classes - we may just have to rely on the Pydantic V1 CI check.

### DIFF
--- a/src/hera/workflows/io/v1.py
+++ b/src/hera/workflows/io/v1.py
@@ -1,4 +1,4 @@
-"""Input/output models for the Hera runner."""
+"""Pydantic V1 input/output models for the Hera runner."""
 from collections import ChainMap
 from typing import Any, List, Optional, Union
 
@@ -81,7 +81,7 @@ class RunnerOutput(BaseModel):
     """
 
     exit_code: int = 0
-    result: Any
+    result: Any = None
 
     @classmethod
     def _get_outputs(cls) -> List[Union[Artifact, Parameter]]:
--- a/src/hera/workflows/io/v2.py
+++ b/src/hera/workflows/io/v2.py
@@ -1,0 +1,110 @@
+"""Pydantic V2 input/output models for the Hera runner.
+
+RunnerInput/Output are only defined in this file if Pydantic v2 is installed.
+"""
+from collections import ChainMap
+from typing import Any, List, Optional, Union
+
+from hera.shared.serialization import serialize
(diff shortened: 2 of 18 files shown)


elliotgunton:
Changed back to draft while I rethink the approach - mentioned in Slack, we want to:

> define the classes once as mixins, that expect users to subclass BaseModel alongside our RunnerInput which means we can introspect in the _get_parameters for what BaseModel the user has "mixed in" with our class, rather than create V1/V2 RunnerInput/Output classes. This will help get around the problem of using V1 BaseModel while pydantic v2 is installed (almost impossible with current implementation)

src/hera/workflows/io/v2.py
```diff
@@ -0,0 +1,105 @@
...
+    from hera.workflows._inspect import get_annotations  # type: ignore
+
+try:
+    from typing import Annotated, get_args, get_origin  # type: ignore
+except ImportError:
+    from typing_extensions import Annotated, get_args, get_origin  # type: ignore
+
+from importlib.util import find_spec
+
+if find_spec("pydantic.v1"):
```
elliotgunton, line 24: @samj1912 does this make sense to you? Any other ways we can/should do this? Also on file naming - I think  it makes it clear which version clients should use, and it makes sense that the `v2.py` file only defines `RunnerInput`/`RunnerOutput` if pydantic v2 is installed.

elliotgunton:
Ping @samj1912 - I'd like to add all RunnerInput/Output code and docs before next release (also ping @flaviuvadan)

src/hera/workflows/io/v1.py
```diff
@@ -1,4 +1,4 @@
-"""Input/output models for the Hera runner."""
```
samj1912, line 1: We need an init file for io.

There can we do the following - 

```
if find_spec("pydantic.v1"):
	from hera.workflows.io.v2 import RunnerInput, RunnerOutput
else:
	from hera.workflows.io.v1 import RunnerInput, RunnerOutput
```

samj1912 (approved):

PR #942, opened 2024-01-25
refactor: Define a root class from which workflow resources inherit.

DanCardin:
**Pull Request Checklist**
- [x] Addresses the problem given by https://github.com/argoproj-labs/hera/discussions/914 / https://github.com/argoproj-labs/hera/pull/917 in a different way
- [ ] Tests added

**Description of PR**
First i'll summarize the above prior PR/discussion. (feel free to correct any bad k8s/workflows terminology below 😬)

Basically the "issue" is that certain CRDs ([ExternalSecret](https://external-secrets.io/v0.4.4/api-externalsecret/) and `ServiceAccount` are the ones in particular that I've needed) need to be referenced by name during the construction of a hera workflow. And in the case of both of these objects (in my case), they're only being created for the express purpose of the workflows they're being used with.

(perhaps the argument could be made that something like `ServiceAccount` **should** have a native hera representation. But it feels like ExternalSecret is probably out of scope, and/or that generally there would be kinds of CRDs that will inevitably be too niche to be in scope).

Thus, it would be really convenient to be able to author the (very small) amount of k8s manifest required to create those objects from within python/hera itself, and have it be created (again in my case through `hera generate yaml`) automatically alongside the workflows themselves.

As I understand it, all the argo `Workflow.*` objects are themselves CRDs, so my PR takes the tack that there could be a base class from which all the existing hera top-level object types (CRDs) inherit. So far I've named it `ResourceBase`. This base class provides a bunch of the common methods that would be applicable to all CRDs, and generally the shape that subclasses should abide by. For the most part, this class looks a lot like the existing `ModelMapperMixin`, with some additional stolen methods from `Workflow`.

Doing this basically just enables external consumers to also subclass `ResourceBase` and gain the benefits. Namely that the CLI will treat those objects as though they were the same as `Workflow` (or siblings) in `generate yaml` or other future commands.

---
sidebar, I think there is perhaps some missing test coverage (which i'll look into, if you like the PR), as far as this PR goes. I had initially failed to define a `mapped_model` attribute on `WorkflowTemplate`, which I'd have expected to make it produce idk...and incorrect `kind: Workflow` or something.

### DIFF
--- a/src/hera/workflows/cron_workflow.py
+++ b/src/hera/workflows/cron_workflow.py
@@ -4,20 +4,15 @@ See https://argoproj.github.io/argo-workflows/cron-workflows
 for more on CronWorkflows.
 """
 from pathlib import Path
-from typing import Dict, Optional, Type, Union, cast
+from typing import ClassVar, Dict, Optional, Type, Union, cast
 
 try:
-    from typing import Annotated, get_args, get_origin  # type: ignore
+    from typing import Annotated  # type: ignore
 except ImportError:
-    from typing_extensions import Annotated, get_args, get_origin  # type: ignore
+    from typing_extensions import Annotated  # type: ignore
 
 from hera.exceptions import NotFound
-from hera.shared._pydantic import BaseModel
-from hera.workflows._mixins import (
-    ModelMapperMixin,
-    _get_model_attr,
-    _set_model_attr,
-)
+from hera.shared._pydantic import BaseModel, PydanticBaseModel
 from hera.workflows.models import (
     CreateCronWorkflowRequest,
     CronWorkflow as _ModelCronWorkflow,
@@ -28,42 +23,11 @@ from hera.workflows.models import (
     Workflow as _ModelWorkflow,
 )
 from hera.workflows.protocol import TWorkflow
-from hera.workflows.workflow import Workflow, _WorkflowModelMapper
+from hera.workflows.resource_base import ModelMapper, Self, _get_model_attr
+from hera.workflows.workflow import Workflow
 
 
-class _CronWorkflowModelMapper(_WorkflowModelMapper):
-    @classmethod
-    def _get_model_class(cls) -> Type[BaseModel]:
-        return _ModelCronWorkflow
-
-    @classmethod
-    def build_model(
-        cls, hera_class: Type[ModelMapperMixin], hera_obj: ModelMapperMixin, model: TWorkflow
-    ) -> TWorkflow:
-        assert isinstance(hera_obj, ModelMapperMixin)
-
-        for attr, annotation in hera_class._get_all_annotations().items():
-            if get_origin(annotation) is Annotated and isinstance(
-                get_args(annotation)[1], ModelMapperMixin.ModelMapper
-            ):
-                mapper = get_args(annotation)[1]
-                if not isinstance(mapper, _CronWorkflowModelMapper) and mapper.model_path[0] == "spec":
-                    # Skip attributes mapped to spec by parent _WorkflowModelMapper
-                    continue
-
-                # Value comes from builder function if it exists on hera_obj, otherwise directly from the attr
-                value = (
-                    getattr(hera_obj, mapper.builder.__name__)()
-                    if mapper.builder is not None
-                    else getattr(hera_obj, attr)
-                )
-                if value is not None:
-                    _set_model_attr(model, mapper.model_path, value)
-
-        return model
-
-
-class CronWorkflow(Workflow):
+class CronWorkflow(Workflow, traverse_mro=False):
     """CronWorkflow allows a user to run a Workflow on a recurring basis.
 
     Note:
@@ -72,20 +36,19 @@ class CronWorkflow(Workflow):
         spec. See [CronWorkflowSpec](https://argoproj.github.io/argo-workflows/fields/#cronworkflow) for more details.
     """
 
-    concurrency_policy: Annotated[Optional[str], _CronWorkflowModelMapper("spec.concurrency_policy")] = None
-    failed_jobs_history_limit: Annotated[
-        Optional[int], _CronWorkflowModelMapper("spec.failed_jobs_history_limit")
-    ] = None
-    schedule: Annotated[str, _CronWorkflowModelMapper("spec.schedule")]
-    starting_deadline_seconds: Annotated[
-        Optional[int], _CronWorkflowModelMapper("spec.starting_deadline_seconds")
-    ] = None
-    successful_jobs_history_limit: Annotated[
-        Optional[int], _CronWorkflowModelMapper("spec.successful_jobs_history_limit")
-    ] = None
-    cron_suspend: Annotated[Optional[bool], _CronWorkflowModelMapper("spec.suspend")] = None
-    timezone: Annotated[Optional[str], _CronWorkflowModelMapper("spec.timezone")] = None
-    cron_status: Annotated[Optional[CronWorkflowStatus], _CronWorkflowModelMapper("status")] = None
+    mapped_model: ClassVar[Type[BaseModel]] = _ModelCronWorkflow
+
+    api_version: Annotated[Optional[str], ModelMapper("api_version")] = None
+    kind: Annotated[Optional[str], ModelMapper("kind")] = None
+
+    concurrency_policy: Annotated[Optional[str], ModelMapper("spec.concurrency_policy")] = None
+    failed_jobs_history_limit: Annotated[Optional[int], ModelMapper("spec.failed_jobs_history_limit")] = None
+    schedule: Annotated[str, ModelMapper("spec.schedule")]
+    starting_deadline_seconds: Annotated[Optional[int], ModelMapper("spec.starting_deadline_seconds")] = None
+    successful_jobs_history_limit: Annotated[Optional[int], ModelMapper("spec.successful_jobs_history_limit")] = None
+    cron_suspend: Annotated[Optional[bool], ModelMapper("spec.suspend")] = None
+    timezone: Annotated[Optional[str], ModelMapper("spec.timezone")] = None
+    cron_status: Annotated[Optional[CronWorkflowStatus], ModelMapper("status")] = None
 
     def create(self) -> TWorkflow:  # type: ignore
         """Creates the CronWorkflow on the Argo cluster."""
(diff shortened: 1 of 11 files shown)


src/hera/workflows/workflow.py
```diff
@@ -153,99 +146,110 @@ def _build_templates(self) -> Optional[List[TTemplate]]:
                             self.volume_claim_templates.append(claim)
         return templates or None
 
+    mapped_model: ClassVar[Type[BaseModel]] = _ModelWorkflow
```
DanCardin, line 149: As I pointed out in the PR desc. `ModelMapper` doesn't require subclassing, because i've added this field to each subclass, which tells the resource class itself to which model it refers.

Then I use `__init_subclass__` to perform the checks `ModelMapper` had been checking, at subclass creation time. This change could probably have happened as a separate PR. It was sort-of necessary to happen **for** this PR, but it could have happened independently of it, i think.

src/hera/workflows/cron_workflow.py
```diff
@@ -28,42 +23,11 @@
...
-                if value is not None:
-                    _set_model_attr(model, mapper.model_path, value)
-
-        return model
+from hera.workflows.resource_base import ModelMapper, Self, _get_model_attr
+from hera.workflows.workflow import Workflow
 
 
-class CronWorkflow(Workflow):
+class CronWorkflow(Workflow, traverse_mro=False):
```
DanCardin, line 30: CronWorfklow is kind of weird (because it doesn't define all its fields, and retargets a bunch of those fields into a sub-object).

I added a traverse_mro flag to handle the way its implemented now. I'd probably personally just duplicate the set of fields with the different model_paths. But this is how it was implemented before, so I left it as-is.

src/hera/workflows/resource_base.py
```diff
@@ -0,0 +1,186 @@
...
+class ResourceBase(BaseMixin):
+    """Base class for top-level Hera CRDs.
+
+    Subclasses should supply a `mapped_model` attribute, which is the Pydantic model
+    which represents the final k8s resource.
+
+    Note, this class will validate that the annotated attributes map to real fields on the `mapped_model`.
+    """
+
+    mapped_model: ClassVar[Type[PydanticBaseModel]]
```
DanCardin, line 78: I perhaps think at least api_version,** and kind should be defined at the base-class level. It's not a huge deal, but it would save 2 lines in each of the existing subclasses, and generally simplify any external implementers to only those fields which are actually unique to that CRD

Although it will probably take some extra work specifically because of how i handled CronWorfklow being weird.

src/hera/workflows/resource_base.py
```diff
@@ -0,0 +1,197 @@
...
+
+from hera import _yaml
+from hera.shared import BaseMixin
+from hera.shared._pydantic import PydanticBaseModel, get_fields
+
+
+class ModelMapper:
+    """Maps a model attribute to the attribute path of the produced resource.
+
+    For example `foo: MopdelMapper("metadata.name")` would map the annotated object's
```
elliotgunton, line 30: Just a typo
```suggestion
    For example `foo: ModelMapper("metadata.name")` would map the annotated object's
```

src/hera/workflows/resource_base.py
```diff
@@ -0,0 +1,197 @@
...
+            if value is not None:
+                _set_model_attr(model, mapper.model_path, value)
+
+        return model
+
+
+class ResourceBase(BaseMixin):
+    """Base class for top-level Hera CRDs.
+
+    Subclasses should supply a `mapped_model` attribute, which is the Pydantic model
```
elliotgunton, line 71: This should probably be a private attribute, I'm curious why you changed from a mapper subclass used in the annotation to a `mapped_model` attribute?

src/hera/workflows/resource_base.py
```diff
@@ -0,0 +1,197 @@
...
+        super().__init_subclass__(**kwargs)
+
+    @classmethod
+    def _iter_model_mappers(cls, traverse_mro: bool = True):
+        mro = cls.__mro__
+        if not traverse_mro:
+            mro = (cls,)
+
+        # Due to child subclasses being able to override parent fields, traversing the mro might
+        # yield the same attribute more than once, which wont make sense. Track seen fields, and only
```
elliotgunton, line 99: I think this problem was solved using a `ChainMap` but I don't remember exactly

src/hera/workflows/resource_base.py
DanCardin: you mean the
```
class _WorkflowModelMapper(_ModelMapper):
    @classmethod
    def _get_model_class(cls) -> Type[BaseModel]:
        return _ModelWorkflow
```

pattern?

If so, it seemed like a very roundabout way of asserting that all the mappers for a given source class should use the same mapper.

* There was nothing ensuring an author actually used the subclass
* it was a decent amount of boilerplate. less important internally. but this change kind of intentionally makes this a public facing detail now.
* The logic associated with each subclass was the same, it seemed like the only purpose was to specify the desitnation model to map to.

---

By contrast (i think) having it be a class attribute helps in a few ways:

* It's an attribute of the model, so it automatically applies the same destination class to all attributes
* less boilerplate, 1 line configuration for a source class
* validation happens at class creation time. I personally think this is nicer than happening at `_WorkflowModelMapper` instantiation time. you have more information about the class being mapped for error messages and such

DanCardin: as for it being a private attribute, it would seem (to me) to be weird for it to be a private attribute. An external definer of a CRD would be required to make use of this field

```python
class MyCRD(ResourceBase):
    mapped_model: ClassVar[Type] = MyCRDModel

    foo: Annotated[str, ModelMapper('foo')]

class MyCRDModel(BaseModel):
    foo: str
```

elliotgunton: I think my intention about `mapped_model` being private was more the intention that you shouldn't be able to (and shouldn't need to) change it as it's an intrinsic property of your `ResourceBase`, so maybe a property would be more appropriate? Otherwise I can do this when defining a `Workflow`:
```py
with Workflow(mapped_model=IrrelevantModel):
    ...
```

DanCardin: It's a `ClassVar`, which will make pydantic ignore it as far as input parameters go. I'd sooner make it `_mapped_model` than make it a property personally, but I still dont personally think it **needs** to be private.

I could also rename it to something like `hera_model` or something so it's less likely to conflict with arbitrary downstream code.

src/hera/workflows/resource_base.py
DanCardin: jk, done now. it **does** simplify the impl back down. thanks for pointing that out!

DanCardin:
Fixed prior PR suggestions.

I also did end up relocating `api_version` and `kind` into `ResourceBase`. After the ChainMap changes, it was less fiddly than I was anticipating it being. It ultimately has marginal benefit given that all the existing hera objects inherit from `Worflow`, but I think it does actually make more sense there.

---

I **also** gave `build` a default implementation in `ResourceBase`. I feel like there's a world where the `Workflow` build content goes away enough to become default, but right now it's got some stuff in it that'd make that impractical.

src/hera/workflows/resource_base.py
DanCardin: It occurred to me that it might be...something idk...a viable alternative to define it as `__init_subclass__` arguments like `class Workflow(ResourceBase, mapped_model=_WorkflowModel):`

It does cut down on the necessary typing a little bit, and it pulls it out of the class body where, as your prior comments implied, it could be confused as a model input param.

However, `__init_subclass__` is basically opting out of type enforcement on that argument, and the value ultimately would need to be stashed on the type anyways. So it ends up being kind of a wash.

elliotgunton:
Hey @DanCardin - we really like the intention to allow arbitrary CRDs, but we're hesitant about the PR because of a few things -
* allowing arbitrary CRDs opens Hera up _a lot_ further than its intended domain of Workflows. We need to be very careful about inviting that level of support burden on us.
  * we also want to be crystal clear on K8s terminology and implementation details (e.g. `resource_base.py` would need to be shared for Argo Events/CD) - I don't think we're ready/aiming to be "the" K8s Python SDK.
* We would prefer to implement `ResourceBase` as an experimental feature without refactoring existing Workflow code for now - I would be happy to rework the Workflow code at a later date - as I don't think you would be able to make it "experimental" at the same time as refactoring, and it would limit the scope of the PR. You'd be able to do `if isinstance(item, Workflow) or isinstance(item, ResourceBase) and not isinstance(item, type):` for the CLI

Feel free to make a custom CLI/Hera plugin package to achieve the goal of arbitrary CRDs, but we don't think Hera itself is quite ready for it. So, for the time being, we'll be closing this PR. If you're able to, I'd recommend coming to our bi-weekly maintainer's meeting (this Friday 3PM UTC). You can also join the CNCF Slack and reach us in [#hera-argo-sdk](https://cloud-native.slack.com/archives/C03NRMD9KPY) to discuss https://communityinviter.com/apps/cloud-native/cncf

DanCardin:
My feeling was that while hera is certainly specializing in making the Workflows experience better, there's nothing inherently Workflows specific about plenty of what's going on internally with hera. Particularly if you dont call `execute` on the models, but even then really.

It's not obvious to me how this changeset adds much theoretical API surface to hera or implied commitment of support. all of the fields/methods being factored into `ResourceBase` were part of the public API of `Workflow` already.

But either way, without some sort of class handle for the CLI to know to handle as a "resource", I have to revert back to using external scripts to create the associated resources I ideally want hera emitting.

---

What I dont like about adding something completely separate as "experimental", is that a lot of the impl inside `ResourceBase` currently is specifically in support of the existing featureset, not for what I personally want to do.

So I would just anticipate it immediately drifting between the current impl and the "experimental" impl nothing would be actually using

---

If there's nothing functionally wrong implementation-wise with what's in the PR now (which there certainly might be, i dont know how much to trust your tests or not, that i've not broken something), then I dont see why this couldn't be thought of as a simplification/refactor of how the existing resources are defined

By making all the existing changes internal/implementation details:

```python
# _resource_base.py
class _ResourceBase(...):
    ...
```

And then separately (or not at all, and i'll just personally depend on the private impl 😬), expose a separate class which enables "experimental" external extension

```python
class ExperimentalResourceBase(_ResourceBase):
    pass
```

PR #944, opened 2024-01-27
Validate k8s resource requirements for CPU, Memory and Ephemeral Volumes

KengoA:
**Pull Request Checklist**
- [ ] Fixes https://github.com/argoproj-labs/hera/issues/943
- [ ] Tests added

**Description of PR**

Currently, resource requirements are only partially validated for binary units with `hera.workflows.validators.validate_storage_units()` function for string values. The current implementaion misses some edge cases where the input is missing numerial values and only the unit is provided (i.e. `"Ki"` or `"Mi"`), and the comparison between request and limit is done for int values for `cpu_request` and `cpu_limit`. This PR implements validation for the binary representation and adds comparison for requests and limits.

This PR 
- Refactors validation for binary units (renamed from `validate_storage_units`to `validate_binary_units`)
- Fixes float values not compared for CPU requests/limits
- Adds validation for decimal units representation of CPU requests/limits in string
- Adds converters from decimal and binary units to float values
- Adds comparison of decimal units (CPU) and binary units (Memory and Ephemeral Volumes) representations of resources


src/hera/workflows/validators.py
```diff
@@ -43,30 +43,47 @@ def validate_name(name: str, max_length: Optional[int] = None, generate_name: bo
     return name
 
 
-def validate_storage_units(value: str) -> None:
-    """Validates the units of the given value.
+def validate_binary_units(value: str) -> None:
```
KengoA, line 46: renamed from `validate_storage_units` to `validate_binary_units` as this applies to memory and ephemeral volumes

flaviuvadan: Unfortunately this helper is public and we don't know who might use it outside of the core of Hera. Any chance we can leave `validate_storage_units(value: str) -> None` as public and internally it passes the value to `validate_binary_units`? That way we get the new function with improved functionality while maintaining backwards compatibility

src/hera/workflows/validators.py
```diff
@@ -43,30 +43,47 @@ def validate_name(name: str, max_length: Optional[int] = None, generate_name: bo
...
+
+    Parameters
+    ----------
+    value: str
+        The value to validate the decimal unit of.
+
+    Raises:
+    ------
+    ValueError
+        When the identified unit is not a supported one. The supported units are ['m', 'k', 'M', 'G', 'T', 'P', 'E'].
```
flaviuvadan, line 83: > The supported units are ['m', 'k', 'M', 'G', 'T', 'P', 'E'].

Small - maybe this can go into the docstring of `value`

flaviuvadan (commented):
Thanks for the first time contribution @KengoA! This is really great :) Left some small feedback on backwards compatibility and docs. The rest looks great

src/hera/workflows/validators.py
KengoA: Thanks for your quick review!

In the commit below I kept the public function and use aliases to refer to the `_validate_binary_units`, which is now made private so that we can the implementaion later if necessary.

To avoid confusion and also to make the function names consistent, other function names are now named more explicitly for their use cases (`validate_storage_units`, `validate_cpu_units`, `validate_memory_units`) as well as converters.

https://github.com/argoproj-labs/hera/pull/944/commits/31184f84447e6129ab99bdf008306d84ab34b9d7

src/hera/workflows/validators.py
KengoA: Good point! I added supported units in docstrings for validators and converters.

elliotgunton (approved):
LGTM! Thank you @KengoA 🚀

PR #950, opened 2024-01-31
fix: Improve linting/editor experience for hera models.

DanCardin:
**Pull Request Checklist**
- [ ] Fixes #
- [ ] Tests added

**Description of PR**
Today, hera models seem to receive no editor/linter support for the `__init__` arguments of hera models.

This **seems** to be solely from the necessary try/except imports used to support both pydantic v1 and v2. Both mypy and pyright will bail on and not yield useful types when you do this. But it can be hacked around with some **more** conditionals imports 😅

Today:
Screenshot 2024-01-31 at 2 19 19 PM
Screenshot 2024-01-31 at 2 19 27 PM

With this change:
Screenshot 2024-01-31 at 2 18 27 PM
Screenshot 2024-01-31 at 2 18 36 PM

### DIFF
--- a/src/hera/shared/_global_config.py
+++ b/src/hera/shared/_global_config.py
@@ -7,7 +7,7 @@ from dataclasses import dataclass, field
 from typing import Any, Callable, Dict, List, Optional, Type, TypeVar, Union
 
 from hera.auth import TokenGenerator
-from hera.shared._pydantic import BaseModel, root_validator
+from hera.shared._pydantic import BaseModel, get_fields, root_validator
 
 TBase = TypeVar("TBase", bound="BaseMixin")
 TypeTBase = Type[TBase]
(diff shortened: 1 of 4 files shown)


src/hera/shared/_global_config.py
```diff
@@ -7,7 +7,7 @@
 from typing import Any, Callable, Dict, List, Optional, Type, TypeVar, Union
 
 from hera.auth import TokenGenerator
-from hera.shared._pydantic import BaseModel, root_validator
+from hera.shared._pydantic import BaseModel, get_fields, root_validator
```
DanCardin, line 10: this `get_fields` thing I did mainly to avoid littering all the existing raw usages of `__fields__` with `# type: ignore` because there's version specific linter error I was seeming to get.

DanCardin:
The failing py3.8 build gives me no logs, so i wonder if it's actually a transient CI issue?

And the coverage changes, I think are lies. There should be net no line coverage difference, as far as I can tell.

samj1912 (approved):

PR #963, opened 2024-02-16
fix(UserContainer): Fix a small bug in ImagePullPolicy mapper and add tests for error cases

KengoA:
**Pull Request Checklist**
- [ ] Fixes #
- [x] Tests added

**Description of PR**
When checking for string values of `image_pull_policy` in the `UserContainer` class, the current implementation was adding redundant items in the `policy_mapper` dict as it was confusing the `name` and `value` properties.

This PR simplifies this logic while covering the same use cases, and adds tests for normal and error cases.

### DIFF
--- a/src/hera/workflows/user_container.py
+++ b/src/hera/workflows/user_container.py
@@ -26,7 +26,7 @@ class UserContainer(_ModelUserContainer):
     volumes: Optional[List[_BaseVolume]] = None
 
     def _build_image_pull_policy(self) -> Optional[str]:
-        """Processes the image pull policy field and returns a generated `ImagePullPolicy` enum."""
+        """Processes the image pull policy field and optionally returns a string value from `ImagePullPolicy` enum."""
         if self.image_pull_policy is None:
             return None
         elif isinstance(self.image_pull_policy, ImagePullPolicy):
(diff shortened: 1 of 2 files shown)


src/hera/workflows/user_container.py
```diff
@@ -37,10 +37,8 @@ def _build_image_pull_policy(self) -> Optional[str]:
             # the following 2 are "normal" entries
             **{ipp.name: ipp for ipp in ImagePullPolicy},
             **{ipp.value: ipp for ipp in ImagePullPolicy},
-            # some users might submit the policy without underscores
-            **{ipp.value.lower().replace("_", ""): ipp for ipp in ImagePullPolicy},
-            # some users might submit the policy in lowercase
-            **{ipp.name.lower(): ipp for ipp in ImagePullPolicy},
```
KengoA, line 43: These were mostly redundant and created duplicate key-value pairs.

Reference:
```python3 
# src/hera/workflows/models/io/k8s/api/core/v1.py

class ImagePullPolicy(Enum):
    always = "Always"
    never = "Never"
    if_not_present = "IfNotPresent"
```

elliotgunton: Good find, thanks! cc @flaviuvadan JFYI 👀

elliotgunton (approved):
Thank you! 🚀

PR #974, opened 2024-02-22
Fix RunnerInput defaults

elliotgunton:
**Pull Request Checklist**
- [x] Fixes bug 1 from #962 
- [x] Tests added
- [ ] Documentation/examples added
- [x] [Good commit messages](https://cbea.ms/git-commit/) and/or PR title

**Description of PR**
Currently, defaults are required when using Parameters in the RunnerInput. This PR allows defaults to be omitted.

PR #982, opened 2024-02-26
Split up script user guide and fix pydantic io example

### LINKED ISSUE #961: RunnerInput: Got 400 from Argo, a value was not supplied in the parameter
Hello,

Thank you for this framework. I am facing an issue using "RunnerInput" with typical inputs i would like to structure.

### Issue 
The code is as below:
```python
class DefaultParamsModel(BaseModel):
   name: str = "default value"
   my_var: str = "another default str"

class MyDefaultInput (RunnerInput):
    my_env: Annotated[DefaultParamsModel, Parameter(name="scope")] = DefaultParamsModel(
        name="test",
        my_var="bla",
    )
    stdout: Annotated[Path, Parameter(global_name="stdout", name="stdout", output=True)]
    stderr: Annotated[Path, Parameter(global_name="stderr", name="stderr", output=True)]

[...]

@script(constructor="runner") 
def ctn_test(my_input: MyDefaultInput) -> MyDefaultOutput: 
   [my_code...]
   return MyDefaultOutput(exit_code=exit_code, result="???", stdout=my_stdout, stderr=my_stderr,...)
```

I would like to use the statement below mentioned in the documentation:

> Input-Output function parameters
Hera also allows output Parameter/Artifacts as part of the function signature when specified as a Path type, allowing users to write to the path as an output, without needing an explicit return. They require an additional field output=True to distinguish them from the input parameters and must have an underlying Path type (or another type that will write to disk).

I want to use annotation feature so I don't have to duplicate the declaration. In my context, the goal is to use a variable referenced in "inputs" (as parameter, so without having to create a "temporary" file), fill the content during execution, record it as artefact, and after the exit to be available to others tasks (as well as others output). 

1. How to use it with the RunnerInput class ? 
2. If i use an artifact declaration in the function signature, will it be recorded automatically in my registered hera s3 despite the fact that here I declare it as "Parameter" (throught the "stdout" variable, and it is not "written" by myself on the fs) ? 
3. Is it possible to inherit from an existing RunnerInput class to add extra field in a new class ? so I don't have to rewrite all the arguments, trying to do something like that (based on the sample above): 
```python
MySecondDefaultInput(MyDefaultInput): 
    new_var: Annotated[Path, Parameter(name="...", output=True)]
```
4. It's still unclear for me the purpose of "result" with "RunnerOutput". In which use-case I would use it ?  

### Additional information
I am using latest version, I have embedeed the python lib to be run on the image. All works fine for this side. However for the declaration, I am not sure what i am doing wrong here. 
Notes :
- Expiremental Pydantic is enabled
- Experimental Annotations is enabled
- Latest version (5.14.0) is used

Thank you for any helps/pointers.

elliotgunton:
* Script user guide was too long, split into main features. Fix internal links
* Make pydantic io example into a runnable workflow - made it more obvious the scripts would need a custom image

**Pull Request Checklist**
- [x] Fixes #961 (fixes example)
- [ ] Tests added

**Description of PR**
Currently, the pydantic IO example does not work. Combined with fixes in PRs #974 and #977, this doc change shows how to use the Runner IO features.


flaviuvadan (approved):
🎖️ this is huge and it looks awesome

PR #989, opened 2024-03-04
Fix `to` generating to folder and refactor tests

elliotgunton:
* Previously, if a user specifies a `to` option such as `foo.yaml`, the cli will create the target as a folder `foo.yaml/`. This change inspects the suffix of the `to` path; if it is a recognised "yaml" extension, we create the file's parents, and writes the workflow to the file. We therefore do not need to check if the target exists
* The commit also refactors some code for testability and simplifies the tests themselves for readability to avoid magic number accesses

### DIFF
--- a/src/hera/_cli/generate/yaml.py
+++ b/src/hera/_cli/generate/yaml.py
@@ -11,10 +11,13 @@ from typing import Generator, Iterable, List
 from hera._cli.base import GenerateYaml
 from hera.workflows.workflow import Workflow
 
+DEFAULT_EXTENSION = ".yaml"
 YAML_EXTENSIONS = {".yml", ".yaml"}
+
 
 def _write_workflow_to_yaml(target_file: Path, content: str):
     target_file.write_text(content)
+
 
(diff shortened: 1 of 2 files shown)


src/hera/_cli/generate/yaml.py
```diff
@@ -34,18 +38,19 @@ def generate_yaml(options: GenerateYaml):
...
-            options.to.write_text(output)
+            _write_workflow_to_yaml(options.to, output)
 
         else:
             os.makedirs(options.to, exist_ok=True)
 
             for dest_path, content in path_to_output:
-                full_path = (options.to / dest_path).with_suffix(".yaml")
-                full_path.write_text(content)
+                _write_workflow_to_yaml((options.to / dest_path).with_suffix(".yaml"), content)
```
flaviuvadan, line 53: Small suggestion - place the `".yaml."` in a constant above as the DEFAULT_EXTENSION

elliotgunton:
Very sensible suggestions though @flaviuvadan 💯 Also 🤦‍♂️ because I didn't format for CI

PR #991, opened 2024-03-05
Add RunnerException experimental feature

elliotgunton:
* Allows users to raise a custom error that will output values to Argo before forwarding the exception from the Hera runner

This features offers some improvements over the `_save_dummy_outputs` function added as part of #977. The `RunnerException` class is intended to be subclassed by users to let them raise a custom error, while still outputting values from the Argo Dag/Step node. This comes from some suggestions we had internally.

### DIFF
--- a/src/hera/workflows/runner.py
+++ b/src/hera/workflows/runner.py
@@ -1,6 +1,30 @@
 """The entrypoint for the Hera Runner when running on Argo."""
 
+import os
+from typing import Any, Tuple, Union
+
 from hera.workflows._runner.util import _run
+
+
+class RunnerException(Exception):
+    """An exception class that can hold outputs to write to file on Argo before passing the exception up.
+
+    Note: Must be enabled via hera.shared.global_config.experimental_features.
+    """
+
+    def __init__(self, outputs: Union[Tuple[Any], Any]) -> None:
+        """Initialise the RunnerException with outputs to be written that match the function output annotations."""
+        if os.environ.get("hera__script_runner_exception", None) is None:
+            raise ValueError(
+                (
+                    "Unable to instantiate {} since it is an experimental feature."
+                    " Please turn on experimental features by setting "
+                    '`hera.shared.global_config.experimental_features["{}"] = True`.'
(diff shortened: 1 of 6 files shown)


samj1912:
@elliotgunton what's the motivation behind this feature? Why/when would a user use this instead of runner output which has exit code and output fields already?

elliotgunton:
@samj1912 users would lose information about _what_ exception occurred by using the `RunnerOutput`. Instead if they have a top level `try`/`except` in their function, they can `raise RunnerException from e` and easily get the full stack trace output in the container logs, while also getting real outputs from the template.

Quite a niche use case so I'm fine to close if you don't think it adds much value and the usage is a bit convoluted - users should probably _not_ have a top level `try` 😅 and they can log the error stack trace manually if they want.

src/hera/workflows/runner.py
```diff
@@ -1,6 +1,31 @@
...
 from hera.workflows._runner.util import _run
 
+
+class RunnerException(Exception):
+    """An exception class that can hold outputs to write to file on Argo before passing the exception up.
+
+    Note: Must be enabled via hera.shared.global_config.experimental_features.
+    """
+
+    def __init__(self, outputs: Union[Tuple[Any], Any], exit_code: Optional[int] = None) -> None:
```
samj1912, line 15: We should keep exit code as the first param followed by details.

elliotgunton: What if I want the original exception (and exit code) to bubble up? i.e. in `src/hera/workflows/_runner/util.py`:

```py
            except RunnerException as e:
                _save_annotated_return_outputs(e.outputs, output_annotations)
                if e.exit_code is not None:
                    exit(e.exit_code)

                raise e
```

I wouldn't really say `exit_code` is necessary in `RunnerException` as then you should use `RunnerOutput`. If we remove it then we can avoid "two ways to do one thing".

elliotgunton:
Closing as the feature is not substantially different to using `RunnerIO` to do the same thing

PR #996, opened 2024-03-06
Tidy up runner tests

elliotgunton:
While working on #991 I found the runner tests quite messy. This PR cleans up the `test_runner.py` file (changes are in test files only).

* global_config doesn't need setting when using the runner, only the environment variables
* Use monkeypatch.setenv instead of os.environ
* Remove environ_annotations_fixture to make tests more explicit/obvious

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,5 +1,3 @@
-import os
-
 import pytest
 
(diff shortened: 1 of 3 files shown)


flaviuvadan (approved):
🎖️

PR #1018, opened 2024-04-03
Fix script-decorated-function typing

### LINKED ISSUE #1017: Script decorator breaks type hints
Related to [this issue](https://github.com/argoproj-labs/hera/issues/837) is the fact that the `script` decorator fudged up the type hints for the function. The intention of the `_take_annotation_from`, seen here https://github.com/argoproj-labs/hera/blob/91df4deb132e3c3f96aa0f31804e6fa175610a6f/src/hera/workflows/script.py#L611

was to allow the script decorator itself to set the `Script` class values, like `volumes` etc, which _does_ work:

image

_But_, the unintended side effect was the function itself lost type hints because I thought I needed to add the Script type hints at the call site as well, but actually this should be the type hints for Step/Task, e.g. `when` doesn't show up:

image

because it's using the `Script` kwargs for the hints, not `TemplateInvocatorSubNodeMixin` (i.e. Steps/Tasks) kwargs.

Furthermore, types hints for the function itself broke:

image

_Originally posted by @elliotgunton in https://github.com/argoproj-labs/hera/issues/837#issuecomment-2035015237_

elliotgunton:
* Function signature is now left intact when called elsewhere
* Script decorator still suggests Script kwargs
* Also allows call site typing to use Step/Task type hints

**Pull Request Checklist**
- [x] Fixes #1017 

**Description of PR**
See issue for "before" screenshots.

For "after":

Script decorator type hints:
image

Function call type hints:
image

Task/Step type hints (`arguments`, `when`, `name` etc):
image

### DIFF
--- a/src/hera/workflows/script.py
+++ b/src/hera/workflows/script.py
@@ -588,9 +588,9 @@ def _add_type_hints(
 ) -> Callable[
     ...,
     Callable[
-        ScriptIns,  # this adds Script type hints to the underlying function kwargs, i.e. `script`
+        ScriptIns,  # this adds Script type hints to the underlying *library* function kwargs, i.e. `script`
         Callable[  # we will return a function that is a decorator
-            [Callable[FuncIns, FuncR]],  # taking underlying function's inputs
+            [Callable[FuncIns, FuncR]],  # taking underlying *user* function
             Union[  # able to return FuncR | Step | Task | None
                 Callable[FuncIns, FuncR],
                 Callable[StepIns, Optional[Step]],
--- a/src/hera/workflows/script.py
+++ b/src/hera/workflows/script.py
@@ -20,7 +20,6 @@ from typing import (
     TypeVar,
     Union,
     cast,
-    overload,
 )
 
 from typing_extensions import ParamSpec, get_args, get_origin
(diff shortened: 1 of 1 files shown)


src/hera/workflows/script.py
```diff
@@ -662,16 +655,9 @@ def script_wrapper(
...
-        def task_wrapper(*args: FuncIns.args, **kwargs: FuncIns.kwargs) -> FuncR:  # type: ignore
-            ...
-
-        @overload
-        def task_wrapper(*args: ScriptIns.args, **kwargs: ScriptIns.kwargs) -> Union[Step, Task]: ...
-
         @wraps(func)
-        def task_wrapper(*args, **kwargs):
-            """Invokes a `Script` object's `__call__` method using the given `task_params`."""
+        def task_wrapper(*args, **kwargs) -> Union[FuncR, Step, Task, None]:
```
samj1912, line 659: Why did we remove wraps decorator? This sets a bunch of important attributes correctly.

elliotgunton: It's still there

image

samj1912 (approved):

PR #1033, opened 2024-04-13
HEP-0001: Add new decorator syntax

samj1912:
Adds a new decorator syntax HEP that allows us to call all kinds of template and even makes defining steps and dag templates easier.

This will largely involve an implementation on top of libraries like https://github.com/alexmojaki/executing#executing and https://pypi.org/project/varname/

The eventual goal of this syntax is to allow hera templates to be run locally as well. In order to do that, we will also need to provide appropriate local implementation of the `expr` library.

cc: @elliotgunton, @flaviuvadan 

[Readable version](https://github.com/argoproj-labs/hera/blob/decorator-syntax/hep/proposals/0001-decorators.md)

### DIFF
--- a/hep/proposals/0000-decorators.md
+++ b/hep/proposals/0000-decorators.md
@@ -143,7 +143,7 @@ class CalculatorInput(hio.Input):
 
 @wt.script  # Adds a new script template to the workflow template called calculator
 def calculator(calc_input: CalculatorInput) -> hio.Output:
-    if calc_input.operation == "add"
+    if calc_input.operation == "add":
         return hio.Output(calc_input.x + calc_input.y)
     return hio.Output(calc_input.x - calc_input.y)
     
--- a/hep/proposals/0000-decorators.md
+++ b/hep/proposals/0000-decorators.md
@@ -1,0 +1,299 @@
+# Meta
+[meta]: #meta
+- Name: Hera Decorator Syntax
+- Start Date: 2024-04-13
+- Author(s): [@samj1912](https://github.com/elliotgunton), [@elliotgunton](https://github.com/elliotgunton)
+- Supersedes: (put "N/A" unless this replaces an existing HEP, then link to that HEP)
+
+# Table of Contents
+[table-of-contents]: #table-of-contents
+- [Meta](#meta)
+- [Table of Contents](#table-of-contents)
+- [Overview](#overview)
+- [Motivation](#motivation)
+- [Proposal](#proposal)
+- [Code Examples](#code-examples)
+  - [Basic Workflow](#basic-workflow)
+  - [DAG Template](#dag-template)
+  - [Steps Template](#steps-template)
+  - [Container Template](#container-template)
+  - [Workflow Template Ref](#workflow-template-ref)
+  - [How to teach (OPTIONAL)](#how-to-teach-optional)
+- [Implementation (OPTIONAL)](#implementation-optional)
+  - [Link to the Implementation PR](#link-to-the-implementation-pr)
+- [Migration (OPTIONAL)](#migration-optional)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [Prior Art](#prior-art)
+- [Unresolved Questions (OPTIONAL)](#unresolved-questions-optional)
+
+# Overview
+[overview]: #overview
+
+This HEP introduces a new and powerful decorator based syntax for defining all kinds of Workflows and Templates.
+
+# Motivation
+[motivation]: #motivation
+
+The current context manager based syntax is good but it doesn't map well to template types that contain inputs/outputs. With the `@script` decorator, we have made hera feel extremely Pythonic and it has the added benefit of being runnable locally. We wish to extend this syntax to other template types which will allow for better readability and local testing and it will also standardize the template interface for Argo Workflows.
+
+# Proposal
+
+This HEP proposes that every template be mapped into a decorated function with HeraIO classes as the inputs and outputs.
+
+# Code Examples
+
+## Basic Workflow
+
+```python
+from hera.workflows import WorkflowTemplate
+from hera import io as hio
+
+# We start by defining our Workflow Template
+wt = WorkflowTemplate(name="my-template")
+
+# This defines the a template input
+class MyInput(hio.Input):
+    user: str
+
+@wt.entrypoint  # Sets hello_world as the default entrypoint for the workflow template
+@wt.script  # Adds a new script template to the workflow template called hello_world
+def hello_world(my_input: MyInput) -> hio.Output:
+    output = hio.Output()
+    output.result = f"Hello Hera User: {my_input.user}!"
+    return output
+
+# Allows users to instantiate a workflow template into a workflow
+workflow = wt.run(MyInput(name="happy-hera-user"), wait=True)
+# Returned workflow is a normal hera Workflow class and can be accessed as normal
+assert workflow.status.phase == "Succeeded"
+assert workflow.status.outputs.result == "Hello Hera User: happy-hera-user!"
+```
+
+## DAG Template
+
+```python
+from hera.workflows import WorkflowTemplate
+from hera import io as hio
+from hera.expr import spring
+
+# We start by defining our Workflow Template
+wt = WorkflowTemplate(name="my-template")
+
+@wt.script  # Adds a new script template to the workflow template
+def setup() -> hio.Output:
+    return hio.Output(result="Setting things up")
+
+class ConcatInput(hio.Input):
+    word_a: str
+    word_b: str
+
+@wt.script  # Adds a new script template to the workflow template
+def concat(concat_input: ConcatInput) -> hio.Output:
+    return hio.Output(result=f"{concat_input.word_a} {concat_input.word_b}")
+
+class WorkerInput(hio.Input):
+    value_a: str
+    value_b: str
+
+class WorkerOutput(hio.Output):
+    value: str
+
+
+# The great outcome of this new syntax is that the template definition
+# vs the template call becomes very explicit.
+# When a template is defined, it is defined via the decorator which registers it
+# When a template is invoked, it is invoked either via the callable syntax or via
+# wt.run() if it is the entrypoint
+# Another great side-effect is that all the templates can also be "run" locally as functions
+@wt.entrypoint
+@wt.dag
+def worker(worker_input: WorkerInput) -> WorkerOutput:
+    # We will need to use something like python-varname
+    # to automatically create a task called "setup_task" here.
+    setup_task = setup()
+    # Note how easy it is to reference variables from the DAG template input
+    # or previous tasks.
+    task_a = concat(ConcatInput(word_a=worker_input.value_a, word_b=setup_task.result))
+    task_b = concat(ConcatInput(word_a=worker_input.value_b, word_b=setup_task.result))
+    final_task = concat(ConcatInput(word_a=task_a.result, word_b=task_b.result))
+    # Note, we will automatically construct the dependency graph
+    # based on the input/output relationship between tasks.
+    # Users will still be able to explicitly add dependencies between tasks
+    # using the rshift operator if they wish to define dependencies amongst
+    # tasks that don't share variables directly
+    return WorkerOutput(value=final_task.result)
+```
+
+
+## Steps Template
+
+```python
+from hera.workflows import WorkflowTemplate
+from hera import io as hio
+
+# We start by defining our Workflow Template
+wt = WorkflowTemplate(name="my-template")
+
+# This defines the a template input
+class CalculatorInput(hio.Input):
+    x: int
+    y: int
+    operation: Literal["add", "sub"] = "add"
+
+@wt.script  # Adds a new script template to the workflow template called calculator
+def calculator(calc_input: CalculatorInput) -> hio.Output:
+    if calc_input.operation == "add"
+        return hio.Output(calc_input.x + calc_input.y)
+    return hio.Output(calc_input.x - calc_input.y)
+    
+
+# This defines the another template input
+class FiboInput(hio.Input):
+    num: int
+
+class FiboOutput(hio.Output):
+    num: int
+
+@wt.entrypoint
+@wt.steps  # Adds a new script template to the workflow template called fibonacci
+def fibonacci(fibo: FiboInput) -> FiboOutput:
+    # we will need to validate that users can only run/define certain kinds of expressions here
+    # and throw warnings if they are using syntax that is invalid
+    valid_number = hio.expr(fibo.num > 1)
+    previous_num = calculator(CalculatorInput(x=fibo.num, y=1, operation="sub"), when=valid_number)
+    previous_fibo = fibonacci(FiboInput(num=previous_num.result), when=valid_number)
+    current = calculator(CalculatorInput(x=fibo.num, y=previous_fibo.num), when=valid_number)
+    return FiboOutput(num=valid_number.check(current.num, 1))
+```
+
+
+
+## Container Template
+
+```python
+from hera.workflows import WorkflowTemplate, Container
+from hera import io as hio
+
+# We start by defining our Workflow Template
+wt = WorkflowTemplate(name="my-template")
+
+# This defines the a template input
+class MyInput(hio.Input):
+    user: str = "Hera"
+
+class MyOutput(hio.Output):
+    container_greeting: Annotated[str, Parameter(name="container-greeting")]
+
+@wt.entrypoint
+@wt.container(command=["sh", "-c"], args=["echo {{input.parameters.user}}"])
+def basic_hello_world(my_input: MyInput) -> hio.Output:
+    ...
+
+
+@wt.entrypoint
+@wt.container(command=["sh", "-c"])
+def advanced_hello_world(my_input: MyInput, template: Container) -> MyOutput:
+    # hio.output serves two functions. One, it allows users to create the output class
+    # which allows output path variables to be referenced.
+    # Second, it can be used to "mock" a container template when running locally and allow us to
+    # inject outputs based on inputs.
+    output: MyOutput = hio.output(my_input)
+    # template is a special variable that allows you to reference the template itself and modify
+    # its attributes. This is especially useful when trying to reference input params in args
+    template.args = [f"echo {hio.name(my_input.user)} > {hio.path(output.container_greeting)}"]
+    return output
+```
+
+## Workflow Template Ref
+
+```python
+from hera.workflows import WorkflowTemplate
+from hera import io as hio
+from hera.expr import spring
+
+# We start by defining our Workflow Template
+wt = WorkflowTemplate(name="my-template")
+
+ewt = WorkflowTemplate(name="externally-workflow-template")
+awt = WorkflowTemplate(name="another-workflow-template")
+
+@awt.script  # Adds a new script template to a workflow template called "another-workflow-template"
+def setup() -> hio.Output:
+    return hio.Output(result="Setting things up")
+
+class ConcatInput(hio.Input):
+    word_a: str
+    word_b: str
+
+
+class ConcatOutput(hio.Output):
+    value: str
+
+# Adds a new template ref from an external template
+# This can be auto-generated or typed by the user
+@ewt.template_ref
+def concat(concat_input: ConcatInput) -> ConcatOutput:
+    ...
+    # Note: we can optionally use hio.output(concat_input)
+    # again here to run "local" mock versions of this workflow template ref
+    # this will be useful in local testing
+
+class WorkerInput(hio.Input):
+    value_a: str
+    value_b: str
+
+class WorkerOutput(hio.Output):
+    value: str
+
+
+@wt.entrypoint
+@wt.dag
+def worker(worker_input: WorkerInput) -> WorkerOutput:
+    # Note that here setup will be referenced as a workflow template ref
+    # since it belongs in a different workflow template even though it is defined
+    # inside this file
+    setup_task = setup()
+    # concat is a workflow template ref and just serves as a stub
+    task_a = concat(ConcatInput(word_a=worker_input.value_a, word_b=setup_task.result))
+    task_b = concat(ConcatInput(word_a=worker_input.value_b, word_b=setup_task.result))
+    final_task = concat(ConcatInput(word_a=task_a.value, word_b=task_b.value))
+    return WorkerOutput(value=final_task.value)
+```
+
+
+
+## How to teach (OPTIONAL)
+
+If applicable, describe the differences between teaching this to existing users and new users.
+
+
+# Implementation (OPTIONAL)
+
+TODO: Add links about https://github.com/alexmojaki/sorcery#select_from, https://github.com/alexmojaki/executing#libraries-that-use-this and https://github.com/pwwang/python-varname
+
+## Link to the Implementation PR
+
+# Migration (OPTIONAL)
+
+This section should document breaks to public API and breaks in compatibility due to this HEP's proposed changes. In addition, it should document the proposed steps that one would need to take to work through these changes.
+
+# Drawbacks
+
+Why should we **not** do this?
+
+# Alternatives
+
+- What other designs have been considered?
+- Why is this proposal the best?
+- What is the impact of not doing this?
+
+# Prior Art
+
+Discuss prior art, both the good and bad.
+
+# Unresolved Questions (OPTIONAL)
+
+- What parts of the design do you expect to be resolved before this gets merged?
+- What parts of the design do you expect to be resolved through implementation of the feature?
+- What related issues do you consider out of scope for this HEP that could be addressed in the future independently of the solution that comes out of this HEP?
--- a/hep/proposals/0001-decorators.md
+++ b/hep/proposals/0001-decorators.md
@@ -35,7 +35,7 @@ This HEP introduces new and powerful decorator based syntax elements for definin
 # Motivation
 [motivation]: #motivation
 
-The current context manager based syntax is good (especially for bridging the gap from Python to Argo's YAML) but it doesn't map well to non-script template types that contain inputs/outputs like DAGs and Steps. It is also difficult for users to pass Parameters and Artifacts between Tasks and Steps.
+The current context manager based syntax is good (especially for bridging the gap from Python to Argo's YAML) but it doesn't map well to non-script template types like DAGs and Steps that can also have inputs and outputs. It is also difficult for users to pass Parameters and Artifacts between Tasks and Steps.
 
 With the `@script` decorator, we made Hera feel extremely Pythonic and it has the added benefit of being runnable locally. We wish to extend this syntax to other template types, namely DAGs, Steps and Containers, (and eventually other template types) which will allow for better readability and local testing. In this HEP, we also show how we will be able to generate stubs for WorkflowTemplates and improvements for referencing TemplateRefs under the new decorator syntax.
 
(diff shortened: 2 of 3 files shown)


hep/proposals/0000-decorators.md
```diff
@@ -0,0 +1,299 @@
...
+
+# This defines the a template input
+class CalculatorInput(hio.Input):
+    x: int
+    y: int
+    operation: Literal["add", "sub"] = "add"
+
+@wt.script  # Adds a new script template to the workflow template called calculator
+def calculator(calc_input: CalculatorInput) -> hio.Output:
+    if calc_input.operation == "add"
```
elliotgunton, line 146:
```suggestion
    if calc_input.operation == "add":
```
just to fix syntax highlighting

hep/proposals/0000-decorators.md
```diff
@@ -0,0 +1,299 @@
...
+
+# Overview
+[overview]: #overview
+
+This HEP introduces a new and powerful decorator based syntax for defining all kinds of Workflows and Templates.
+
+# Motivation
+[motivation]: #motivation
+
+The current context manager based syntax is good but it doesn't map well to template types that contain inputs/outputs. With the `@script` decorator, we have made hera feel extremely Pythonic and it has the added benefit of being runnable locally. We wish to extend this syntax to other template types which will allow for better readability and local testing and it will also standardize the template interface for Argo Workflows.
```
elliotgunton, line 38:
```suggestion
The current context manager based syntax is good (especially for bridging the gap from Python to Argo's YAML) but it doesn't map well to template types that contain inputs/outputs. With the `@script` decorator, we made Hera feel extremely Pythonic and it has the added benefit of being runnable locally. We wish to extend this syntax to other template types, namely DAGs, Steps, Containers and TemplateRefs, which will allow for better readability and local testing.
```

Just some tidy up and adding colour. 

On
>and it will also standardize the template interface for Argo Workflows.

I _think_ you mean "template interface for Hera"? But I would remove this bit for now as there are other template types we're not addressing here (HTTP/resource).

hep/proposals/0000-decorators.md
```diff
@@ -0,0 +1,299 @@
...
+
+This HEP proposes that every template be mapped into a decorated function with HeraIO classes as the inputs and outputs.
+
+# Code Examples
+
+## Basic Workflow
+
+```python
+from hera.workflows import WorkflowTemplate
+from hera import io as hio
```
elliotgunton, line 50: Should be `hera.workflows.io`? The `io` classes are templates-specific and eventually `hera.events` might have something `io`

hep/proposals/0000-decorators.md
```diff
@@ -0,0 +1,299 @@
...
+# We start by defining our Workflow Template
+wt = WorkflowTemplate(name="my-template")
+
+# This defines the a template input
+class MyInput(hio.Input):
+    user: str
+
+@wt.entrypoint  # Sets hello_world as the default entrypoint for the workflow template
+@wt.script  # Adds a new script template to the workflow template called hello_world
+def hello_world(my_input: MyInput) -> hio.Output:
```
elliotgunton, line 61: I would think these `wt.scripts` _have_ to use the runner? Omitted for brevity or should `constructor="runner"` be somewhere in the code (and called out in the HEP)?

hep/proposals/0000-decorators.md
```diff
@@ -0,0 +1,299 @@
...
+# When a template is defined, it is defined via the decorator which registers it
+# When a template is invoked, it is invoked either via the callable syntax or via
+# wt.run() if it is the entrypoint
+# Another great side-effect is that all the templates can also be "run" locally as functions
+@wt.entrypoint
+@wt.dag
+def worker(worker_input: WorkerInput) -> WorkerOutput:
+    # We will need to use something like python-varname
+    # to automatically create a task called "setup_task" here.
+    setup_task = setup()
```
elliotgunton, line 114: Would be worth pointing out `setup_task` will be an `hio.Output` when running locally/testing, but then a `Task` when building the workflow - and that we can only realistically do this by limiting inputs/outputs to the Hera IO classes

hep/proposals/0000-decorators.md
```diff
@@ -0,0 +1,299 @@
...
+# This defines the a template input
+class MyInput(hio.Input):
+    user: str = "Hera"
+
+class MyOutput(hio.Output):
+    container_greeting: Annotated[str, Parameter(name="container-greeting")]
+
+@wt.entrypoint
+@wt.container(command=["sh", "-c"], args=["echo {{input.parameters.user}}"])
+def basic_hello_world(my_input: MyInput) -> hio.Output:
```
elliotgunton, line 190: Just to make sure the example is complete - looking at the `container-greeting` param, looks like it needs a `value_from`, and the command should be writing to that path (as we don't have the auto-IO stuff from the runner) - using `tee` as seen in upstream examples:

```suggestion
class MyOutput(hio.Output):
    container_greeting: Annotated[
        str,
        Parameter(
            name="container-greeting",
            value_from={"path": "/tmp/hello_world.txt"},
        ),
    ]

@wt.entrypoint
@wt.container(command=["sh", "-c"], args=["echo {{input.parameters.user}} | tee /tmp/hello_world.txt"])
def basic_hello_world(my_input: MyInput) -> hio.Output:
```

Adding the whole
```py
        Parameter(
            name="container-greeting",
            value_from={"path": "/tmp/hello_world.txt"},
        ),
```
looks to be annoying though

hep/proposals/0000-decorators.md
```diff
@@ -0,0 +1,299 @@
...
+    ...
+
+
+@wt.entrypoint
+@wt.container(command=["sh", "-c"])
+def advanced_hello_world(my_input: MyInput, template: Container) -> MyOutput:
```
elliotgunton, line 196: Same here in terms of specifying output path - I see `hio.path` below but a user probably doesn't want to start interpolating strings in a simple container command.

hep/proposals/0000-decorators.md
```diff
@@ -0,0 +1,299 @@
...
+    # hio.output serves two functions. One, it allows users to create the output class
+    # which allows output path variables to be referenced.
+    # Second, it can be used to "mock" a container template when running locally and allow us to
+    # inject outputs based on inputs.
+    output: MyOutput = hio.output(my_input)
+    # template is a special variable that allows you to reference the template itself and modify
+    # its attributes. This is especially useful when trying to reference input params in args
+    template.args = [f"echo {hio.name(my_input.user)} > {hio.path(output.container_greeting)}"]
```
elliotgunton, line 204: I think then you could literally run the command through Python? Maybe we should expose that somehow - as in, if there's no implementation when running locally, just do

```py
subprocess.run(template.args)
```

Also might be good to point out for the reader that mocking would be good in the case of external calls to avoid maintaining two sets of code.

hep/proposals/0000-decorators.md
```diff
@@ -0,0 +1,299 @@
...
+
+```python
+from hera.workflows import WorkflowTemplate
+from hera import io as hio
+from hera.expr import spring
+
+# We start by defining our Workflow Template
+wt = WorkflowTemplate(name="my-template")
+
+ewt = WorkflowTemplate(name="externally-workflow-template")
```
elliotgunton, line 218:
```suggestion
ewt = WorkflowTemplate(name="external-workflow-template")
```

hep/proposals/0000-decorators.md
```diff
@@ -0,0 +1,299 @@
...
+
+class ConcatOutput(hio.Output):
+    value: str
+
+# Adds a new template ref from an external template
+# This can be auto-generated or typed by the user
+@ewt.template_ref
+def concat(concat_input: ConcatInput) -> ConcatOutput:
+    ...
+    # Note: we can optionally use hio.output(concat_input)
```
elliotgunton, line 238:
```suggestion
    # Note: we can optionally return hio.output(concat_input)
```

hep/proposals/0000-decorators.md
```diff
@@ -0,0 +1,299 @@
...
+    value_b: str
+
+class WorkerOutput(hio.Output):
+    value: str
+
+
+@wt.entrypoint
+@wt.dag
+def worker(worker_input: WorkerInput) -> WorkerOutput:
+    # Note that here setup will be referenced as a workflow template ref
```
elliotgunton, line 253:
```suggestion
    # Here, `setup` will be referenced as a workflow template ref
```

hep/proposals/0000-decorators.md
```diff
@@ -0,0 +1,299 @@
...
+    word_a: str
+    word_b: str
+
+
+class ConcatOutput(hio.Output):
+    value: str
+
+# Adds a new template ref from an external template
+# This can be auto-generated or typed by the user
+@ewt.template_ref
```
elliotgunton, line 235: Would `cluster_scope` go in the decorator args, or be derived from the class referenced (i.e. `WorkflowTemplate`/`ClusterWorkflowTemplate`)?

hep/proposals/0000-decorators.md
```diff
@@ -0,0 +1,299 @@
...
+This HEP introduces a new and powerful decorator based syntax for defining all kinds of Workflows and Templates.
+
+# Motivation
+[motivation]: #motivation
+
+The current context manager based syntax is good but it doesn't map well to template types that contain inputs/outputs. With the `@script` decorator, we have made hera feel extremely Pythonic and it has the added benefit of being runnable locally. We wish to extend this syntax to other template types which will allow for better readability and local testing and it will also standardize the template interface for Argo Workflows.
+
+# Proposal
+
+This HEP proposes that every template be mapped into a decorated function with HeraIO classes as the inputs and outputs.
```
elliotgunton, line 42:
```suggestion
This HEP proposes that the DAG, Steps, Container templates and TemplateRefs be mapped into a decorated function with HeraIO classes as the inputs and outputs.
```

hep/proposals/0000-decorators.md
```diff
@@ -0,0 +1,299 @@
...
+    template.args = [f"echo {hio.name(my_input.user)} > {hio.path(output.container_greeting)}"]
+    return output
+```
+
+## Workflow Template Ref
+
+```python
+from hera.workflows import WorkflowTemplate
+from hera import io as hio
+from hera.expr import spring
```
elliotgunton, line 213:
```suggestion
```

hep/proposals/0000-decorators.md
```diff
@@ -0,0 +1,299 @@
...
+assert workflow.status.phase == "Succeeded"
+assert workflow.status.outputs.result == "Hello Hera User: happy-hera-user!"
+```
+
+## DAG Template
+
+```python
+from hera.workflows import WorkflowTemplate
+from hera import io as hio
+from hera.expr import spring
```
elliotgunton, line 78: "spring" lol - unused
```suggestion
```

hep/proposals/0000-decorators.md
elliotgunton: Discussed and decided to remove the `template_ref` decorator from the HEP

hep/proposals/0000-decorators.md
```diff
@@ -0,0 +1,299 @@
+# Meta
```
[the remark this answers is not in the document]
elliotgunton: Yes, updated lol

hep/proposals/0001-decorators.md
```diff
@@ -0,0 +1,313 @@
...
+This HEP introduces a new and powerful decorator based syntax for defining all kinds of Workflows and Templates.
+
+# Motivation
+[motivation]: #motivation
+
+The current context manager based syntax is good (especially for bridging the gap from Python to Argo's YAML) but it doesn't map well to template types that contain inputs/outputs. With the `@script` decorator, we made Hera feel extremely Pythonic and it has the added benefit of being runnable locally. We wish to extend this syntax to other template types, namely DAGs, Steps and Containers, which will allow for better readability and local testing. In this HEP, we also show how we will be able to generate stubs for WorkflowTemplates and improvements for referencing TemplateRefs under the new decorator syntax.
+
+# Proposal
+
+This HEP proposes that Argo's DAG, Steps and Container templates be mapped into Hera as decorated functions with HeraIO classes as the inputs and outputs.
```
flaviuvadan, line 42: Is is possible to introduce this without HeraIO classes?

hep/proposals/0001-decorators.md
```diff
@@ -0,0 +1,313 @@
...
+    value: str
+
+
+# The great outcome of this new syntax is is the separation of concerns so that
+# the template definition vs the template call becomes very explicit.
+# When a template is defined, it is defined via the decorator which registers it
+# When a template is invoked, it is invoked either via the callable syntax or via
+# wt.run() if it is the entrypoint
+# Another great side-effect is that all the templates can run locally as functions
+@wt.entrypoint
```
flaviuvadan, line 109: Is there a design principle that clearly delineates what will be available as a decorator and what won't?

hep/proposals/0001-decorators.md
```diff
@@ -0,0 +1,313 @@
...
+        ),
+    ]
+
+@wt.entrypoint
+@wt.container(command=["sh", "-c"], args=["echo {{input.parameters.user}} | tee /tmp/hello_world.txt"])
+def basic_hello_world(my_input: MyInput) -> hio.Output:
+    ...
+
+
+@wt.entrypoint
```
flaviuvadan, line 200: I assume this would raise an error bc there cannot be 2 entrypoints

hep/proposals/0001-decorators.md
```diff
@@ -0,0 +1,313 @@
...
+def worker(worker_input: WorkerInput) -> WorkerOutput:
+    # Here, the call to `setup` will create a template ref task since it belongs
+    # in a different workflow template even though it is defined inside this file
+    setup_task = setup()
+
+    # concat serves as a stub, and will also become a template ref task
+    task_a = concat(ConcatInput(word_a=worker_input.value_a, word_b=setup_task.result))
+    task_b = concat(ConcatInput(word_a=worker_input.value_b, word_b=setup_task.result))
+    final_task = concat(ConcatInput(word_a=task_a.value, word_b=task_b.value))
+    return WorkerOutput(value=final_task.value)
```
flaviuvadan, line 275: Since this implicitly sets up the dependencies based on I/O it is now equivalent to a Steps context in terms of the implicit dependencies. That might cause some conceptual tension. Can you explain a bit about how this does not conflict with Steps?

hep/proposals/0001-decorators.md
```diff
@@ -0,0 +1,313 @@
...
+
+# Alternatives
+
+- What other designs have been considered?
+- Why is this proposal the best?
+- What is the impact of not doing this?
+
+# Prior Art
+
+Discuss prior art, both the good and bad.
```
flaviuvadan, line 307: There might be specific elements of the Prefect syntax this is similar to, particularly the DAG worker examples. Might be good to share them here and explain how this improves that experience

samj1912:
An experimental PoC demonstrating how this can be possible - https://github.com/samj1912/varname-experiments/blob/main/experiments.py

hep/proposals/0001-decorators.md
samj1912: Unfortunately not. We went through a lot discussion to see if we could eliminate it but it's not possible. There are multiple reasons but primarily this allows us to -
- differentiate task inputs from task params
- allows us to dynamically switch get attribute access to pass appropriate varnames around.

hep/proposals/0000-decorators.md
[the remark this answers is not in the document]
samj1912: It shouldn't be. The filename starts with 0000 and is updated with a number on merge. The reason being that we don't know the sequence until things are merged.

hep/proposals/0000-decorators.md
samj1912: The decorator via the template class needs to enforce it. There are no other options allowed.

hep/proposals/0001-decorators.md
samj1912: Currently every template type and certain properties associated with the template would be decorators. I believe entrypoint is the only one aside from various template types currently.

hep/proposals/0000-decorators.md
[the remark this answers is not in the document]
samj1912: Will add it to the implementation section since it will be tied to how we implement it.

hep/proposals/0001-decorators.md
samj1912: It's not equivalent. Steps will still have a parallel context manager and will be executed in sequence. For DAGs, some of the dependencies can be implicitly discovered by us, but the user can also specify them.

The same argument can be put forward for steps vs dags in Argo Workflows. You can convert every DAG into step and the choice is largely upto the user.

hep/proposals/0001-decorators.md
elliotgunton: To explain more of the thinking behind the proposal, I'm not sure there's a particular design principle we're using (maybe DRY or dependency inversion?) but the intention is to avoid multiples places where we're performing definition logic, while keeping in mind a user-friendly and well-scoped way of writing WorkflowTemplates:

1. Declare WorkflowTemplate
2. Add templates at definition via decorator

As opposed to:
1. Write "free" (as in unbound) template definitions
2. Declare WorkflowTemplate
3. Add the templates

Which means anything we want to declare as part of the WorkflowTemplate needs to be from that local scope and not "forward declaring," so for entrypoint we want to avoid:
```py
wt = WorkflowTemplate(..., entrypoint="my-template")
```

As Sam said, seems like entrypoint is the only attribute that refers to something "yet to be created."

hep/proposals/0001-decorators.md
elliotgunton: Added a parallel steps example for more colour, and explained the auto-dag-construction in the DAG example. Agree with Sam that it comes down to Argo Workflows itself and the user's preference

hep/proposals/0001-decorators.md
```diff
@@ -0,0 +1,608 @@
...
+
+class FiboOutput(hio.Output):
+    num: int
+
+@wt.entrypoint
+@wt.steps  # Adds a new script template to the workflow template called fibonacci
+def fibonacci(fibo: FiboInput) -> FiboOutput:
+    # we will need to validate that users can only run/define certain kinds of expressions here
+    # and throw warnings if they are using syntax that is invalid
+    valid_number = hio.expr(fibo.num > 1)
```
mshatkhin23, line 179: will expressions work for local execution? For both `when` clauses and argument construction?

samj1912: We want to support this, yes. i.e. allow for local expression evaluation. May not make the first cut implementation though.

hep/proposals/0001-decorators.md
```diff
@@ -0,0 +1,611 @@
...
+
+@wt.entrypoint  # Sets hello_world as the default entrypoint for the workflow template, it is an error to set entrypoint on multiple functions
+@wt.script  # Adds a new script template to the workflow template called hello_world. This will enforce use of the script runner.
+def hello_world(my_input: MyInput) -> hio.Output:  # A subclass of hio.Output must be used for the output of the function
+    output = hio.Output()
+    output.result = f"Hello Hera User: {my_input.user}!"
+    return output
+
+# `run` is a new function for users to instantiate a workflow template as a workflow
+workflow = wt.run(MyInput(name="happy-hera-user"), wait=True)
```
elliotgunton, line 70: @samj1912 We already have `create_as_workflow` - https://hera.readthedocs.io/en/stable/walk-through/advanced-hera-features/#submit-workflowtemplates-and-clusterworkflowtemplates-as-workflows

Also mentioned in https://hera.readthedocs.io/en/stable/user-guides/core-concepts/#the-workflowtemplate

We should make sure `Workflow.create` is able to take arguments to the entrypoint too.

hep/proposals/0001-decorators.md
```diff
@@ -0,0 +1,611 @@
...
+import hera.workflows.io as hio
+
+# We start by defining our Workflow Template
+wt = WorkflowTemplate(name="my-template")
+
+# Users must subclass `hio.Input` to define the template's inputs
+class MyInput(hio.Input):
+    user: str
+
+@wt.entrypoint  # Sets hello_world as the default entrypoint for the workflow template, it is an error to set entrypoint on multiple functions
```
elliotgunton, line 62: `entrypoint` is already a property of `Workflow` - do you see a sensible way to refactor and allow this decorator name or should we just use `@wt.set_entrypoint`?

samj1912: It should be possible to make entrypoint a setter and then have dynamic behavior based on the input type.

hep/proposals/0001-decorators.md
```diff
@@ -0,0 +1,611 @@
...
+def basic_hello_world(my_input: MyInput) -> hio.Output:
+    ...
+
+
+@wt.entrypoint
+@wt.container(command=["sh", "-c"])
+def advanced_hello_world(my_input: MyInput, template: Container) -> MyOutput:
+    # A 'MyOutput' object can be used to "mock" the output when running locally and allow us to
+    # inject outputs based on inputs.
+    output: MyOutput = MyOutput(my_input)
```
samj1912, line 268: This is invalid as it shouldn't be the user defined type taking in another one. This was originally a `hio.output` function.

elliotgunton (approved):
Approving for FCP

samj1912:
FCP ends April 25th.

hep/proposals/0001-decorators.md
```diff
@@ -0,0 +1,656 @@
...
+
+@wt.entrypoint  # Sets hello_world as the default entrypoint for the workflow template, it is an error to set entrypoint on multiple functions
+@wt.script  # Adds a new script template to the workflow template called hello_world. This will enforce use of the script runner.
+def hello_world(my_input: MyInput) -> hio.Output:  # A subclass of hio.Output must be used for the output of the function
+    output = hio.Output()
+    output.result = f"Hello Hera User: {my_input.user}!"
+    return output
+
+# `run` is a new function for users to instantiate a workflow template as a workflow
+workflow = wt.run(MyInput(name="happy-hera-user"), wait=True)
```
Pierre-Brunelle, line 99: What is the purpose of the wait parameter?

Where does this execute the workflow (locally or on a pre-configured Argo cluster)?

hep/proposals/0001-decorators.md
```diff
@@ -0,0 +1,656 @@
...
+
+class WorkerInput(hio.Input):
+    value_a: str
+    value_b: str
+
+class WorkerOutput(hio.Output):
+    value: str
+
+
+# The great outcome of this new syntax is is the separation of concerns so that
```
Pierre-Brunelle, line 135:
```suggestion
# The great outcome of this new syntax is the separation of concerns so that
```

hep/proposals/0001-decorators.md
```diff
@@ -0,0 +1,656 @@
...
+# the template definition vs the template call becomes very explicit.
+# When a template is defined, it is defined via the decorator which registers it
+# When a template is invoked, it is invoked either via the callable syntax or via
+# wt.run() if it is the entrypoint
+# Another great side-effect is that all the templates can run locally as functions
+@wt.entrypoint
+@wt.dag
```
Pierre-Brunelle, line 142: To be consistent with the way you have defined script templates so far:

```suggestion
@wt.dag  # Adds a new DAG template to the workflow template
```

hep/proposals/0001-decorators.md
```diff
@@ -0,0 +1,656 @@
...
+
+# This defines another template's inputs
+class FiboInput(hio.Input):
+    num: int
+
+class FiboOutput(hio.Output):
+    num: int
+
+@wt.entrypoint
+@wt.steps  # Adds a new script template to the workflow template called fibonacci
```
Pierre-Brunelle, line 207: Did you mean?

```suggestion
@wt.steps  # Adds a new Steps template to the workflow template called fibonacci
```

hep/proposals/0001-decorators.md
```diff
@@ -0,0 +1,656 @@
...
+def fibonacci(fibo: FiboInput) -> FiboOutput:
+    # we will need to validate that users can only run/define certain kinds of expressions here
+    # and throw warnings if they are using syntax that is invalid
+    valid_number = hio.expr(fibo.num > 1)
+    previous_num = calculator(CalculatorInput(x=fibo.num, y=1, operation="sub"), when=valid_number)
+    previous_fibo = fibonacci(FiboInput(num=previous_num.result), when=valid_number)
```
Pierre-Brunelle, line 213: (discussion) How does this work if we run the Workflow on Argo? Is the idea that statically we figure out what the Workflow should look like? Or can we have an Argo Workflow with a dynamic number of nodes?

hep/proposals/0001-decorators.md
```diff
@@ -0,0 +1,656 @@
...
+    def steps(**steps_kwargs) -> Callable:
+        ...
+
+    def script(**script_kwargs) -> Callable:
+        ...
+```
+
+We will enforce the single input and output of the function within the decorator to be the `hera.workflows.io.Input` and `hera.workflows.io.Output` classes. We will repurpose the RunnerInput/RunnerOutput classes and deprecate the `script_pydantic_io` experimental feature, as we will stop development on the old `script` decorator to instead promote the new decorators as "the golden path" for development with Hera.
+
+We require the use of special classes for the input/output of the function to allow a mechanism to switch between "build" code for when building a Workflow, versus local "running" code. Using a custom class means we can implement dunder methods like `__getattribute__` to intercept the call so that for a statement within a DAG/Steps function like from the DAG example:
```
Pierre-Brunelle, line 419: > to intercept the call so that for a statement within a DAG/Steps function like from the DAG example

I have trouble understanding this phrase.

hep/proposals/0001-decorators.md
```diff
@@ -0,0 +1,656 @@
...
+```
+
+we need to make use of libraries that can perform AST inspection. `sorcery` introduces the ability to get the name of the variable itself through the `assigned_names` [function](https://github.com/alexmojaki/sorcery?tab=readme-ov-file#assigned_names). A library that builds off this is [python-varname](https://github.com/pwwang/python-varname), which is able to retrieve the name of the variable being assigned to by a function, from _inside_ the function:
+
+```py
+from varname import varname
+def function():
+    return varname()
+
+func = function()  # func == 'func'
```
Pierre-Brunelle, line 453: Pretty neat! Just beware of corner cases:

```py
a = b = function()
# MultiTargetAssignmentWarning: Multiple targets in assignment, variable name on the very right is used.
```

Pierre-Brunelle (commented):
Looks great Mr. Gunton!

hep/proposals/0001-decorators.md
samj1912: It's static. The idea is that we do source code analysis to compile a supported subset of python to Argo.

hep/proposals/0001-decorators.md
samj1912: It waits for the workflow to complete execution. In this case the workflow is running on Argo but there can be a local mode as well that is eventually planned once all of the other templates are callable.

flaviuvadan (commented):
Would love to see 2 more examples:
- fanout 
- some plan to address the potential typehint issues introduced by multiple decorators

elliotgunton:
@flaviuvadan - I moved fan-out discussion to "unresolved questions" after discussing on slack - it will still be possible to use the standard `"{{item}}"` syntax to pass inputs from `with_items` into the function, but it will harder to implement local running for fan-outs/ins (without a new feature like a `hio.Item` class).

For the concerns around decorators, for normal decorators it's a solved problem in Python (https://mypy.readthedocs.io/en/stable/generics.html#declaring-decorators) - see POC using multiple decorators:

image

We'll need to add the step/task kwargs, but this was recently fixed in the existing code in #1018 so I have a good idea of how to handle it, plus the mypy page linked above shows an example where you can add an argument to the function args via the decorator which might help us figure it out:

```py
from typing import Callable, TypeVar
from typing_extensions import Concatenate, ParamSpec

P = ParamSpec('P')
T = TypeVar('T')

def printing_decorator(func: Callable[P, T]) -> Callable[Concatenate[str, P], T]:
    def wrapper(msg: str, /, *args: P.args, **kwds: P.kwargs) -> T:
        print("Calling", func, "with", msg)
        return func(*args, **kwds)
    return wrapper

@printing_decorator
def add_forty_two(value: int) -> int:
    return value + 42

a = add_forty_two('three', 3)
```

samj1912:
Merging early as the HEP is unanimously approved by all maintainers.

PR #1041, opened 2024-04-25
New script decorator

### LINKED ISSUE #1042: HEP0001 - new script decorator
See https://github.com/argoproj-labs/hera/blob/873c5b9b441644b2f5d28025eee085906da6156f/hep/proposals/0001-decorators.md

elliotgunton:
**Pull Request Checklist**
- [x] Fixes #1042 
- [x] Tests added

**Description of PR**

* Added basic script example from HEP0001, with a second script template to demonstrate setting the entrypoint using the alternative template name

### DIFF
--- a/examples/workflows/experimental/new_decorators_basic_script.py
+++ b/examples/workflows/experimental/new_decorators_basic_script.py
@@ -11,9 +11,17 @@ class MyInput(RunnerInput):
     user: str
 
 
-@w.set_entrypoint()
 @w.script()
 def hello_world(my_input: MyInput) -> RunnerOutput:
     output = RunnerOutput()
     output.result = f"Hello Hera User: {my_input.user}!"
+    return output
+
+
+# Pass script kwargs (including an alternative public template name) in the decorator
+@w.set_entrypoint()
+@w.script(name="goodbye-world", labels={"my-label": "my-value"})
+def goodbye(my_input: MyInput) -> RunnerOutput:
+    output = RunnerOutput()
+    output.result = f"Goodbye Hera User: {my_input.user}!"
     return output
--- a/examples/workflows/experimental/new_decorators_basic_script.py
+++ b/examples/workflows/experimental/new_decorators_basic_script.py
@@ -19,7 +19,7 @@ def hello_world(my_input: MyInput) -> RunnerOutput:
 
 
 # Pass script kwargs (including an alternative public template name) in the decorator
-@w.set_entrypoint()
+@w.set_entrypoint
 @w.script(name="goodbye-world", labels={"my-label": "my-value"})
 def goodbye(my_input: MyInput) -> RunnerOutput:
     output = RunnerOutput()
--- a/examples/workflows/experimental/new_decorators_basic_script.py
+++ b/examples/workflows/experimental/new_decorators_basic_script.py
@@ -1,0 +1,19 @@
+from hera.shared import global_config
+from hera.workflows import RunnerInput, RunnerOutput, WorkflowTemplate
+
+global_config.experimental_features["script_annotations"] = True
+global_config.experimental_features["script_pydantic_io"] = True
+
+w = WorkflowTemplate(name="my-template")
+
+
+class MyInput(RunnerInput):
+    user: str
+
+
+@w.set_entrypoint()
+@w.script()
+def hello_world(my_input: MyInput) -> RunnerOutput:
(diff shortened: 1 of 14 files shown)


samj1912:
@elliotgunton can we rearrange the examples in a separate PR?

examples/workflows/experimental/new_decorators_basic_script.py
```diff
@@ -0,0 +1,27 @@
...
+
+@w.script()
+def hello_world(my_input: MyInput) -> RunnerOutput:
+    output = RunnerOutput()
+    output.result = f"Hello Hera User: {my_input.user}!"
+    return output
+
+
+# Pass script kwargs (including an alternative public template name) in the decorator
+@w.set_entrypoint()
```
samj1912, line 22: Can we change this to `w.set_entrypoint` instead?

Alternatively should we just have an entrypoint keyword in w.script?

samj1912, line 22:
```suggestion
@w.set_entrypoint
```

examples/workflows/experimental/new_decorators_basic_script.py
```diff
@@ -0,0 +1,27 @@
...
+@w.script(name="goodbye-world", labels={"my-label": "my-value"})
```
samj1912, line 23:
```suggestion
@w.script(name="goodbye-world", labels={"my-label": "my-value"}, entrypoint=True)
```

samj1912: We can either do this or the other one.

examples/workflows/experimental/new_decorators_basic_script.py
elliotgunton: I had some problem using a non-bracketed decorator before, seems fine now though 🤷

examples/workflows/experimental/new_decorators_basic_script.py
elliotgunton: Just wanted to comment that even though we're going with the `set_entrypoint` fix, I don't think adding this in the `script` decorator makes sense as it's a property of the workflow, not an individual template. Plus it would mess up the `kwargs` type hinting unless we added it to a mixin (except the previous point means doing that doesn't make sense).

samj1912:
@elliotgunton re:set entrypoint erroring out, I think we should make it idempotent instead. ie if the same entrypoint is set, let it be (this might happen if the module is imported multiple times which is quite common). In case of a different entrypoint, maybe a warning rather than an error would be better and have the last one be the default one.

src/hera/workflows/workflow.py
```diff
@@ -476,5 +481,16 @@ def get_workflow_link(self) -> str:
...
         return self.workflows_service.get_workflow_link(self.name)
 
+    def set_entrypoint(self, func: Callable[P, T]) -> Callable[P, T]:
+        """Decorator function to set entrypoint."""
+        if not hasattr(func, "template_name"):
+            raise SyntaxError("`set_entrypoint` decorator must be above template decorator")
+
+        if self.entrypoint is not None:
+            logging.warning("entrypoint is being reassigned, the last assigned entrypoint will be used")
+
```
samj1912, line 491:
```suggestion
        if self.entrypoint is not None:
            if self.entrypoint == func.template_name:
                return func
            logging.warning("entrypoint is being reassigned from {self.entrypoint} to {func.template_name}")

```

src/hera/workflows/_meta_mixins.py
```diff
@@ -404,3 +431,113 @@ def _get_deduped_params_from_items(self, parameter_names: Set[str], items: List[
...
+            else:
+                source = func
+
+            if "name" in script_kwargs:
+                # take the client-provided `name` if it is submitted, pop the name for otherwise there will be two
+                # kwargs called `name`
+                name = script_kwargs.pop("name")
+            else:
+                # otherwise populate the `name` from the function name
+                name = source.__name__.replace("_", "-")
```
samj1912, line 515: Can we avoid this implicit replacement?

elliotgunton: Nope

```
workflow failed validation: spec.templates[0].name: 'whalesay_with_underscore' is invalid: name must consist of alpha-numeric characters or '-', and must start with an alpha-numeric character (e.g. My-name1-2, 123-NAME)"
zzz

PR #1053, opened 2024-05-02
Support user specified Bearer tokens

### LINKED ISSUE #1016: Hera's token handling feels a bit awkward

- [x] No

- [ ] exporting to YAML
- [x] submitting to Argo
- [ ] other: 

<h2>Bug report</h2>

**Describe the bug**

When submitting a flow, the current way the token is provided feels a bit awkward based on the way the token 'comes' from the Argo UI.

For example, on version 3.5.5 the ArgoUI will provide a token in the shape of `ARGO_TOKEN="Bearer v2:<token>"`, which we are currently setting on the argo side by setting it as a parameter in `ArgoWorkflowsServerConfig`

```Python
ArgoWorkflowServerConfig(
        host=host, namespace=namespace, service_account_name=service_account_name, token=token
)
```

This is in turn used downstream by the `create_workflow` 
https://github.com/argoproj-labs/hera/blob/main/src/hera/workflows/service.py#L842 which contains on the Authorisation header a `Bearer ` prefix hardcoded.

The user needs to know about this quirk to understand that the value that needs to be set for the token is actually `v2:<token>` which seems unexpected.

_Error log if applicable_:

Receiving a 401 error code when submitting the workflow

**To Reproduce**
```py
import os

with Workflow(name="my-workflow") as wf:
    pass

wf.workflows_service = ArgoWorkflowServerConfig(token=os.getenv("ARGO_TOKEN"))
wf.create()
```

**Expected behavior**
The above or equivalent should submit a workflow

**Environment**
- Hera Version: 5.14.2
- Python Version: 3.10.X
- Argo Version: 3.5.5

**Additional context**

flaviuvadan:
**Pull Request Checklist**
- [x] Fixes #1016
- [x] Tests added

See #1016 for a description provided by users. This PR adds support for user specified Bearer tokens while providing a backwards compatible experience.

### DIFF
--- a/src/hera/events/service.py
+++ b/src/hera/events/service.py
@@ -68,7 +68,6 @@ class EventsService:
         else:
             self.token = token
 
-        self.token = token or global_config.token
         self.namespace = namespace or global_config.namespace
 
     def list_event_sources(
--- a/src/hera/events/service.py
+++ b/src/hera/events/service.py
@@ -57,16 +57,19 @@ class EventsService:
         # Note that the `Bearer` token can be specified through the global configuration as well. In order to deliver
         # a fix on Hera V5 without introducing breaking changes, we have to support both
         global_config_token = global_config.token  # call only once because it can be a user specified function!
-        if token and token.lower().startswith("bearer"):
-            self.token: Optional[str] = token
-        elif token and not token.lower().startswith("bearer"):
-            self.token = "Bearer " + token
-        elif global_config_token and global_config_token.lower().startswith("bearer"):
-            self.token = global_config_token
-        elif global_config_token and not global_config_token.lower().startswith("bearer"):
-            self.token = "Bearer " + global_config_token
+
+        def format_token(t):
+            parts = t.split()
+            if len(parts) == 1:
+                return "Bearer " + t
+            return t
+
+        if token:
+            self.token: Optional[str] = format_token(token)
+        elif global_config_token:
+            self.token = format_token(global_config_token)
         else:
-            self.token = token
+            self.token = None
 
         self.namespace = namespace or global_config.namespace
 
--- a/src/hera/events/service.py
+++ b/src/hera/events/service.py
@@ -59,7 +59,7 @@ class EventsService:
         global_config_token = global_config.token  # call only once because it can be a user specified function!
 
         def format_token(t):
-            parts = t.split()
+            parts = t.strip().split()
             if len(parts) == 1:
                 return "Bearer " + t
             return t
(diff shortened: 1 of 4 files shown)


src/hera/events/service.py
```diff
@@ -50,7 +50,24 @@ def __init__(
...
         self.verify_ssl = verify_ssl if verify_ssl is not None else global_config.verify_ssl
-        self.token = token or global_config.token
+
+        # some users reported in https://github.com/argoproj-labs/hera/issues/1016 that it can be a bit awkward for
+        # Hera to assume a `Bearer` prefix on behalf of users. Some might pass it and some might not. Therefore, Hera
+        # only prefixes the token with `Bearer ` if it's not already specified and lets the uses specify it otherwise.
+        # Note that the `Bearer` token can be specified through the global configuration as well. In order to deliver
+        # a fix on Hera V5 without introducing breaking changes, we have to support both
+        global_config_token = global_config.token  # call only once because it can be a user specified function!
+        if token and token.lower().startswith("bearer"):
```
samj1912, line 60: I think a better logic would be to split this header on whitespace and if it has only one part, prefix it with Bearer otherwise allow users to specify the scheme.

flaviuvadan: Can you say a bit more about why? I know this is relatively trivial but I chose to be explicit about the "bearer" part so it's clear what Hera's looking for. Doing it on length seems a bit more implicit. While it can surely be documented it's not as "self describing". Open to switching it, of course, just want to know more!

elliotgunton: I was doing a bit of research (new to me as well) - https://stackoverflow.com/a/77426797 - basically `Bearer` is an authentication scheme but it isn't the _only_ option

flaviuvadan: Ah, got it

src/hera/events/service.py
```diff
@@ -50,7 +50,27 @@ def __init__(
...
+
+        # some users reported in https://github.com/argoproj-labs/hera/issues/1016 that it can be a bit awkward for
+        # Hera to assume a `Bearer` prefix on behalf of users. Some might pass it and some might not. Therefore, Hera
+        # only prefixes the token with `Bearer ` if it's not already specified and lets the uses specify it otherwise.
+        # Note that the `Bearer` token can be specified through the global configuration as well. In order to deliver
+        # a fix on Hera V5 without introducing breaking changes, we have to support both
+        global_config_token = global_config.token  # call only once because it can be a user specified function!
+
+        def format_token(t):
+            parts = t.split()
```
samj1912, line 62:
```suggestion
            parts = t.strip().split()
```

flaviuvadan:
CC @ljyanesm @Adam-D-Lewis

samj1912 (approved):

PR #1054, opened 2024-05-04
add validation for duplicate template and node names

crflynn:
**Pull Request Checklist**
- [ ] Fixes #
- [x] Tests added

**Description of PR**
Currently, hera lacks validation for duplicate template names, causing templates with duplicate names to be missing when rendering to yaml. Hera also lacks validation for duplicate node names, which results in rendered yaml that is invalid when submitted to argo-workflows.

This PR adds validation for both situations, preventing the user from rendering incorrect or invalid yaml when Workflows contain multiple templates or nodes with the same name by raising a TemplateNameConflict or NodeNameConflict, respectively.

Note that the order of operations has been adjusted in `_HeraContext.add_sub_node`. This change was required to continue to support workflows with recursive references.


elliotgunton:
Thanks for the contribution @crflynn! Although I'm not entirely opposed to the change, I am hesitant about starting to recreate all of Argo's [validation logic](https://github.com/argoproj/argo-workflows/blob/a4fc31887041e80f7f8f51836bbf692a6ed83c98/workflow/validate/validate.go).

Trying to understand your use-case/motivation for the change --
* Is your use-case mainly coming from the developer experience? Or are you hitting problems with gitops flows where the YAML is already rendered and intended to deploy?
* Are you doing things in the Workflow definition that are only possible in Hera, such as conditionally adding templates/tasks to the Workflow definition?
* Are you unable to run an `argo lint` on the cluster (via hera or CLI) before submitting? (During tests?)

I think this change is small and helpful enough that we'd accept it, but it would be good to understand the motivation behind the change and to figure out in advance how far we want to go with matching Argo's on-cluster-validation (cc @samj1912 @flaviuvadan).

crflynn:
We found that the most common issue of users was name conflicts like these, which wouldn’t manifest until execution time when they would error on the controller, so we added some validation at build time to reduce iterations. I figured it would be helpful upstream but completely understand the reasoning behind not wanting to go down this path with hera.

---

For context:

We have a downstream library that wraps hera and provides more functionality. The main features are:
* some basic validation including name conflicts like this
* rendering helm templates in addition to plain yaml, i.e. escaping argo templating for helm compatibility
* providing a local runner
* providing a CronWorkflowTemplate abstraction that builds a WorkflowTemplate and a CronWorkflow that references it. This allows us to submit these workflows with custom arguments via the UI (can’t do this with CronWorkflow in the UI afaik)

among other features for local development.

We build and deploy our workflows using helm charts. This allows us to template things like the schedule, env vars, or entire WorkflowTemplates by environment. It also gives us the benefits of helm like helm rollbacks and shipping the workflows in a declarative way together with other kubernetes resources, e.g. ConfigMaps, Secrets, RBAC, and associated applications.

I gave `argo lint` a try and with helm it would look something like this where we helm template first and then lint the output.

```bash
helm template . --output-dir output
argo lint output --output simple --kinds=cronworkflows
argo lint output --output simple --kinds=workflowtemplates
```

We are only using CronWorkflows and WorkflowTemplates.

It seems to lint our CronWorkflows or objects that reference other templates we would have to expose the API to our CI, which we don’t do, so this just results in authorization errors.

Linting the WorkflowTemplates results in `json: unknown field "hooks"` on every template, despite the hooks functioning just fine during operation. (might be a bug on the cli?)

Perhaps what we are doing is a bit unconventional, I'm not sure, but it has been working fairly well despite the lack of validation in hera. To a degree, the lack of validation also helps us, since sometimes we template things like the cron schedule with helm, whereas strict cron string validation would actually break how we are using it in this particular case.

elliotgunton:
Thanks for the detailed response! Really helpful to understand how Hera is used and built upon 😄 I think I'm happy to approve the intent of the PR, will take a closer look at the code now (i.e. the `add_sub_node` changes)! 🚀

src/hera/workflows/_context.py
```diff
@@ -76,27 +76,29 @@ def add_sub_node(self, node: Union[SubNodeMixin, TTemplate]) -> None:
...
-            # if the above fails, it means the user invoked a decorated function e.g. `@script`. Hence,
-            # the object needs to be added as a template to the piece of context at [-1]. This will be the case for
-            # DAGs and Steps
-            pieces[-1]._add_sub(node.template)  # type: ignore
-
-        # when the above does not raise an exception, it means the user invoked a decorated function e.g. `@script`
+        # for particular types, the user invoked a decorated function e.g. `@script`
         # inside a proper context. Here, we add the object to the overall workflow context, directly as a template,
         # in case it is not found (based on the name). This helps users save on the number of templates that are
         # added when using an object that is a `Script`
```
elliotgunton, line 82: We can simplify/correct the comment
```suggestion
        # When the user invokes a decorated function e.g. `@script inside a sub-context (dag/steps),
        # we also add the step/task's template to the overall workflow context, if it is not already added.
```

src/hera/workflows/_context.py
```diff
@@ -76,27 +76,29 @@ def add_sub_node(self, node: Union[SubNodeMixin, TTemplate]) -> None:
...
                 pieces[0]._add_sub(node.template)
 
+        try:
+            # here, we are trying to add a node to the last piece of context in the hopes that it is a subbable
+            pieces[-1]._add_sub(node)
+        except InvalidType:
+            # if the above fails, it means the user invoked a decorated function e.g. `@script`. Hence,
+            # the object needs to be added as a template to the piece of context at [-1]. This will be the case for
+            # DAGs and Steps
+            pieces[-1]._add_sub(node.template)  # type: ignore
```
elliotgunton, line 101: The old comment here didn't really make sense to me, and checking out the code it turns out it's never hit in our tests. cc @flaviuvadan I think the backend logic changed such that we don't need the try/except (i.e. callers of `_add_sub_node` always pass in a valid node type as it's called in `SubNodeMixin`, and I don't see a way to pass an invalid type). Tried removing it and tests all pass.

```suggestion
        pieces[-1]._add_sub(node)
```

elliotgunton (commented):
Hey @crflynn! Sorry for the delay. The PR looks good overall, some nitpicks and things slightly unrelated to your changes, the only real requests are to add a test and remove another. As it's been a few weeks I'd be happy to wrap up the PR for the fixes and resolve the changes to get it merged (I'll do this tomorrow if no reply or you confirm your preference either way).

crflynn:
@elliotgunton thanks for the review. The changes all make sense to me. I probably won't have time to make them myself until next week, so feel free to resolve them if you have the time.

elliotgunton (approved):

PR #1059, opened 2024-05-07
New dag decorator

### LINKED ISSUE #1043: HEP0001 - new DAG/Steps decorator
See https://github.com/argoproj-labs/hera/blob/main/hep/proposals/0001-decorators.md#dag-template

elliotgunton:
**Pull Request Checklist**
- [x] Fixes #1043 
- [x] Tests added

**Description of PR**
Adds the new dag decorator from HEP0001.

* Within the decorator function call we perform a static inspection of
inputs and outputs, add a `DAG` object to the Workflow, and finally
do a "run" of the DAG function code using a templated input to collect the tasks
* Tasks that pass arguments simply take the templated input strings by using pydantic's `construct` method to skip validation, and then pass the strings around
* The dependencies between tasks are inferred automatically via __getattr__ calls
on the Task object
* Add varname as a dependency to be able to get the python variable name to use
as the task name (in the YAML)
* Make the _add_type_hints function work for *any* pydantic class

### DIFF
--- a/docs/examples/workflows/experimental/new_dag_decorator_params.md
+++ b/docs/examples/workflows/experimental/new_dag_decorator_params.md
@@ -16,6 +16,7 @@
 
     global_config.experimental_features["script_annotations"] = True
     global_config.experimental_features["script_pydantic_io"] = True
+    global_config.experimental_features["decorator_syntax"] = True
 
 
     w = Workflow(generate_name="my-workflow-")
@@ -135,6 +136,8 @@
           - default: ''
             name: word_a
           - name: word_b
+          - default: '{"reverse": false}'
+            name: concat_config
         name: concat
         script:
           args:
@@ -163,6 +166,8 @@
                 value: '{{inputs.parameters.value_a}}'
               - name: word_b
                 value: '{{tasks.setup_task.outputs.parameters.environment_parameter}}{{tasks.setup_task.outputs.parameters.dummy-param}}'
+              - name: concat_config
+                value: '{"reverse": false}'
             depends: setup_task
             name: task_a
             template: concat
@@ -172,6 +177,8 @@
                 value: '{{inputs.parameters.value_b}}'
               - name: word_b
                 value: '{{tasks.setup_task.outputs.result}}'
+              - name: concat_config
+                value: '{"reverse": false}'
             depends: setup_task
             name: task_b
             template: concat
@@ -181,6 +188,8 @@
                 value: '{{tasks.task_a.outputs.result}}'
               - name: word_b
                 value: '{{tasks.task_b.outputs.result}}'
+              - name: concat_config
+                value: '{"reverse": false}'
             depends: task_a && task_b
             name: final_task
             template: concat
@@ -191,6 +200,8 @@
           - name: value_b
           - default: '42'
             name: an_int_value
+          - default: '{"param_1": "Hello", "param_2": "world"}'
+            name: a_basemodel
         name: worker
         outputs:
           parameters:
--- a/docs/examples/workflows/experimental/new_dag_decorator_params.md
+++ b/docs/examples/workflows/experimental/new_dag_decorator_params.md
@@ -1,0 +1,173 @@
+# New Dag Decorator Params
+
+
+
+
+
+
+=== "Hera"
+
+    ```python linenums="1"
+    from typing_extensions import Annotated
+
+    from hera.shared import global_config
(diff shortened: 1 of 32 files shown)


elliotgunton:
I want to add an inner-dag example with associated code fix in `def dag_call_wrapper` and then PR will be ready for review. The current code shouldn't change though if you want to take a look @samj1912 @flaviuvadan

src/hera/workflows/io/_io_mixins.py
```diff
@@ -165,7 +202,10 @@ def _get_output(cls, field_name: str) -> Union[Artifact, Parameter]:
         return Parameter(name=field_name, default=default)  # type: ignore
 
     def _get_as_output(self) -> List[Union[Artifact, Parameter]]:
```
elliotgunton, line 204: Note @samj1912 this function is new but was accidentally added to the IO refactor (but was unused in that commit). TBD how/if we should change from "value from" `parameter` to `expression`.

src/hera/workflows/io/_io_mixins.py
```diff
@@ -37,6 +38,25 @@
...
+            return instance
+        else:
+            return super(InputMixin, cls).__new__(cls)
+
+    def __init__(self, /, **kwargs):
+        if _context.declaring:
+            # Return in order to skip validation of `construct`ed instance
+            return
+
+        return super().__init__(**kwargs)
```
samj1912, line 58: We should not have a return from init

src/hera/workflows/io/_io_mixins.py
elliotgunton: Confirmed that valueFrom->parameter and valueFrom->expression are functionally equivalent (and both work for the special `.result` output) using the following yaml:

```yaml
      outputs:
        parameters:
          - name: value
            valueFrom:
              parameter: '{{tasks.whalesay.outputs.result}}'
```

```yaml
      outputs:
        parameters:
          - name: value
            valueFrom:
              expression: 'tasks.whalesay.outputs.result'
```

Where they differ would be for garbage references and how the error show up in the UI, where the valueFrom->parameter makes a bit more sense:
```yaml
      outputs:
        parameters:
          - name: value
            valueFrom:
              parameter: '{{garbage-parameter}}'
```
image

```yaml
      outputs:
        parameters:
          - name: value
            valueFrom:
              expression: "garbage-expression"
```
image

Neither actually resolves the reference when submitting the workflow, and only fails when running 😐 This shouldn't be a problem because we're constructing the `valueFrom`.

pyproject.toml
```diff
@@ -30,6 +30,7 @@ pyyaml = { version = ">=6.0", optional = true }
 requests = "*"
 pydantic = { extras = ["email"], version = ">=1.7,<3.0" }
 cappa = {version = ">=0.14.3,<0.20.0", optional = true}
+varname = "^0.13.1"
```
samj1912, line 33: @elliotgunton this should be an optional dependency

elliotgunton: I would agree if it's just while the new decorators are "experimental"? After that, I think we want it to be the golden path so should have minimal friction to use it (and so come packaged in)

elliotgunton: Any preference on what the `extras` group should be called in `pyproject.toml`? Going with `experimental` for now which will encompass current/future optional dependencies for any experimental features (aside from the cli which specifically pulls in cappa)

```
varname = {version = "^0.13.1", optional = true}

[tool.poetry.extras]
yaml = ["PyYAML"]
cli = ["cappa", "PyYAML"]
experimental = ["varname"]
```

docs/examples/workflows/experimental/new_dag_decorator_params.md
```diff
@@ -0,0 +1,187 @@
...
+
+
+    class WorkerOutput(Output):
+        value: str
+
+
+    @w.set_entrypoint
+    @w.dag()
+    def worker(worker_input: WorkerInput) -> WorkerOutput:
+        setup_task = setup()
```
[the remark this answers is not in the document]
elliotgunton: Thanks - this uncovered a bug in `_get_as_arguments`, fixed in latest push, where we need to serialize the value from the model dump:
```py
            # The value may be a static value (of any time) if it has a default value, so we need to serialize it
            # If it is a templated string, it will be unaffected as `"{{mystr}}" == serialize("{{mystr}}")``
            templated_value = serialize(self_dict[field])
```

PR #1068, opened 2024-05-15
New container decorator

### LINKED ISSUE #1065: HEP0001 - new container decorator
See https://github.com/argoproj-labs/hera/blob/main/proposals/heps/0001-decorators.md#container-template

elliotgunton:
**Pull Request Checklist**
- [x] Fixes #1065 
- [x] Tests added

**Description of PR**
Adds the container decorator, and the special `path` function on Outputs


examples/workflows/experimental/new_container_decorator.py
```diff
@@ -0,0 +1,49 @@
+from typing import Optional
+
+from typing_extensions import Annotated
+
+from hera.shared import global_config
+from hera.workflows import Container, Input, Output, Parameter, WorkflowTemplate
+
+global_config.experimental_features["script_annotations"] = True
+global_config.experimental_features["script_pydantic_io"] = True
```
samj1912, line 9: Can we consolidate all of the flags?

elliotgunton: Yeah, was thinking to do a small PR on top as it'll probably touch other parts of the codebase

elliotgunton: Small one here https://github.com/argoproj-labs/hera/pull/1076

examples/workflows/experimental/new_container_decorator_auto_path.py
```diff
@@ -0,0 +1,33 @@
...
+    # We will let Hera figure out the output path
+    container_greeting: Annotated[str, Parameter(name="container-greeting")]
+
+
+@w.set_entrypoint
+@w.container(command=["sh", "-c"])
+def advanced_hello_world(my_input: MyInput, template: Optional[Container] = None) -> MyOutput:
+    output: MyOutput = MyOutput(container_greeting=f"Hello {my_input.user}")
+
+    if template:
```
samj1912, line 28: Not sure if I like this pattern of doing if template.

elliotgunton: I don't love it but it was the best I could find for linter/developer sanity - magically adding the arg in the decorator is a bit too opaque even for devs IMO

samj1912:
@elliotgunton how about instead of using template, we can use the function itself?


```py
def my_container(my_args:...):
    my_container.<attribute>
```

elliotgunton:
@samj1912 is there a sensible way to add `container_call_wrapper.template = container_template` as a type-hintable attribute to the function? (From some [searching](https://stackoverflow.com/questions/75085009/how-to-type-hint-a-function-added-to-class-by-class-decorator-in-python) it seems not)

image

Otherwise I don't think it's a nice user experience. For the `template: Optional[Container] = None` argument to the function, in terms of user experience it at least kind of acts as an explicit opt-in for advanced users, who would probably prefer nice code completion over nice syntax. (If we _can_ add type hints to the function attribute then I'm 100% for it!)

Another idea is some global value that we can set before calling the user function? 

```py
from hera.workflows import current_template  # no idea on naming 😅

@w.container(command=["sh", "-c"])
def advanced_hello_world(my_input: MyInput) -> MyOutput:
    output: MyOutput = MyOutput(container_greeting=f"Hello {my_input.user}")
    current_template.args = [f"echo Hello {my_input.user} | tee {output.path(output.container_greeting)}"]
    return output
```

flaviuvadan (commented):
impl wise this looks good to me. We already chatted about the user experience in one of the maintainer syncs. Anything else to review @elliotgunton? Happy to approve bc it lgtm!

samj1912:
I would prefer to go back to the drawing board for this specific feature as I'm not sure of the UX it promotes. Happy to release the other parts that are already merged. We barely use container templates in Hera anyway compared to scripts.

elliotgunton:
@samj1912 I'm not sure there is a better balance of UX for the "advanced container" use case from the proposal, as the proposal itself broke the paradigm of config-as-code (contained within type signatures etc), instead running the function to create the container config/spec. I can remove the code for the advanced use case if the basic one is fine?

samj1912:
Sure, let's start with that

src/hera/workflows/io/_io_mixins.py
```diff
@@ -175,7 +206,31 @@ def __init__(self, /, **kwargs):
         super().__init__(**kwargs)
 
     @classmethod
-    def _get_outputs(cls) -> List[Union[Artifact, Parameter]]:
+    def path(cls, attribute: Any) -> Path:
```
samj1912, line 209: This is not used anywhere right? Do we need this?

elliotgunton: Ah yep, will move to #1098

samj1912 (approved):

samj1912:
@elliotgunton if it's simple, can we also extend this to other template types?

elliotgunton:
@samj1912 as in stubbing the function for inputs/outputs? Yes that should be okay. I'll make a new issue for it

samj1912:
Yes, given a template being able to set its properties in the decorator and the i/o in the signature.

PR #1087, opened 2024-06-03
Add client certificates for workflow service and global. Fixes #1079

### LINKED ISSUE #1079: Support user-provided certificates and keys
The `requests` package that Hera relies on to send requests to the Argo server supports specifying [client-provided certificates and keys](https://requests.readthedocs.io/en/latest/user/advanced/#client-side-certificates). This is very useful for specific user setups, such as having a proxy that intercepts a request and performs authentication not on the bearer token but on the certificate. 

An API on the global config similar to the [token API](https://github.com/argoproj-labs/hera/blob/main/src/hera/shared/_global_config.py#L98) that allows either a string path to the certificate / key or a function to load that from a custom location (useful for when the cert is loaded dynamically from another location that needs auth on the user side). Maybe `cert` would be enough, and all it does is load a [tuple that's passed to requests like this](https://requests.readthedocs.io/en/latest/user/advanced/#client-side-certificates)

Could explore ways to intercept a requests package request but that feels like extra burden on the user and conflicts with Hera's ambition to support "configuring a platform" via hooks 

**Additional context**
None atm

mrendi29:
**Pull Request Checklist**
- [x] Fixes #1079 
- [x] Tests added

**Description of PR**
This PR adds support to add user supported certificates and keys.


src/hera/shared/_global_config.py
```diff
@@ -40,7 +40,7 @@ class _GlobalConfig:
     # are processed upon accessing. The rest, which use primitive types, such as `str`, can remain public
     _token: Optional[Union[str, TokenGenerator, Callable[[], Optional[str]]]] = None
     """an optional authentication token used by Hera in communicating with the Argo server"""
-
+    _client_certs: Optional[Tuple[str, str]] = None
```
flaviuvadan, line 43: What do you think about making this a `Path` object? That way Hera can check that the cert/key files do exist

src/hera/shared/_global_config.py
```diff
@@ -99,6 +99,16 @@ def token(self, t: Union[Optional[str], TokenGenerator, Callable[[], Optional[st
         """Sets the Argo Workflows token at a global level so services can use it."""
         self._token = t
 
+    @property
+    def client_certs(self) -> Optional[Tuple[str, str]]:
+        return self._client_certs
+
+    @client_certs.setter
+    def client_certs(self, certs: Tuple[str, str]) -> None:
```
flaviuvadan, line 107: Would it be beneficial for this to have the signature
```suggestion
    def client_certs(self, certs: Tuple[str, str] | Callable(None, Tuple[str, str])) -> None:
```

I am thinking whether it's possible for users to "fetch" the cert/key files using some client logic, store those on local, then return those paths as tuples?

src/hera/workflows/service.py
```diff
@@ -130,6 +131,7 @@ def list_archived_workflows(
             headers={"Authorization": self.token},
             data=None,
             verify=self.verify_ssl,
+            cert=self.client_certs,
```
flaviuvadan, line 134: It seems like this was modified manually. This service is actually generated by [this script](https://github.com/argoproj-labs/hera/blob/main/scripts/service.py). Concretely, the `cert` would be added [here](https://github.com/argoproj-labs/hera/blob/main/scripts/service.py#L184) and the `self.client_certs` line would go [here](https://github.com/argoproj-labs/hera/blob/main/scripts/service.py#L441)

flaviuvadan: Lastly, if you call `make services` ([reference](https://github.com/argoproj-labs/hera/blob/main/Makefile#L109)) after contributing the above it will generate the service for you!

src/hera/shared/_global_config.py
mrendi29: You are right, I must have overlooked this at first. Thank you!

src/hera/workflows/service.py
mrendi29: Thats a really good trick, I was not aware of that!

src/hera/shared/_global_config.py
```diff
@@ -99,6 +100,16 @@ def token(self, t: Union[Optional[str], TokenGenerator, Callable[[], Optional[st
         """Sets the Argo Workflows token at a global level so services can use it."""
         self._token = t
 
+    @property
+    def client_certs(self) -> Optional[Tuple[Path, Path]]:
+        return self._client_certs
+
+    @client_certs.setter
+    def client_certs(self, certs: Tuple[Path, Path] | Callable[[], Tuple[Path, Path]]) -> None:
+        if not all(certs) or not isinstance(certs, tuple):
```
mrendi29, line 109: @flaviuvadan  I am not sure if I should do any additional checking here, what are your thoughts?

mrendi29:
@flaviuvadan  I am also thinking to add some documentation into `hera/docs/walk-through/authentication.md` what do you think?

src/hera/shared/_global_config.py
flaviuvadan: Not sure we can perform any checks here, actually, since we have a tuple or a callable. I'd make the `property` `client_certs` do something like 
```
if certs is callable
    return self._certs()
return self._certs
```
and not do any checking. Maybe expand the type signature too to `Path | str` and that's it

flaviuvadan (commented):
Thanks for applying the first round of feedback @mrendi29! This looks great

flaviuvadan:
> @flaviuvadan I am also thinking to add some documentation into `hera/docs/walk-through/authentication.md` what do you think?

This sounds great!

samj1912:
> @samj1912 thoughts on this? This looks like a great addition to me

Will take a look tomorrow morning LDN time :)

src/hera/shared/_global_config.py
```diff
@@ -41,6 +42,11 @@ class _GlobalConfig:
     _token: Optional[Union[str, TokenGenerator, Callable[[], Optional[str]]]] = None
     """an optional authentication token used by Hera in communicating with the Argo server"""
 
+    _client_certs: Optional[
```
elliotgunton, line 45: What do people think of having a custom type (similar to custom generator) to capture all the variations? Or is there one already we can use? I know we're passing through to `requests.get` but I'm not super familiar with SSL/certs standards/patterns, just wondering if the pair is common enough.

mrendi29: @elliotgunton  AFAIK `requests` always expect a path of  (cert,key) `str` pair. So perhaps a custom type might not be useful in this case? I could be wrong though

flaviuvadan (approved):
LGTM! Thank you again @mrendi29 for the contribution! @elliotgunton do you mind offering an approval as well?

samj1912 (approved):

PR #1093, opened 2024-06-07
Support inherited fields for RunnerInput, other niceties

mitrydoug:
## Pull Request Checklist
- [ ] Fixes #
- [x] Tests added

## Summary

Currently, users cannot leverage inheritance when defining `RunnerInput` models that utilize `Annotated` fields. For example:

```
class ResourceIndex(RunnerInput):
    resource_name: Annotated[str, Parameter(name="resource-name")]
    resource_namespace: Annotated[str, Parameter(name="resource-namespace")]

class GetStatusInput(ResourceIndex):
    wait: Annotated[bool, Parameter(name="wait")]
    timeout_s: Annotated[int, Parameter(name="timeout")]
```

This would fail at runtime with a KeyError at [this line](https://github.com/argoproj-labs/hera/blob/main/src/hera/workflows/_runner/script_annotations_util.py#L152), because apparently `cls.__annotations__` are not inherited by child classes.

This PR enables the above inheritance structure by utilizing the existing [get_field_annotations](https://github.com/argoproj-labs/hera/blob/main/src/hera/shared/_pydantic.py#L69) instead of `my_runner_input.__annotations__`.

#### Parameter Default Names

This PR also changes logic so that `Input` field `Parameters` adopt the name of their field if no name is provided. For example, in

```
class MyInput(Input):
    my_int: Annotated[int, Parameter(description="this is my int")]
```

The parameter is converted to `Parameter(name="my_int", description="this is my int")`. This avoids an error from [here](https://github.com/argoproj-labs/hera/blob/main/src/hera/workflows/parameter.py#L30) which is thrown when trying to generate a workflow yaml.


mitrydoug:
applying a simplifying commit and additional test

mitrydoug:
applying one more change, please do not merge

samj1912:
@mitrydoug can you please update the PR description and add some comments to the code describing "why" we are performing certain operations? For eg the copy.

src/hera/workflows/io/_io_mixins.py
```diff
@@ -64,8 +64,11 @@ def _get_parameters(cls, object_override: Optional[Self] = None) -> List[Paramet
 
         for field, field_info in get_fields(cls).items():
             if get_origin(annotations[field]) is Annotated:
-                param = get_args(annotations[field])[1]
+                # Copy so as to not modify the Input fields themselves
+                param = get_args(annotations[field])[1].copy()
```
mitrydoug, line 68: This copy is inspired by [this one](https://github.com/argoproj-labs/hera/blob/main/src/hera/workflows/script.py#L506). However, I can also make sense of it by looking at [this test](https://github.com/argoproj-labs/hera/blob/main/tests/script_annotations/pydantic_io_v1.py#L68). In this case, we would not want to modify the annotations of the `ParamOnlyInput` class, because it may be used by another script template (possibly with different defaults).

mitrydoug:
I removed the logic about finding a unique Parameter/Artifact since it became clear to me that the remainder of the codebase assumes Parameters/Artifacts to be at index 1 of `Annotated`, so I wanted to stay consistent with that.

@samj1912 I have updated my PR description, and commented on the copy.

elliotgunton:
>the remainder of the codebase assumes Parameters/Artifacts to be at index 1

Yeah we haven't heard any complaints from people who want to use more Annotations, so if/when they do we should convert all these parts of the codebase at once

samj1912 (approved):

PR #1098, opened 2024-06-12
Add advanced container use case

elliotgunton:
Creating a draft PR to hold the code and for any discussion


elliotgunton:
We should come back to this in a maintainer's meeting. Not worth prioritising right now

Issue #812, opened 2023-10-21
Input Parameter annotations should only be able to set default via standard Python


elliotgunton:
Kind of a bug/feature request: for input Parameters, a function such as 

```py
def echo(
    an_int: Annotated[int, Parameter(default=1)],
): -> None
```

will not behave the same when unit tested locally vs on Argo. e.g. you would have to double up the defaults whenever you want to test locally, vs letting the default value be used automatically. We should only allow the `default` value to be set via the standard Python syntax, to ensure devs see the same behaviour in local unit tests and on Argo.

Feature request:

```py
# Allowed as-is
def echo(
    an_int: Annotated[int, Parameter(name="alt-name")] = 1,
): -> None


# Should instead raise a validation error:
def echo(
    an_int: Annotated[int, Parameter(name="alt-name", default=1)],
): -> None
```

## Tasks
- [x] Raise validation error for `default` assignment in Annotation
- [ ] Update docs

elliotgunton:
Script annotations have been superseded by Pydantic IO which makes functions much less verbose - is there any reason to keep raw `Annotated` function params around or should we push for a migration to Pydantic IO (once its reliability is more proven)? In which case, `default` should also not be allowed on the `Parameter` itself. Also, note that `script_annotations` hasn't yet graduated experimental features.

PR #1102, opened 2024-06-17
Add warning for removal of default in annotated Parameter args

elliotgunton:
**Pull Request Checklist**
- [x] Groundwork for #812 
- [x] Tests added

**Description of PR**
To complete #812 it would be preferable to give a version's warning of the backwards incompatible change, although I'm happy to make the change in one go.

This PR also fixes a serialization problem when setting the default via a Python function arg.


elliotgunton:
@samj1912 / @flaviuvadan any preference on this change having a notice or just make the change directly? It's an experimental feature after all 👀

samj1912:
I would prefer to give a notice. Let's also provide a couple of minor versions of bandwidth. The way I would prefer to go about removing things is as follows -
- Give warning
- Raise exception which can be turned off by a flag
- Removed feature

src/hera/workflows/script.py
```diff
@@ -513,12 +515,23 @@ class will be used as inputs, rather than the class itself.
...
+                    #         "default cannot be set via the Parameter's default, use a Python default value instead"
+                    #     )
                 if func_param.default != inspect.Parameter.empty:
+                    # TODO: remove this check in 5.18:
                     if new_object.default is not None:
                         raise ValueError(
                             "default cannot be set via both the function parameter default and the Parameter's default"
                         )
-                    new_object.default = str(func_param.default)
+                    new_object.default = serialize(func_param.default)
```
elliotgunton, line 534: This change is technically a bug fix

samj1912 (approved):

PR #1109, opened 2024-06-20
Add ability to subclass from both Input and Output allowing for passthrough IO

### LINKED ISSUE #922: Arguments mapping is very verbose depending on use case
It sucks to write 

```
func1 = my_func1(...)
func2 = my_func2(
    arguments={'a': func1.get_parameter('a').value, 'b': func2.get_parameter('b').value}   # <<<<<<<<
)
```

Something that allows clients to write 
```
func1 = my_func1(...)
func2 = my_func2(
    arguments=func1.outputs  # <<<<<<<<
)
```

Hera can, by default, assume that users want to map all the output parameters from func1 to func2

CC: @ljyanesm

elliotgunton:
**Pull Request Checklist**
- [x] Fixes #922 
- [x] Tests added

**Description of PR**
This PR makes it possible to subclass `Input` and `Output` for the new decorators in HEP0001, and then pass an output from a script template call directly into the next step or task. i.e. instead of

```py
@w.dag()
def my_dag():
    output_task = give_output()
    take_input(
        PassthroughIO(
            my_str=output_task.my_str,
            my_int=output_task.my_int,
            my_artifact=output_task.my_artifact,
        )
    )
```
you can do
```py
@w.dag()
def my_dag():
    output_task = give_output()
    take_input(output_task)
```

Note this functionality is limited to script-decorated functions only for now, as other template types would need an even more involved approach where this code had already become complex enough, as we are reconstructing an `Output` from a step or task object.

### DIFF
--- a/docs/examples/workflows/experimental/new_decorators_passthrough_io.md
+++ b/docs/examples/workflows/experimental/new_decorators_passthrough_io.md
@@ -1,0 +1,124 @@
+# New Decorators Passthrough Io
+
+
+
+
+
+
+=== "Hera"
+
+    ```python linenums="1"
+    from typing_extensions import Annotated
+
+    from hera.shared import global_config
+    from hera.workflows import Artifact, ArtifactLoader, Input, Output, WorkflowTemplate
+
+    global_config.experimental_features["decorator_syntax"] = True
+
+    w = WorkflowTemplate(name="my-template")
+
+
+    class PassthroughIO(Input, Output):
+        my_str: str
+        my_int: int
+        my_artifact: Annotated[str, Artifact(name="my-artifact", loader=ArtifactLoader.json)]
+
+
+    @w.script()
+    def give_output() -> PassthroughIO:
+        return PassthroughIO(my_str="test", my_int=42)
+
+
+    @w.script()
+    def take_input(inputs: PassthroughIO) -> Output:
+        return Output(result=f"Got a string: {inputs.my_str}, got an int: {inputs.my_int}")
+
+
+    @w.dag()
+    def my_dag():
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -29,7 +29,7 @@ codegen: models services examples init-files
 check-codegen: ## Check if the code is up to date
 check-codegen:
 	@$(MAKE) codegen
-	@git diff --exit-code || echo "Code is not up-to-date. Please run 'make codegen'"
+	@git diff --exit-code || echo "Code is not up-to-date. Please run 'make codegen'" && exit 1
```
elliotgunton, line 32: Changed this so we stop missing non-regenerated examples in CI e.g. for #1108 - https://github.com/argoproj-labs/hera/actions/runs/9566065952/job/26370421288 - had the `Code is not up-to-date. Please run 'make codegen'` message but no failure

samj1912 (changes requested):
I would prefer if we stay consistent with the functionality around our new templates. I think this will cause a lot of confusion if it only works for script templates, no that they look exactly the same as other templates. Also how would this work for script template stubs?

elliotgunton:
>I think this will cause a lot of confusion if it only works for script templates

Fair point, I wanted to see if it was possible for script templates first. As for stubs it should work fine as long as they are decorated as script templates - we only need the signature, which is why it's fairly easy to do for script templates as we can access `template.source`. Haven't looked into other template types extensively

Makefile
elliotgunton: Moved to #1110

elliotgunton:
Closing this as stale and it's only a POC which isn't high priority for now

PR #1117, opened 2024-07-01
fix: Update script to explicitly reexport models.

### LINKED ISSUE #1116: Explicitly re-export models
Commonly installed python linter `pyright` complains about implicit reexport

Screenshot 2024-07-01 at 5 06 39 PM

While [hera/workflows/models/__init__.py](https://github.com/argoproj-labs/hera/blob/main/src/hera/workflows/models/__init__.py) imports all the deeper models, it does not add them to an `__all__` block.

Ideally all the models are explicitly reexported, so linters dont complain about it

Perhaps there's a way of configuring pyright to not care about this, but it seems beneficial to fix the default behavior here. particularly in this case where the file seems to be autogenerated and defining the explicit reexport block ought to be relatively straightforward?

---

I'd be down to contribute the fix if interested. It's [this file](https://github.com/argoproj-labs/hera/blob/main/scripts/models.py#L96)?

DanCardin:
**Pull Request Checklist**
- [x] Fixes #1116 
- [ ] Tests added

**Description of PR**
Ensures all models implicitly exported in scripts/models.py through `models/__init__.py` are included in an `__all__` assignment.


scripts/models.py
```diff
@@ -4,13 +4,36 @@
...
 from pathlib import Path
+from typing import List, Optional
 
 import requests
 
 model_types = {"workflows", "events"}
 
 
+@dataclass(frozen=True, order=True)
+class ObjectReference:
```
elliotgunton, line 17: Minor but can we call it an `ImportReference` or something?

DanCardin: For sure!

samj1912 (approved):

PR #1125, opened 2024-07-10
Remove anti-pattern script example

elliotgunton:
Saw this example that we should remove. Using a `Script` directly is an anti-pattern that we should discourage. Renamed the existing script decorator example `dag_diamond_with_callable_decorators` to `dag_diamond_with_callable_script`.


docs/examples/workflows/misc/suspend_input_duration.md
```diff
@@ -58,6 +58,13 @@
             valueFrom:
               supplied: {}
         suspend: {}
+      - inputs:
```
elliotgunton, line 61: Not sure why ordering of templates changed here.

flaviuvadan (approved):

PR #1147, opened 2024-08-09
Update script to handle Optional and Union parameters

### LINKED ISSUE #1012: Support Optional/Union types in script runner

- [x] No

- [ ] exporting to YAML
- [ ] submitting to Argo
- [x] running on Argo with the Hera runner
- [ ] other: 

<h2>Bug report</h2>

**Describe the bug**

Using a function with an optional (string, but I think could be any type) argument in the runner results in the following stack trace:

_Error log if applicable_:
```
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/layers/.../requirements/lib/python3.11/site-packages/hera/workflows/runner.py", line 6, in <module>
    _run()
  File "/layers/.../requirements/lib/python3.11/site-packages/hera/workflows/_runner/util.py", line 254, in _run
    result = _runner(args.entrypoint, kwargs_list)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/layers/.../requirements/lib/python3.11/site-packages/hera/workflows/_runner/util.py", line 225, in _runner
    raise e
  File "/layers/.../requirements/lib/python3.11/site-packages/hera/workflows/_runner/util.py", line 222, in _runner
    output = _save_annotated_return_outputs(function(**kwargs), output_annotations)
                                            ^^^^^^^^^^^^^^^^^^
  File "/layers/.../requirements/lib/python3.11/site-packages/hera/workflows/_runner/util.py", line 58, in inner
    filtered_kwargs = {key: _parse(value, key, f) for key, value in kwargs.items() if _is_kwarg_of(key, f)}
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/layers/.../requirements/lib/python3.11/site-packages/hera/workflows/_runner/util.py", line 58, in <dictcomp>
    filtered_kwargs = {key: _parse(value, key, f) for key, value in kwargs.items() if _is_kwarg_of(key, f)}
                            ^^^^^^^^^^^^^^^^^^^^^
  File "/layers/.../requirements/lib/python3.11/site-packages/hera/workflows/_runner/util.py", line 89, in _parse
    if _is_str_kwarg_of(key, f) or _is_artifact_loaded(key, f) or _is_output_kwarg(key, f):
       ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/layers/.../requirements/lib/python3.11/site-packages/hera/workflows/_runner/util.py", line 142, in _is_str_kwarg_of
    return issubclass(type_, str)
           ^^^^^^^^^^^^^^^^^^^^^^
TypeError: issubclass() arg 1 must be a class
```

**To Reproduce**
```py
from hera.workflows import Workflow, script

@script(constructor="runner", image="my-image")
def optional_str(my_str: Optional[str], another_str: str | None):  # note the old/new ways to specify optional
    print(my_str)
    print(another_str)

with Workflow(name="my-workflow", entrypoint="optional-str") as w:
    optional_str()
```

**Expected behavior**
No error trace.

**Environment**
- Hera Version: 5.14.3
- Python Version: 3.11
- Argo Version: 3.4.16

**Additional context**
We can currently work around this by using an empty string as the default.

```py
@script(constructor="runner", image="my-image")
def optional_str(my_str: str = "", another_str: str = ""):
    print(my_str)
    print(another_str)
```

jeongukjae:
**Pull Request Checklist**
- [x] Fixes #1012 
- [x] Tests added
- [x] Documentation/examples added : I think no needs for document?

**Description of PR**
Currently, `script` decorator cannot handle `Union` or `Optional` parameters because of the error in `issubclass`, so updated it to be able to handle it as expected.


jeongukjae:
Crashed py<3.10. need to fix

samj1912 (changes requested):
Can we implement a slightly more robust union/optional checker? https://github.com/litestar-org/type-lens from the linked issue has some inspiration. I generally don't like adding additional dependencies unless absolutely necessary, so see if we can extract the necessary minimal code. Hera is usually kept lean as it's also run in production.

jeongukjae:
@samj1912 reasonable. thanks! I checked related codes, and here's my suggestion.

It looks like there are too many functions handling types in different ways, and needs to support other types (typevar, other annotations like [required, notrequired](https://docs.python.org/3/library/typing.html#typing.NotRequired)). This can be a big change to fully convert robust way in here. So how about support this one first and refine it in following PRs? I'll create corresponding issues and PRs.

elliotgunton (changes requested):
We also need to change how we generate the YAML for Optional parameters. I checked out this PR and generated for the workflow in the issue:

```py
from typing import Optional, Union

from hera.workflows import Workflow, script

@script(constructor="runner", image="my-image")
def optional_str(my_str: Optional[str], another_str: Union[str, None]):
    print(my_str)
    print(another_str)

with Workflow(name="my-workflow", entrypoint="optional-str") as w:
    optional_str()
```

Which gets YAML:

```yaml
kind: Workflow
metadata:
  name: my-workflow
spec:
  entrypoint: optional-str
  templates:
  - inputs:
      parameters:
      - name: my_str
      - name: another_str
    name: optional-str
    script:
      args:
      - -m
      - hera.workflows.runner
      - -e
      - examples.workflows.scripts.optional_param:optional_str
      command:
      - python
      image: my-image
      source: '{{inputs.parameters}}'
```

Notice how the inputs `my_str` and `another_str` don't have values or defaults. This means Argo will reject the Workflow when you try to submit it.

I confirmed this with the following workflow:

```py

@script()
def dummy_input(dummy):
    print("I didn't use the variable")

with Workflow(generate_name="test-no-value-", entrypoint="d") as w:
    with DAG(name="d") as d:
        dummy_input()
```

Which has YAML:

```yaml
kind: Workflow
metadata:
  generateName: test-no-value-
spec:
  entrypoint: d
  templates:
  - dag:
      tasks:
      - name: dummy-input
        template: dummy-input
    name: d
  - inputs:
      parameters:
      - name: dummy
    name: dummy-input
    script:
      command:
      - python
      image: python:3.8
      source: |-
        import os
        import sys
        sys.path.append(os.getcwd())
        import json
        try: dummy = json.loads(r'''{{inputs.parameters.dummy}}''')
        except: dummy = r'''{{inputs.parameters.dummy}}'''

        print("I didn't use the variable")
```

This gets the submission error:
```
templates.d.tasks.dummy-input templates.dummy-input inputs.parameters.dummy was not supplied
```

jeongukjae:
@elliotgunton thanks! I will check it

jeongukjae:
I think this issue is much bigger than I thought. so it should be better to close this and come back soon :)

jeongukjae:
Hi @elliotgunton,
thank you for the detailed review.

I checked the related code bases, and try to run what you commented.
and I wonder that, do we really need to auto-fill optional parameter as None?

I think this can be considered as expected behaviour.

On the other hand, if user provides None as default value, that should work as expected, like following code:

```py
from typing import Optional

@script()
def dummy_input(dummy: Optional[str] = None):
    print("I didn't use the variable")

with Workflow(
    namespace=...,
    generate_name="test-no-value-",
    entrypoint="d",
    workflows_service=...
) as w:
    with DAG(name="d") as d:
        dummy_input()
```

yaml code:

```yaml
metadata:
  generateName: test-no-value-
spec:
  entrypoint: d
  templates:
  - dag:
      tasks:
      - name: dummy-input
        template: dummy-input
    name: d
  - inputs:
      parameters:
      - default: 'null'
        name: dummy
    name: dummy-input
    script:
      command:
      - python
      image: python:3.8
      source: |-
        import os
        import sys
        sys.path.append(os.getcwd())
        import json
        try: dummy = json.loads(r'''{{inputs.parameters.dummy}}''')
        except: dummy = r'''{{inputs.parameters.dummy}}'''

        print("I didn't use the variable")
```

elliotgunton:
Hey @jeongukjae! Yes on second thought it looks okay - if you're using `Optional` you'll usually add a default of `None`, but here it would be mandated, else your code is not runnable, only "buildable" to yaml - we should instead make it an error to use `Optional` without a default.

I would also ask @samj1912 for a second opinion.

jeongukjae:
> we should instead make it an error to use Optional without a default.

Reasonable. I will contribute subsequent PRs related to this.

jeongukjae:
Recreated PR in https://github.com/argoproj-labs/hera/pull/1160

(I didn't know that I can't reopen PR when I force pushed 😅)

PR #1160, opened 2024-08-20
Update script to handle Optional and Union input parameters

jeongukjae:
**Pull Request Checklist**
- [x] Fixes #1012 
- [x] Tests added
- [x] Documentation/examples added : I think no needs for document?
- [x] [Good commit messages](https://cbea.ms/git-commit/) and/or PR title

**Description of PR**
Currently, `script` decorator cannot handle `Union` or `Optional` input parameters because of the error in `issubclass`, so updated it to be able to handle it as expected.

---

original PR: https://github.com/argoproj-labs/hera/pull/1147

Issue #1167, opened 2024-08-21
Cannot run script with tuple annotation in return type


jeongukjae:
<h2>Pre-bug-report checklist</h2>

**1. This bug can be reproduced using pure Argo YAML**
- [ ] Yes [👉 Please report a bug to the Argo Workflows GitHub 👈](https://github.com/argoproj/argo-workflows/issues/new/choose)
- [x] No

_If yes, it is more likely to be an Argo bug unrelated to Hera. Please double check before submitting an issue to Hera._

**2. This bug occurs in Hera when...**
- [x] exporting to YAML
- [ ] submitting to Argo
- [ ] running on Argo with the Hera runner
- [ ] other: 

<h2>Bug report</h2>

**Describe the bug**
_A clear and concise description of what the bug is:_

_Error log if applicable_:
```
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
src/hera/workflows/_runner/util.py:206: in _runner
    _save_dummy_outputs(output_annotations)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

output_annotations = [(), ()]

    def _save_dummy_outputs(
        output_annotations: List[
            Union[
                Tuple[type, Union[Parameter, Artifact]],
                Union[Type[OutputV1], Type[OutputV2]],
            ]
        ],
    ) -> None:
        """Save dummy values into the outputs specified.
    
        This function is used at runtime by the Hera Runner to create files in the container so that Argo
        does not log confusing error messages that obfuscate the real error, which look like:
        ```
        msg="cannot save parameter /tmp/hera-outputs/parameters/my-parameter"
        argo=true
[...]

PR #1168, opened 2024-08-21
Allow Artifact/Parameter in any position in `Annotated` metadata

jeongukjae:
**Pull Request Checklist**
- [x] ~~Fixes #~~ this PR follows up on #1160
- [x] Tests added
- [x] Documentation/examples added : this is refactoring. no need

**Description of PR**

Currently, we are checking annotated, parameter/artifact, or other types in different ways.
In this PR, I added `hera._utils.type_util` module and unified them into one place.
Followings can be expected:
- If there's another metadata along with Parameter or Artifact, they should be ignored. (e.g. `Annotated[type, metadata1, metadata2, Parameter(...)]`, this can be happened when `Annotated`s are nested)
- `Optional` parameter handling should be more precise which is added in #1160.

Suggestions on function names/module names are welcome.


src/hera/_utils/type_util.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+def can_consume_primitive(annotation: type, target: type) -> bool:
+    """If annotation can be considered as target type or subclass of it, return True."""
+    if is_annotated(annotation):
+        annotation = consume_annotated_type(annotation)
+    casted = may_cast_subscripted_type(annotation)
+    if casted is Union or casted is UnionType:
+        return any(can_consume_primitive(arg, target) for arg in get_args(annotation))
+    return issubclass(casted, target)
```
elliotgunton, line 88: We are not consuming anything here. And there are some confusing variable/function names and repeated execution (no need to check `is_annotated` if we check that inside `unwrap_annotation`). We should follow a combination of `get_origin`/`issubclass` naming precedent:

```suggestion
def origin_type_issubclass(cls: type, type_: type) -> bool:
    """Return True if cls can be considered as a subclass of target."""
    unwrapped_type = unwrap_annotation(annotation)
    origin_type = get_origin_or_builtin(unwrapped_type)
    if origin_type is Union or origin_type is UnionType:
        return any(origin_type_issubclass(arg, type_) for arg in get_args(annotation))
    return issubclass(origin_type, type_)
```

src/hera/_utils/type_util.py
```diff
@@ -0,0 +1,93 @@
...
+
+@overload
+def consume_annotated_metadata(_: Any, __: Type[T]) -> Optional[T]: ...
+
+
+@overload
+def consume_annotated_metadata(_: Any, __: Tuple[Type[T], Type[V]]) -> Optional[Union[T, V]]: ...
+
+
+# FIXME: Currently, mypy cannot guess following type hint properly: https://github.com/python/mypy/issues/17700
```
elliotgunton, line 54: Could be a use case of `TypeVarTuple` but I haven't looked into it much.

elliotgunton (commented):
First pass review mainly on readability/naming 👍

src/hera/_utils/type_util.py
```diff
@@ -0,0 +1,93 @@
...
+
+def is_annotated(annotation: Any):
+    """Check annotation has Annotated type or not."""
+    return get_origin(annotation) is Annotated
+
+
+_Types: TypeAlias = Union[type, Tuple["_Types", ...]]
+
+
+def has_annotated_metadata(annotation: Any, type_: _Types) -> bool:
```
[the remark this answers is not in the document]
jeongukjae: Oh good catch!

src/hera/_utils/type_util.py
```diff
@@ -0,0 +1,93 @@
...
+except ImportError:
+    UnionType = Union  # type: ignore
+
+try:
+    from typing import Annotated, get_args, get_origin  # type: ignore
+except ImportError:
+    from typing_extensions import Annotated, get_args, get_origin  # type: ignore
+
+
+def is_annotated(annotation: Any):
```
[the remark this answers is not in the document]
jeongukjae: okay. that looks better

src/hera/_utils/type_util.py
```diff
@@ -0,0 +1,93 @@
...
+def has_annotated_metadata(annotation: Any, type_: _Types) -> bool:
+    """If given annotation has metadata typed type_, return True."""
+    args = get_args(annotation)
+    for arg in args[1:]:
+        if isinstance(arg, type_):
+            return True
+    return False
+
+
+def consume_annotated_type(annotation: Any) -> type:
```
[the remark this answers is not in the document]
jeongukjae: `unwrap_annotation` looks better!

src/hera/_utils/type_util.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+@overload
+def consume_annotated_metadata(_: Any, __: Tuple[Type[T], Type[V]]) -> Optional[Union[T, V]]: ...
+
+
+# FIXME: Currently, mypy cannot guess following type hint properly: https://github.com/python/mypy/issues/17700
+#        def consume_annotated_metadata(_: Any, __: Union[Type[T], Tuple[Type[T], ...]]) -> Optional[T]: ...
+#        Once fixed, remove overloads and add simpler type hints.
+def consume_annotated_metadata(annotation, type_):
```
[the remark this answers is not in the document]
jeongukjae: Makes sense. Thanks!

src/hera/_utils/type_util.py
```diff
@@ -0,0 +1,93 @@
...
+            for t in type_:
+                if isinstance(arg, t):
+                    return arg
+        else:
+            if isinstance(arg, type_):
+                return arg
+    return None
+
+
+def may_cast_subscripted_type(t: type) -> type:
```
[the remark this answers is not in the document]
jeongukjae: Makes sense.
\+ I added this function because of the cast(type, ...) statement. I don't want to raise errors when checking isinstance or issubclass.

src/hera/workflows/_runner/script_annotations_util.py
```diff
@@ -132,12 +128,11 @@ def map_runner_input(
     If the field is annotated, we look for the kwarg with the name from the annotation (Parameter or Artifact).
     Otherwise, we look for the kwarg with the name of the field.
     """
-    from hera.workflows._runner.util import _get_type
-
     input_model_obj = {}
 
     def load_parameter_value(value: str, value_type: type) -> Any:
-        if issubclass(_get_type(value_type), str):
+        raw_type = type_util.consume_annotated_type(value_type)
```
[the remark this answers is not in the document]
jeongukjae: Oh thank you. good catch!

src/hera/_utils/type_util.py
jeongukjae: I agree with naming issue, and thank you for simpler codes!

src/hera/_utils/type_util.py
jeongukjae: I looked into it, but it doesn't seem to be used in here. anyway thanks for suggestion!

src/hera/_utils/type_util.py
```diff
@@ -0,0 +1,93 @@
+"""Module that handles types and annotations."""
```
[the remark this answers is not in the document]
jeongukjae: In this decision, I don't have strong opinion. If there will be more util files, could be better to have _utils module. If not, could be better to have type_util in shared module.

For now, I will place this in shared!

jeongukjae:
Checked all review comments, feel free to check this PR again :)

samj1912:
@alicederyn can you also review this please.

src/hera/workflows/script.py
```diff
@@ -578,11 +567,10 @@ def _extract_all_output_annotations(source: Callable) -> List:
...
     for _, func_param in inspect.signature(source).parameters.items():
-        if get_origin(func_param.annotation) is Annotated:
-            annotation_args = get_args(func_param.annotation)
-            annotated_type = annotation_args[1]
-            if isinstance(annotated_type, (Artifact, Parameter)) and annotated_type.output:
-                output.append(annotation_args)
+        if (
+            annotated := _type_util.get_annotated_metadata(func_param.annotation, (Artifact, Parameter))
+        ) and annotated.output:
+            output.append(get_args(func_param.annotation))
```
elliotgunton, line 573: I think this was technically a bug in the old code (but the actual list of value is never used so it didn't surface) and this should be
```suggestion
            output.append(annotated)
```
Which also saves another call :)

src/hera/workflows/script.py
```diff
@@ -555,7 +544,7 @@ def _extract_return_annotation_output(source: Callable) -> List:
     return_annotation = inspect.signature(source).return_annotation
     origin_type = get_origin(return_annotation)
     annotation_args = get_args(return_annotation)
-    if origin_type is Annotated:
+    if _type_util.is_annotated(return_annotation):
         output.append(annotation_args)
```
elliotgunton, line 548: I think this is also slightly buggy old code due to the assumption of only having 1 piece of metadata that must be an Artifact or Parameter. We can use the new `get_annotated_metadata(return_annotation, (Artifact, Parameter))` here.

```suggestion
    if param_or_artifact := get_annotated_metadata(return_annotation, (Artifact, Parameter)):
        output.append(param_or_artifact)
```

src/hera/workflows/script.py
```diff
@@ -383,14 +383,14 @@ def append_annotation(annotation: Union[Artifact, Parameter]):
...
-        append_annotation(get_args(return_annotation)[1])
+    if param_or_artifact := _type_util.get_annotated_metadata(return_annotation, (Artifact, Parameter)):
+        append_annotation(param_or_artifact)
     elif get_origin(return_annotation) is tuple:
         for annotation in get_args(return_annotation):
             if isinstance(annotation, type) and issubclass(annotation, (OutputV1, OutputV2)):
                 raise ValueError("Output cannot be part of a tuple output")
 
-            append_annotation(get_args(annotation)[1])
+            append_annotation(_type_util.get_annotated_metadata(annotation, (Artifact, Parameter)))
```
elliotgunton, line 393: We need to revisit this code because of #1167 anyway, so I would suggest the same `if param_or_artifact :=` pattern here, and remove the `None` from the `append_annotation` union:

```suggestion
            if param_or_artifact := _type_util.get_annotated_metadata(return_annotation, (Artifact, Parameter)):
                append_annotation(param_or_artifact)
```

src/hera/workflows/script.py
```diff
@@ -372,7 +372,7 @@ def _get_outputs_from_return_annotation(
     parameters = []
     artifacts = []
 
-    def append_annotation(annotation: Union[Artifact, Parameter]):
+    def append_annotation(annotation: Union[Artifact, Parameter, None]):
```
elliotgunton, line 375:
```suggestion
    def append_annotation(annotation: Union[Artifact, Parameter]):
```

elliotgunton (commented):
Looking really good! Some minor nitpicks/opportunities to clean up, then LGTM! 🚀

src/hera/shared/_type_util.py
```diff
@@ -0,0 +1,78 @@
...
+def get_annotated_metadata(_: type, __: Type[T]) -> Optional[T]: ...
+
+
+@overload
+def get_annotated_metadata(_: type, __: Tuple[Type[T], Type[V]]) -> Optional[Union[T, V]]: ...
+
+
+# FIXME: Currently, mypy cannot guess following type hint properly: https://github.com/python/mypy/issues/17700
+#        def get_annotated_metadata(_: Any, __: Union[Type[T], Tuple[Type[T], ...]]) -> Optional[T]: ...
+#        Once fixed, remove overloads and add simpler type hints.
```
alicederyn, line 42: We can solve this for our limited usage with `typing.overload`.

```python
@overload
def get_annotated_metadata(annotation: Any, type_: Type[T]) -> Optional[T]: ...

@overload
def get_annotated_metadata(annotation: Any, type_: Tuple[Type[T1], Type[T2]]) -> Optional[Union[T1, T2]]: ...
```

alicederyn (changes requested):
Hitting review timebox for now (needed to poke at mypy and pyright to answer some questions); will pick up again once these issues are addressed :)

src/hera/workflows/script.py
elliotgunton: IMO the refactor has already incorporated a lot of implicit assumption fixes as described in https://github.com/argoproj-labs/hera/pull/1168/files#r1734338480 which would have to be undone and verified.

We've simplified a lot of the `if X is Annotated` checks which were implicitly tying up `if Annotated` with the assumption that the annotation is a `Parameter` or `Artifact`. I doubt users have `Annotated` with non-Hera metadata in the context of Hera if they want their code to work as expected, plus it's still an experimental feature so no need to be super careful.

I think it would be simpler to fix everything at once as it would be harder to guarantee it as a strict refactor (and I'm not fussed about it needing to be a strict refactor).

src/hera/shared/_type_util.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+def get_origin_or_builtin(t: type) -> type:
+    """Get the unsubscripted version of t or the type of t itself if it's a built it, and cast it as a Python `type`.
+
+    This can be helpful if you want to use t with isinstance, issubclass, etc.,
+    """
+    if origin_type := get_origin(t):
+        return cast(type, origin_type)
+    return t
```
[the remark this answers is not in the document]
jeongukjae: some types are not the instance of type class. especially union.

```py
>>> from typing import Union, get_origin
>>> isinstance(get_origin(Union[int, str]), type)
False
>>> isinstance(Union, type)
False
```

src/hera/workflows/io/_io_mixins.py
```diff
@@ -123,14 +121,10 @@ def _get_as_templated_arguments(cls) -> Self:
...
-                    name = annotation.name
-                    if isinstance(annotation, Parameter):
-                        object_dict[field] = "{{inputs.parameters." + f"{name}" + "}}"
-                    elif isinstance(annotation, Artifact):
-                        object_dict[field] = "{{inputs.artifacts." + f"{name}" + "}}"
+            if param := _type_util.get_annotated_metadata(annotations[field], Parameter):
+                object_dict[field] = "{{inputs.parameters." + f"{param.name}" + "}}"
+            elif artifact := _type_util.get_annotated_metadata(annotations[field], Artifact):
+                object_dict[field] = "{{inputs.artifacts." + f"{artifact.name}" + "}}"
             else:
```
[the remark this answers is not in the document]
jeongukjae: yep. makes sense. I'm strongly sure that `_get_parameters` only want to find `Parameter` instance or non-`Artifact` function inputs, so I'll fix that more specifically.

also, in terms of label, I agree.

src/hera/shared/_type_util.py
```diff
@@ -0,0 +1,78 @@
...
+            for t in type_:
+                if isinstance(arg, t):
+                    return arg
+        else:
+            if isinstance(arg, type_):
+                return arg
+    return None
+
+
+def get_origin_or_builtin(t: type) -> type:
```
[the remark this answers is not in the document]
jeongukjae: great point. let me fix that as `get_unsubscripted_type`

jeongukjae:
Thank you all for great reviews! I applied it @elliotgunton @alicederyn

src/hera/shared/_type_util.py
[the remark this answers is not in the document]
alicederyn: Good point...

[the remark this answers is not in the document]
alicederyn: (Though we could still simplify the code to `return cast(type, get_origin(t) or t)`)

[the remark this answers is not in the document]
alicederyn: If we're passing unions into this function then it seems as if it should be annotated as `(Any) -> Any`, since e.g. `int | str` is not considered a `type` by type-checkers, and neither is `UnionType`.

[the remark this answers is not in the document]
elliotgunton: In this case the signature of the function should probably use `Any` (as we are expecting `Union`s to possibly appear in `origin_type_issubclass` below).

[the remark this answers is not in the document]
jeongukjae: I checked related codes and just found that we don't need cast(type, ...) anymore. also, any -> any is right. fixed it!

src/hera/workflows/io/_io_mixins.py
```diff
@@ -166,39 +145,16 @@ def _get_as_arguments(self) -> ModelArguments:
...
-                if isinstance(annotation, Parameter) and annotation.name:
-                    params.append(ModelParameter(name=annotation.name, value=templated_value))
-                elif isinstance(annotation, Artifact) and annotation.name:
-                    artifacts.append(ModelArtifact(name=annotation.name, from_=templated_value))
+            if (param := get_annotated_metadata(annotations[field], Parameter)) and param.name:
+                params.append(ModelParameter(name=param.name, value=templated_value))
+            elif (artifact := get_annotated_metadata(annotations[field], Artifact)) and artifact.name:
+                artifacts.append(ModelArtifact(name=artifact.name, from_=templated_value))
             else:
                 params.append(ModelParameter(name=field, value=templated_value))
```
alicederyn, line 153: Is this definitely correct? I'm surprised we want to create a `ModelParameter` if something is annotated as an `Artifact` but happens to be missing the `name`. Same question for the code below.

alicederyn (commented):
The code is looking a lot cleaner! Just a few questions about some of the subtleties.

src/hera/workflows/_meta_mixins.py
```diff
@@ -181,17 +176,14 @@ def build_model(
...
-                    mapper = get_args(annotation)[1]
+                if mapper := get_annotated_metadata(annotation, ModelMapperMixin.ModelMapper):
                     # Value comes from builder function if it exists on hera_obj, otherwise directly from the attr
                     value = (
                         getattr(hera_obj, mapper.builder.__name__)()
                         if mapper.builder is not None
                         else getattr(hera_obj, attr)
                     )
-                    if value is not None:
+                    if value is not None and mapper.model_path:
```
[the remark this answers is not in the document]
jeongukjae: I changed this to remove type errors (mapper got clear type than before, so mypy raises type error here), but at the second thought, it should be better to update default model_path as empty list. It doesn't change behaviors and also, it won't raise type errors as well.

src/hera/workflows/_meta_mixins.py
```diff
@@ -155,7 +150,7 @@ class ModelMapperMixin(BaseMixin):
 
     class ModelMapper:
         def __init__(self, model_path: str, hera_builder: Optional[Callable] = None):
-            self.model_path = None
```
jeongukjae, line 158: This one is to remove mypy errors. Some code got more clear types due to the refactoring, so calling `_get_model_attr` start to raise type errors. So updating the model_path to be `list[str]`, not `optional[list[str]]` to avoid errors. Checked related codes and pretty sure this doesn't change the current behaviors.

src/hera/workflows/script.py
```diff
@@ -591,11 +575,8 @@ def _extract_all_output_annotations(source: Callable) -> List:
     output = []
 
     for _, func_param in inspect.signature(source).parameters.items():
-        if get_origin(func_param.annotation) is Annotated:
-            annotation_args = get_args(func_param.annotation)
-            annotated_type = annotation_args[1]
-            if isinstance(annotated_type, (Artifact, Parameter)) and annotated_type.output:
-                output.append(annotation_args)
+        if (annotated := get_annotated_metadata(func_param.annotation, (Artifact, Parameter))) and annotated.output:
+            output.append(annotated)
```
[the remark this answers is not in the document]
jeongukjae: Agree. great catch. Updated to append annotation_args again.
For the `_extract_all_output_annotations`, it is used when building Script template, so we cannot delete it.

src/hera/workflows/io/_io_mixins.py
jeongukjae: I'm not 100% sure it is correct behavior, but found that this also changes the behavior. Maybe we don't have test cases for not having parameter or artifact (or its name) while annotated. I'll update this to act like before, and will follow up in next pr.

src/hera/workflows/script.py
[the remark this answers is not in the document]
alicederyn: Ha, my mistake! I did a search and it only showed the definition; I missed the small "and one more"

src/hera/workflows/_meta_mixins.py
alicederyn: I defer to @elliotgunton on this change 😄

alicederyn (approved):

src/hera/workflows/io/_io_mixins.py
elliotgunton: It's not really correct but I want to revisit name-related script runner stuff for some bugfixes/enhancements, see https://github.com/argoproj-labs/hera/issues/1165. It would be in non-`io_mixin` code too so better as a separate PR

tests/test_unit/test_shared_type_utils.py
```diff
@@ -0,0 +1,81 @@
...
+        [str, Parameter, None],
+        [Annotated[str, Parameter(name="a_str")], Parameter, Parameter(name="a_str")],
+        [Annotated[str, "some metadata"], Parameter, None],
+        # Must support variadic annotated
+        [Annotated[str, "some metadata", Parameter(name="a_str")], Parameter, Parameter(name="a_str")],
+        # Must support multiple types
+        [Annotated[str, "some metadata", Parameter(name="a_str")], (Parameter, Artifact), Parameter(name="a_str")],
+        # Must consume in order
+        [
+            Annotated[str, "some metadata", Artifact(name="a_str"), Parameter(name="a_str")],
```
elliotgunton, line 50: I'm not a fan of allowing multiple `Parameter`/`Artifact`s in the metadata, we should make this an error (then make the PR a minor version bump). This could be done within `get_annotated_metadata`, or `get_annotated_metadata` should return a list of all metadata matching the type and the caller should check `len(...) == 1`. Option 2 would be a more-correct function implementation (otherwise how do you get the second piece of metadata with the same type?), but it's a private function so... 🤷 Can do option 1 if 2 is too annoying to implement/use

src/hera/workflows/_meta_mixins.py
elliotgunton: Looks like a sensible fix :shipit:

elliotgunton (changes requested):
From the test doing `Annotated[str, "some metadata", Artifact(name="a_str"), Parameter(name="a_str")]`, I realised that using multiple Artifact/Param annotations should be an error, and this PR that allows any-position metadata is a good time to fix it.

Technically it's not backwards compat, but we were always only using the first piece of metadata anyway, so the PR itself is a bugfix/enhancement to an experimental feature, and adding an error should help devs uncover bugs if they double-annotated through type aliases or something:

```py
MyParamType = Annotated[int, Parameter(name="param-input")]
MyOtherParam = Annotated[MyParamType, Parameter(name="another-name")]
```

I don't think this would come up much but better to fix now, as the behaviour might be unexpected/buggy

src/hera/workflows/script.py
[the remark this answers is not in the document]
elliotgunton: This was a small bug fix I asked for in https://github.com/argoproj-labs/hera/pull/1168/files#r1734309545

We don't use the list (for now) but it would make more sense to get the `Artifact`/`Parameter` metadata rather than ALL the metadata (my bad from previous implementation 😆)

src/hera/workflows/script.py
elliotgunton: Could we implement this change @jeongukjae ?

tests/test_unit/test_shared_type_utils.py
jeongukjae: @elliotgunton wow great. thanks! I couldn't catch it.

There is some callee that calls this with other types rather than Artifact or Parameter (ModelMapper), so I'll update this as option 2.

jeongukjae:
I updated to guard multiple type annotations: https://github.com/argoproj-labs/hera/pull/1168/files/77757b9bef13ee4f749736d57125e06120a54a65..5df6da25630dd50c77022e7fa2d196547b146972
Thank you again @elliotgunton !

src/hera/shared/_type_util.py
```diff
@@ -0,0 +1,106 @@
...
+
+
+@overload
+def get_annotated_metadata(annotation: Any, type_: Tuple[Type[T], Type[V]]) -> List[Union[T, V]]: ...
+
+
+def get_annotated_metadata(annotation, type_):
+    """If given annotation has metadata typed type_, return the metadata.
+
+    Prefer get_workflow_annotatede_metadata if you want to call this with Artifact or Parameter.
```
elliotgunton, line 49: I think you renamed the function 😆 
```suggestion
    Prefer get_workflow_annotation if you want to call this with Artifact or Parameter.
```

src/hera/shared/_type_util.py
```diff
@@ -0,0 +1,106 @@
...
+def get_workflow_annotation(annotation: Any) -> Optional[Union[Artifact, Parameter]]:
+    """If given annotation has Artifact or Parameter metadata, return it.
+
+    Note that this function will raise the error when multiple Artifact or Parameter metadata are given.
+    """
+    metadata = get_annotated_metadata(annotation, (Artifact, Parameter))
+    if not metadata:
+        return None
+    if len(metadata) > 1:
+        raise ValueError("Annotation should have one or zero artifact and parameter annotation.")
```
elliotgunton, line 75: Prefer to capitalise Artifact/Parameter, and my suggestion for the error string:

```suggestion
        raise ValueError("Annotation metadata cannot contain more than one Artifact/Parameter.")
```

elliotgunton (commented):
Minor suggestion on error string then LGTM! 🚀

src/hera/shared/_type_util.py
jeongukjae: Oh my you are right. haha thank you!

elliotgunton (approved):
🚀

samj1912 (approved):

PR #1177, opened 2024-08-27
Use sys.version_info to guard imports

alicederyn:
**Pull Request Checklist**
- [ ] Fixes #
- [ ] Tests added

**Description of PR**
Currently, conditional imports that depend on Python version are using `try/except ImportError` to import from `typing_extensions` on older Python versions. This is not natively supported by typecheckers, forcing use of error suppression, and losing potential validation of downstream code. (Additionally, pyright is erroneously flagging a lot of code that uses `Annotated` with "Call expression not allowed in type expression, making it harder to spot real issues in VSCode.)

This PR refactors these conditional imports to use the recommended `if sys.version_info >= (3, minor):` pattern instead.

See also: [mypy issue comment](https://github.com/python/mypy/issues/1393#issuecomment-1694796858), [pyright issue comment](https://github.com/microsoft/pyright/discussions/4160#discussioncomment-4082791)

### DIFF
--- a/src/hera/workflows/io/_io_mixins.py
+++ b/src/hera/workflows/io/_io_mixins.py
@@ -1,6 +1,26 @@
+import sys
 import warnings
 from pathlib import Path
 from typing import TYPE_CHECKING, List, Optional, Union
+
+if sys.version_info >= (3, 9):
+    from typing import Annotated
+else:
+    from typing_extensions import Annotated
+
+if sys.version_info >= (3, 10):
+    from typing import get_args, get_origin
+else:
+    # Python 3.8 has get_origin() and get_args() but those implementations aren't
+    # Annotated-aware. Python 3.9's versions don't support ParamSpecArgs and
+    # ParamSpecKwargs.
+    from typing_extensions import get_args, get_origin
+
+if sys.version_info >= (3, 11):
+    from typing import Self
+else:
+    from typing_extensions import Self
+
 
 from hera.shared._pydantic import _PYDANTIC_VERSION, get_field_annotations, get_fields
 from hera.shared.serialization import MISSING, serialize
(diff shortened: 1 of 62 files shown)


src/hera/workflows/io/_io_mixins.py
```diff
@@ -1,7 +1,27 @@
...
+else:
+    from typing_extensions import Annotated
+
+if sys.version_info >= (3, 10):
+    from typing import get_args, get_origin
+else:
+    # Python 3.8 has get_origin() and get_args() but those implementations aren't
+    # Annotated-aware. Python 3.9's versions don't support ParamSpecArgs and
+    # ParamSpecKwargs.
+    from typing_extensions import get_args, get_origin
```
alicederyn, line 17: The old version of this code was importing `Self` in the try/except block, causing it to fall back to the bespoke implementation in `typing_extensions` for Python 3.9. I'm unclear if this was intentional, so retained this behaviour to avoid any potential regression. For Python 3.10, `typing_extensions` imports Python's built-in version, so we can import it directly.

src/hera/workflows/script.py
```diff
@@ -24,6 +25,12 @@
     cast,
 )
 
+if sys.version_info >= (3, 9):
+    from typing import Annotated
+else:
+    from typing_extensions import Annotated
+
+
 from typing_extensions import ParamSpec, get_args, get_origin
```
alicederyn, line 34: Unclear if this was intentionally not guarded, so left as-is.

jeongukjae, line 34: Not strong opinion. looks great if we update this statement as well.

jeongukjae (approved):
:+1:

src/hera/workflows/script.py
alicederyn: 👍  Will leave to a separate PR to avoid risking a behaviour change

samj1912 (approved):

PR #1180, opened 2024-08-28
Use overloads in script decorator types

### LINKED ISSUE #1174: Mypy call-arg errors when using script decorator

- [X] No

- [ ] exporting to YAML
- [ ] submitting to Argo
- [X] other: Type checking

<h2>Bug report</h2>

**Describe the bug**

`@script`-decorated functions cannot be used without `# type: ignore [call-arg]  # pyright: ignore [reportCallIssue]`.

_Mypy errors_:
```
error: Unexpected keyword argument "arguments"  [call-arg]
error: Missing named argument "name"  [call-arg]
Revealed type is "Union[None, hera.workflows.steps.Step, hera.workflows.task.Task]"
```

_pyright errors_:
```
error: Argument missing for parameter "name" (reportCallIssue)
error: Argument missing for parameter "job_message" (reportCallIssue)
error: No parameter named "arguments" (reportCallIssue)
Type of "result" is "Unknown"
```

**To Reproduce**
```py
from typing import TYPE_CHECKING
from hera.workflows import Steps, Workflow, script

@script()
def some_script(job_message: str) -> None:
  print(job_message)

    with Steps(name="steps"):
        result = some_script(arguments={"job_message": "Hello world!"})
        if TYPE_CHECKING:
            reveal_type(result)
```

**Expected behavior**
No type-checker errors. Type revealed to be `Step | Task`.

**Environment**
- Hera Version: 5.X.X
- Python Version: 3.9
- Argo Version: N/A

**Additional context**
I believe the issue comes down to the use of a union of callables here: https://github.com/argoproj-labs/hera/blob/523a9b91d5af34a6f797907d5d02ac4a89240d6a/src/hera/workflows/script.py#L624-L628

That means that the function could return any three of those callables, so callers should only make calls which are compatible with all three. This is technically correct, as without knowledge of the context, any of these three types could be returned; however, it makes the decorator very awkward to use.

I suggest we instead return a single Protocol with an overloaded function definition that assumes that the function is being called inside a context block if the parameters match, and otherwise falls back to the original signature. This would allow users to call script functions (both in context blocks and in tests) without a `type: ignore`. Unfortunately, this would require writing out a full signature on the first overload, with a test to keep the duplicates in-sync, rather than using `ParamSpec` to capture the signatures of `Step` and `Task`, as neither pyright nor mypy support having two overloads both using `ParamSpec`.

The return type would still be `Step | Task`, so users using a type-checker may need to `assert isinstance(result, Step)` or `assert isinstance(result, Task)` to pass validation. This remaining issue is covered by #837 and #911.

_note: Some of the following discussion was made prior to this description being edited to focus on fixing the call-arg issue._

alicederyn:
**Pull Request Checklist**
- [X] Fixes #1174
- [X] Tests added

**Description of PR**
Currently, @script-decorated user functions return a union of callables, making it impossible to type-check without suppressing mypy's 'call-arg' error (or equivalent), which completely disables parameter type-checking.

This PR changes the type annotations to use an overloaded callable Protocol to select the signature most likely to be what the user was intending to pick.

Unfortunately, the call signatures to Step and Task have to be duplicated, as:
 - mypy and pyright do not support having multiple ParamSpec-decorated overloads
 - the `name` parameter is required in Step and Task but optional when calling a @script-decorated user function

### VSCode tooltips
The help text here comes from the docstrings in the new Protocol; to obtain this, I had to suppress [ruff's D418 rule](https://docs.astral.sh/ruff/rules/overload-with-docstring/).

First VSCode tooltip, showing a no-argument overload that may return a step, a task, or None
Second VSCode tooltip, showing an overload that may return a step or task
Third VSCode tooltip, showing an overload that will return a task
Final VSCode tooltip, showing an overload that will call the user code

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -133,7 +133,7 @@ src = ["src"]
 [tool.ruff.lint]
 select = ["E", "F", "D"]
 ignore = [
-  "D418",  # Bans docstrings on overloads. Pylance supports these and they are useful.
+  "D418",  # Bans docstrings on overloads, but VS Code displays these to users in tooltips.
   "E501",  # Line too long.
 ]
 extend-select = ["I"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -132,7 +132,10 @@ src = ["src"]
 
 [tool.ruff.lint]
 select = ["E", "F", "D"]
-ignore = ["E501"]
+ignore = [
+  "D418",  # Bans docstrings on overloads. Pylance supports these and they are useful.
+  "E501",  # Line too long.
+]
 extend-select = ["I"]
 
 [tool.ruff.lint.per-file-ignores]
--- a/src/hera/workflows/script.py
+++ b/src/hera/workflows/script.py
@@ -630,6 +630,16 @@ class _ScriptDecoratedFunction(Generic[FuncIns, FuncRCov], Protocol):
     @overload
     def __call__(  # type: ignore [overload-overlap]
         self,
+    ) -> Optional[Union[Step, Task]]:
+        """@script-decorated function invoked within a workflow, step or task context.
+
+        May return None, a Step, or a Task, depending on the context. Use `assert isinstance(result, Step)`
+        or `assert isinstance(result, Task)` to select the correct type if using a type-checker.
+        """
+
+    @overload
+    def __call__(  # type: ignore [overload-overlap]
+        self,
         *,
         name: str = ...,
         continue_on: Optional[ContinueOn] = ...,
--- a/src/hera/workflows/script.py
+++ b/src/hera/workflows/script.py
@@ -625,8 +625,6 @@ ScriptIns = ParamSpec("ScriptIns")  # For input attributes of Script class
 class _ScriptDecoratedFunction(Generic[FuncIns, FuncRCov], Protocol):
     """Type assigned to functions decorated with @script."""
 
-    # Note: name is required for Step + Task, but is automatically inferred from
-    #       the name of the decorated function for @script.
     # Note: For more details about overload-overlap, see https://github.com/python/typeshed/issues/12178
 
     @overload
@@ -651,6 +649,8 @@ class _ScriptDecoratedFunction(Generic[FuncIns, FuncRCov], Protocol):
         May return a Step or a Task, depending on the context. Use `assert isinstance(result, Step)`
         or `assert isinstance(result, Task)` to select the correct type if using a type-checker.
         """
+        # Note: signature must match the Step constructor, except that while name is required for Step,
+        # it is automatically inferred from the name of the decorated function for @script.
 
     @overload
     def __call__(  # type: ignore [overload-overlap]
@@ -672,6 +672,8 @@ class _ScriptDecoratedFunction(Generic[FuncIns, FuncRCov], Protocol):
         depends: Optional[str] = ...,
     ) -> Task:
         """@script-decorated function invoked within a task context."""
+        # Note: signature must match the Task constructor, except that while name is required for Task,
+        # it is automatically inferred from the name of the decorated function for @script.
 
     @overload
     def __call__(self, *args: FuncIns.args, **kwargs: FuncIns.kwargs) -> FuncRCov:
--- a/src/hera/workflows/script.py
+++ b/src/hera/workflows/script.py
@@ -14,15 +14,19 @@ from functools import wraps
 from typing import (
     Any,
     Callable,
+    Dict,
+    Generic,
     List,
     Literal,
     Optional,
+    Protocol,
     Sequence,
     Tuple,
     Type,
     TypeVar,
     Union,
     cast,
+    overload,
 )
 
 if sys.version_info >= (3, 9):
@@ -45,8 +49,10 @@ from hera.shared.serialization import serialize
 from hera.workflows._context import _context
 from hera.workflows._meta_mixins import CallableTemplateMixin
 from hera.workflows._mixins import (
+    ArgumentsT,
     ContainerMixin,
     EnvIOMixin,
+    OneOrMany,
     ResourceMixin,
     TemplateMixin,
     VolumeMountMixin,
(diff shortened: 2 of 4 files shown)


src/hera/workflows/script.py
```diff
@@ -606,27 +617,81 @@ def _output_annotations_used(source: Callable) -> bool:
...
+        name: str = ...,
+        continue_on: Optional[ContinueOn] = ...,
+        hooks: Optional[Dict[str, LifecycleHook]] = ...,
+        on_exit: Optional[Union[str, Templatable]] = ...,
+        template: Optional[Union[str, _ModelTemplate, TemplateMixin, CallableTemplateMixin]] = ...,
+        template_ref: Optional[TemplateRef] = ...,
+        inline: Optional[Union[_ModelTemplate, TemplateMixin]] = ...,
+        when: Optional[str] = ...,
+        with_sequence: Optional[_ModelSequence] = ...,
+        arguments: ArgumentsT = ...,
```
alicederyn, line 645: While technically all of these are optional, I wonder if we should make `arguments` required here? At the moment, there's no way to correctly unit-test the underlying function if it takes no parameters, or it happens to have parameters with the same name as some of these. It would be easy to put `arguments={}` in your step/task if you don't want to pass arguments.

tests/typehints/test_script_decorator.py
```diff
@@ -0,0 +1,109 @@
...
+    print(input)
+    return len(input)
+""".strip()
+
+
+def run_mypy(python_code: str):
+    with TemporaryDirectory() as d:
+        python_file = Path(d) / "example.py"
+        python_file.write_text(python_code)
+        mypy_cmd = ["mypy", "--config-file", "tests/typehints/test-mypy.toml", str(python_file)]
```
alicederyn, line 25: These tests are a little bit slow (a couple of seconds each?) because they're invoking mypy here. We could combine them into fewer tests (e.g. one test that passes every Step parameter at once instead of one test per parameter), at the cost of making test failures more complex to diagnose

src/hera/workflows/script.py
```diff
@@ -606,27 +617,81 @@ def _output_annotations_used(source: Callable) -> bool:
...
 FuncR = TypeVar("FuncR")  # For return type of given func to script decorator
+FuncRCov = TypeVar("FuncRCov", covariant=True)
 
 ScriptIns = ParamSpec("ScriptIns")  # For input attributes of Script class
-StepIns = ParamSpec("StepIns")  # # For input attributes of Step class
-TaskIns = ParamSpec("TaskIns")  # # For input attributes of Task class
 
 
-# Pass actual classes of Script, Step and Task to bind inputs to the ParamSpecs above
+class _ScriptDecoratedFunction(Generic[FuncIns, FuncRCov], Protocol):
```
jeongukjae, line 625: question. I want to make sure I understood alright or not.
Script has lots of arguments than this class is overloading. And doesn't seem to be covered through typevar. are these arguments are selected to resolve type issues? or any other reason for this?

pyproject.toml
```diff
@@ -132,7 +132,10 @@ src = ["src"]
 
 [tool.ruff.lint]
 select = ["E", "F", "D"]
-ignore = ["E501"]
+ignore = [
+  "D418",  # Bans docstrings on overloads. Pylance supports these and they are useful.
```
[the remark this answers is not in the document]
alicederyn: We could ignore it on the specific file, but given that Pylance (VSCode's content assistance plugin) uses docstrings from overloads, I thought it might accidentally discourage use in future if we didn't disable it globally (and it also is a good place to explain why we're suppressing it). Making them comments instead of docstrings will stop Pylance picking them up.

I will add in a screenshot of this information showing up in VSCode to the PR description as motivation, thanks for flagging!

src/hera/workflows/script.py
alicederyn: Only the parameters to Step and Task are supported here, not the full set of arguments to Script (those are instead used at a higher level in the decorator call chain, specifically they are accepted by `@script` itself).

src/hera/workflows/script.py
```diff
@@ -606,27 +617,81 @@ def _output_annotations_used(source: Callable) -> bool:
...
+    """Type assigned to functions decorated with @script."""
+
+    # Note: name is required for Step + Task, but is automatically inferred from
+    #       the name of the decorated function for @script.
```
jeongukjae, line 629: thanks for answering! much clear. other things looks good.
how about adding other note comment here saying like following parameters should match with the parameter of step or task? I think this is the first place to look when the mypy raises the errors about script decorated objects.

jeongukjae (approved):

alicederyn:
Just spotted that I missed the Workflow overload. Moving back to draft until I fix and test that.

src/hera/workflows/script.py
```diff
@@ -606,27 +617,83 @@ def _output_annotations_used(source: Callable) -> bool:
...
+        with_param: Optional[Any] = ...,
+        with_items: Optional[OneOrMany[Any]] = ...,
+    ) -> Union[Step, Task]:
+        """@script-decorated function invoked within a step or task context.
+
+        May return a Step or a Task, depending on the context. Use `assert isinstance(result, Step)`
+        or `assert isinstance(result, Task)` to select the correct type if using a type-checker.
+        """
+        # Note: signature must match the Step constructor, except that while name is required for Step,
+        # it is automatically inferred from the name of the decorated function for @script.
```
elliotgunton, line 663: Just a small clarification
```suggestion
        # it is automatically inferred from the name of the decorated function when invoked.
```

tests/typehints/test_script_decorator.py
```diff
@@ -0,0 +1,109 @@
...
+    STEP_INVOCATION = """
+        result = simple_script()
+        reveal_type(result)
+    """
+    result = run_mypy(SIMPLE_SCRIPT + dedent(STEP_INVOCATION))
+    assert 'Revealed type is "Union[hera.workflows.steps.Step, hera.workflows.task.Task, None]"' in result
+
+
+def test_parameter_named_name():
+    """If a script has a 'name' argument, the Step/Task overload will take precedence."""
```
elliotgunton, line 68: I don't think this is the right way - if we take a scenario where a user is adding the Hera library to their existing code and adds the `@script` decorator to a function, the user's lint checks will break. We are an optional wrapper and should be treating the code as such.

src/hera/workflows/script.py
```diff
@@ -606,27 +617,83 @@ def _output_annotations_used(source: Callable) -> bool:
 
 FuncIns = ParamSpec("FuncIns")  # For input types of given func to script decorator
 FuncR = TypeVar("FuncR")  # For return type of given func to script decorator
+FuncRCov = TypeVar("FuncRCov", covariant=True)
```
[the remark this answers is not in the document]
alicederyn: The type checker failed without it, complaining that we shouldn't be using an invariant TypeVar in the definition of `_ScriptDecoratedFunction` when we should be using a covariant one. But it is technically correct to do so: a `_ScriptDecoratedFunction[Foo, list[str]]` can safely be passed where a `_ScriptDecoratedFunction[Foo, Sequence[str]]` is required, for instance, because anything that returns a list is returning a sequence.

src/hera/workflows/script.py
alicederyn: With the change in precedence to prefer the user's function, this makes perfect sense, thanks.

PR #1183, opened 2024-08-31
Add optional Session to Service to be able to reuse connections

AchilleAsh:
**Pull Request Checklist**
- [ ] Fixes #
- [x] Tests added

**Description of PR**

The WorkflowService uses direct calls to requests library, which under the hood creates a Session, makes the request, then closes the session.

This means that the TCP connection is opened and closed for each call instead of being reused - which causes unnecessary overhead if the service is making multiple calls.

This PR adds the possibility to use a Session object in the Service, by default not enabled to keep the current behavior as-is.

From a performance point of view, if Argo is running locally the advantage will be virtually non-existent ; however, we can simulate the outcome with a remote Argo server  using [toxiproxy](https://github.com/Shopify/toxiproxy/):

```bash
toxiproxy-cli create --listen localhost:2747 --upstream localhost:2746 argo 
toxiproxy-cli toxic add -t latency -a latency=20 -a jitter=5 argo  
```

Using ipython and the magic command %timeit:
```
from hera.workflows import WorkflowsService

# default current behavior: no session
ws = WorkflowsService("https://localhost:2747", verify_ssl=False)

%timeit -n10 -r10 ws.list_workflows("argo")
133 ms ± 2.17 ms per loop (mean ± std. dev. of 10 runs, 10 loops each)

ws = WorkflowsService("https://localhost:2747", verify_ssl=False, use_session=True)
%timeit -n10 -r10 ws.list_workflows("argo")
77.4 ms ± 2.01 ms per loop (mean ± std. dev. of 10 runs, 10 loops each)
```

In toxiproxy logs (and the kubectl port-forward logs) , we can see that the tcp open is done now only once.

Thanks!

### DIFF
--- a/scripts/service.py
+++ b/scripts/service.py
@@ -468,7 +468,7 @@ class {models_type}Service:
         
     def __exit__(self, *_):
         \"\"\"Close the service.\"\"\"
-        self.session.close()
+        self.close()
 """
 
 
--- a/scripts/service.py
+++ b/scripts/service.py
@@ -176,7 +176,8 @@ class ServiceEndpoint:
         return f"""
     {signature}
         assert valid_host_scheme(self.host), "The host scheme is required for service usage"
-        resp = requests.{self.method}(
+        resp = self.request(
+            method="{self.method}",
             url={req_url},
             params={params},
             headers={headers},
@@ -422,6 +423,7 @@ class {models_type}Service:
         token: Optional[str] = None,
         client_certs: Optional[Tuple[str, str]] = None,
         namespace: Optional[str] = None,
+        use_session: Optional[bool] = None,
     ) -> None:
         \"\"\"{models_type} service constructor.\"\"\"
         self.host = cast(str, host or global_config.host)
@@ -448,7 +450,20 @@ class {models_type}Service:
         else:
             self.token = None
 
+        use_session = use_session if use_session is not None else global_config.use_session
+        if use_session:
+            self.session: Optional[requests.Session] = requests.Session()
+        else: 
+            self.session = None
+
         self.namespace = namespace or global_config.namespace
+        
+    def request(self, method, **kwargs):
+        \"\"\"Make a request using the session if enabled.\"\"\"
+        if self.session is None:
+            return requests.request(method, **kwargs)
+        else:
+            return self.session.request(method, **kwargs)
 """
 
 
--- a/scripts/service.py
+++ b/scripts/service.py
@@ -176,7 +176,7 @@ class ServiceEndpoint:
         return f"""
     {signature}
         assert valid_host_scheme(self.host), "The host scheme is required for service usage"
-        resp = self.request(
+        resp = self._request(
             method="{self.method}",
             url={req_url},
             params={params},
@@ -423,7 +423,7 @@ class {models_type}Service:
         token: Optional[str] = None,
         client_certs: Optional[Tuple[str, str]] = None,
         namespace: Optional[str] = None,
-        use_session: Optional[bool] = None,
+        session: Optional[requests.Session] = None,
     ) -> None:
         \"\"\"{models_type} service constructor.\"\"\"
         self.host = cast(str, host or global_config.host)
@@ -450,20 +450,25 @@ class {models_type}Service:
         else:
             self.token = None
 
-        use_session = use_session if use_session is not None else global_config.use_session
-        if use_session:
-            self.session: Optional[requests.Session] = requests.Session()
-        else: 
-            self.session = None
+        self.session = session or requests.Session()
 
         self.namespace = namespace or global_config.namespace
         
-    def request(self, method, **kwargs):
+    def _request(self, method, **kwargs):
         \"\"\"Make a request using the session if enabled.\"\"\"
-        if self.session is None:
-            return requests.request(method, **kwargs)
-        else:
-            return self.session.request(method, **kwargs)
+        return self.session.request(method, **kwargs)
+    
+    def close(self):
+        \"\"\"Close the service session.\"\"\"
+        self.session.close()    
+    
+    def __enter__(self):
+        \"\"\"Open the service - session doesn't need to be opened.\"\"\"
+        return self
+        
+    def __exit__(self, *_):
+        \"\"\"Close the service.\"\"\"
+        self.session.close()
 """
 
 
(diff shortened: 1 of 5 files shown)


src/hera/events/service.py
```diff
@@ -73,8 +74,21 @@ def format_token(t):
...
 
+        use_session = use_session if use_session is not None else global_config.use_session
+        if use_session:
+            self.session: Optional[requests.Session] = requests.Session()
+        else:
+            self.session = None
+
         self.namespace = namespace or global_config.namespace
 
+    def request(self, method, **kwargs):
```
alicederyn, line 85: I'd mark this private; I don't think we want this added to the public API. (Same for WorkflowsService.)
```suggestion
    def _request(self, method, **kwargs):
```

src/hera/events/service.py
```diff
@@ -47,6 +47,7 @@ def __init__(
         token: Optional[str] = None,
         client_certs: Optional[Tuple[str, str]] = None,
         namespace: Optional[str] = None,
+        use_session: Optional[bool] = None,
```
alicederyn, line 50: Would it not be more flexible to pass in a `session` directly? (Same for WorkflowsService.)

elliotgunton:
> by default not enabled to keep the current behavior as-is

Just out of curiosity - is there any reason not to enable it by default, i.e. is it not a strict enhancement to go from `request.get` to `self.session.get`? What could break? Looks like a straight upgrade from [docs](https://requests.readthedocs.io/en/latest/user/advanced/) and [stack overflow](https://stackoverflow.com/questions/32986228/difference-between-using-requests-get-and-requests-session-get) as `request.get` creates a one-time-use `Session` anyway.

Thinking about client usage of `WorkflowsService` with a `Session`, I think it might be better to implement context management for `WorkflowsService` for basic `Session` usage:

```py
with WorkflowsService(...) as ws:
    ws.create(...)
    ws.list(...)
```

This would use sessions implicitly. If we want to also allow users to pass in the session for flexibility as per @alicederyn's comment, then they can follow a similar pattern.

```py
with request.Session(...) as session:
    ws = WorkflowsService(session=session, ...)
    ws.create(...)
    ws.list(...)
```

It seems reusing a "closed" session just reopens it, and you can close multiple times, so using a "dead" session doesn't seem to be a problem:
```py
>>> import requests
>>> session = requests.Session()
>>> session.get("https://example.com/")
<Response [200]>
>>> session.close()
>>> session.get("https://example.com/")
<Response [200]>
>>> session.close()
>>> session.close()
```

alicederyn:
I was unable to get a clear statement that sessions were thread-safe in requests. I did find [a closed issue asking for it to be documented one way or another] (https://github.com/psf/requests/issues/2766) but I couldn't tell what the close status was intended to be. I'm also not sure if the service classes are intended to be thread-safe themselves?

AchilleAsh:
> Just out of curiosity - is there any reason not to enable it by default, i.e. is it not a strict enhancement to go from `request.get` to `self.session.get`? 

Oh, I fully agree :) and I think its fits anyway the "spirit" of having a WorkflowsService: it makes sense for it to hold a session to the Argo Server for the duration of its existence. I just initially thought of introducing this change as opt-in, but if we all agree then it can be the default.

Note that the Workflow class [creates a default WorkflowsService if not provided](https://github.com/argoproj-labs/hera/blob/main/src/hera/workflows/workflow.py#L276), so this would also implicitly initialize and keep one session alive for the duration of the Workflow object (i.e. keep connections opened if calls were made) - and this would also not call it as context manager. So maybe the default service for workflows shouldn't use a session?


> Thinking about client usage of `WorkflowsService` with a `Session`, I think it might be better to implement context management for `WorkflowsService` for basic `Session` usage:

I can add a `.close()` method to the Service which explicitly calls the `.close()` of the Session, and then yes it's just convenient to expose the context manager interface too.


> This would implicitly use a single session. If we want to also allow users to pass in the session for flexibility as per @alicederyn's comment, then they can follow a similar pattern.


Perhaps it would be better to let the WorkflowService open and close the user provided Session, rather than relying on the user to open/close it. Also because we need to do it anyway if the user doesn't pass a custom Session and we use the regular one.


```python
session = MyCustomSession()
with WorkflowsService(session=session) as ws:
    # will open/close the session
    ws.list(...)
    with Workflow(..., workflows_service=ws) as wf1:
        ...
    with Workflow(..., workflows_service=ws) as wf2:
        ...
    with Workflow(...) as wf3:
        # wf3.workflows_service != ws
        ...

```

It is probably better to still have to explicitly pass the service to any Workflow created under the context manager block ( `Explicit is better than implicit.`)


> I was unable to find a clear statement that sessions were thread-safe in requests.

It's a long-standing question indeed, I've always sided with the "better safe than sorry" approach and used one session per thread. But since the urllib3 pool is thread safe in the newer versions (2022+), it is *probably* fine

> I'm also not sure if the service classes are intended to be thread-safe themselves?

I feel like it should be thread safe: If you have tens of Workflows to submit at once, it makes sense to use one service and reuse connections. Else you just end up with the current behavior of opening one session thus one connection per call.

scripts/service.py
```diff
@@ -448,7 +450,25 @@ def format_token(t):
...
+        \"\"\"Close the service session.\"\"\"
+        self.session.close()    
+    
+    def __enter__(self):
+        \"\"\"Open the service - session doesn't need to be opened.\"\"\"
+        return self
+        
+    def __exit__(self, *_):
+        \"\"\"Close the service.\"\"\"
+        self.session.close()
```
elliotgunton, line 471: In case we add any more teardown logic, I think we should use `close`

```suggestion
        self.close()
```

src/hera/events/service.py
```diff
@@ -73,8 +74,26 @@ def format_token(t):
...
+        """Close the service session."""
+        self.session.close()
+
+    def __enter__(self):
+        """Open the service - session doesn't need to be opened."""
+        return self
+
+    def __exit__(self, *_):
+        """Close the service."""
+        self.close()
```
elliotgunton, line 95: I think you had it before but didn't regenerate the code?

elliotgunton (commented):
I saw the checks failing with ` Code is not up-to-date. Please run 'make codegen'` - I'll run this with the proposed fix in `__exit__`, then LGTM! 🚀

elliotgunton (approved):
Thank you! 🚀

PR #1186, opened 2024-09-02
Support various output tuple annotation

### LINKED ISSUE #1167: Cannot run script with tuple annotation in return type

jeongukjae:
**Pull Request Checklist**
- [x] Fixes #1167 
- [x] Tests added
- [x] Documentation/examples added; bug fix

**Description of PR**
Currently, hera doesn't support un-annotated tuple output type, and there's no guard for partially annotated type.

So updating hera like as following.
- Support un-annotated tuple output.
- Raise the error when the output tuple is partially annotated.

### DIFF
--- a/src/hera/workflows/script.py
+++ b/src/hera/workflows/script.py
@@ -555,11 +555,15 @@ def _extract_return_annotation_output(source: Callable) -> List:
     if get_workflow_annotation(return_annotation):
         output.append(annotation_args)
     elif origin_type is tuple:
-        workflow_args = [get_args(annotated_type) for annotated_type in annotation_args if get_workflow_annotation(annotated_type)]
+        workflow_args = [
+            get_args(annotated_type) for annotated_type in annotation_args if get_workflow_annotation(annotated_type)
+        ]
         if len(workflow_args) == len(annotation_args):
             output.extend(workflow_args)
         elif workflow_args:
-            raise ValueError("All elements in the tuple should be annotated with Artifact/Parameter.")
+            raise ValueError(
+                f"Function '{source.__name__}' output has partially annotated return type. Hera allows tuple annotation to be fully annotated or not."
+            )
     elif (
         origin_type is None
         and isinstance(return_annotation, type)
--- a/src/hera/workflows/script.py
+++ b/src/hera/workflows/script.py
@@ -555,10 +555,10 @@ def _extract_return_annotation_output(source: Callable) -> List:
     if get_workflow_annotation(return_annotation):
         output.append(annotation_args)
     elif origin_type is tuple:
-        if all(get_workflow_annotation(annotated_type) for annotated_type in annotation_args):
-            for annotated_type in annotation_args:
-                output.append(get_args(annotated_type))
-        elif any(get_workflow_annotation(annotated_type) for annotated_type in annotation_args):
+        workflow_args = [get_args(annotated_type) for annotated_type in annotation_args if get_workflow_annotation(annotated_type)]
+        if len(workflow_args) == len(annotation_args):
+            output.extend(workflow_args)
+        elif workflow_args:
             raise ValueError("All elements in the tuple should be annotated with Artifact/Parameter.")
     elif (
         origin_type is None
--- a/src/hera/workflows/script.py
+++ b/src/hera/workflows/script.py
@@ -558,11 +558,12 @@ def _extract_return_annotation_output(source: Callable) -> List:
         workflow_args = [
             get_args(annotated_type) for annotated_type in annotation_args if get_workflow_annotation(annotated_type)
         ]
+        # If all tuple elements are annotated as Parameter/Artifact
         if len(workflow_args) == len(annotation_args):
             output.extend(workflow_args)
-        elif workflow_args:
+        elif workflow_args:  # Only some tuple elements are annotated as Parameter/Artifact
             raise ValueError(
-                f"Function '{source.__name__}' output has partially annotated return type. Hera allows tuple annotation to be fully annotated or not."
+                f"Function '{source.__name__}' output has partially annotated tuple return type. Tuple elements must be all Annotated as Parameter/Artifact, or contain no Parameter/Artifact annotations for a raw tuple return type."
             )
     elif (
         origin_type is None
--- a/src/hera/workflows/script.py
+++ b/src/hera/workflows/script.py
@@ -555,8 +555,11 @@ def _extract_return_annotation_output(source: Callable) -> List:
     if get_workflow_annotation(return_annotation):
         output.append(annotation_args)
     elif origin_type is tuple:
-        for annotated_type in annotation_args:
-            output.append(get_args(annotated_type))
+        if all(get_workflow_annotation(annotated_type) for annotated_type in annotation_args):
+            for annotated_type in annotation_args:
+                output.append(get_args(annotated_type))
+        elif any(get_workflow_annotation(annotated_type) for annotated_type in annotation_args):
+            raise ValueError("All elements in the tuple should be annotated with Artifact/Parameter.")
     elif (
         origin_type is None
         and isinstance(return_annotation, type)
--- a/src/hera/workflows/script.py
+++ b/src/hera/workflows/script.py
@@ -552,8 +552,8 @@ def _extract_return_annotation_output(source: Callable) -> List:
     return_annotation = inspect.signature(source).return_annotation
     origin_type = get_origin(return_annotation)
     annotation_args = get_args(return_annotation)
-    if param_or_artifact := get_workflow_annotation(return_annotation):
-        output.append(param_or_artifact)
+    if get_workflow_annotation(return_annotation):
+        output.append(annotation_args)
     elif origin_type is tuple:
         if all(get_workflow_annotation(annotated_type) for annotated_type in annotation_args):
             for annotated_type in annotation_args:
--- a/src/hera/workflows/script.py
+++ b/src/hera/workflows/script.py
@@ -558,12 +558,15 @@ def _extract_return_annotation_output(source: Callable) -> List:
         workflow_args = [
             get_args(annotated_type) for annotated_type in annotation_args if get_workflow_annotation(annotated_type)
         ]
+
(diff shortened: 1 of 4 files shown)


src/hera/workflows/script.py
```diff
@@ -555,8 +555,11 @@ def _extract_return_annotation_output(source: Callable) -> List:
     if get_workflow_annotation(return_annotation):
         output.append(annotation_args)
     elif origin_type is tuple:
-        for annotated_type in annotation_args:
-            output.append(get_args(annotated_type))
+        if all(get_workflow_annotation(annotated_type) for annotated_type in annotation_args):
+            for annotated_type in annotation_args:
+                output.append(get_args(annotated_type))
+        elif any(get_workflow_annotation(annotated_type) for annotated_type in annotation_args):
+            raise ValueError("All elements in the tuple should be annotated with Artifact/Parameter.")
```
alicederyn, line 562: I think we should include the function name in the error message and explain that this is an error about the return type annotation

src/hera/workflows/script.py
```diff
@@ -555,8 +555,11 @@ def _extract_return_annotation_output(source: Callable) -> List:
     if get_workflow_annotation(return_annotation):
         output.append(annotation_args)
     elif origin_type is tuple:
-        for annotated_type in annotation_args:
-            output.append(get_args(annotated_type))
+        if all(get_workflow_annotation(annotated_type) for annotated_type in annotation_args):
+            for annotated_type in annotation_args:
+                output.append(get_args(annotated_type))
+        elif any(get_workflow_annotation(annotated_type) for annotated_type in annotation_args):
```
alicederyn, line 561: nit: more DRY to do the transformation loop only once? perhaps
```suggestion
        num_workflow_annotations = sum(1 for annotated_type in annotation_args if get_workflow_annotation(annotated_type))
        if len(annotation_args) == num_workflow_annotations:
            output.extend(get_args(annotated_type) for annotated_type in annotation_args)
        elif num_workflow_annotations > 0:
```

src/hera/workflows/script.py
```diff
@@ -555,8 +555,11 @@ def _extract_return_annotation_output(source: Callable) -> List:
     if get_workflow_annotation(return_annotation):
         output.append(annotation_args)
```
elliotgunton, line 556: This is a good opportunity to clean up the function for correctness and better readability. Also `annotation_args` is only needed if the origin_type is tuple, so we can move that

```suggestion
    if param_or_artifact := get_workflow_annotation(return_annotation):
        output.append(param_or_artifact)
```

jeongukjae: This change breaks the tests. I tried to fix it, but seems the original type (annotation_args[0]) is needed.

Following lines are breaking 😢 
https://github.com/argoproj-labs/hera/blob/2eec4c282557823c661dae6b410b470d13a90be5/src/hera/workflows/_runner/script_annotations_util.py#L241-L254

src/hera/workflows/script.py
jeongukjae: Makes sense. Fixed!

src/hera/workflows/script.py
elliotgunton: This is IMO harder to understand than the use of `all` and `any` - happy to save us doing the transformation twice but could we add a comment to explain?

src/hera/workflows/script.py
```diff
@@ -555,8 +555,15 @@ def _extract_return_annotation_output(source: Callable) -> List:
...
     elif origin_type is tuple:
-        for annotated_type in annotation_args:
-            output.append(get_args(annotated_type))
+        workflow_args = [
+            get_args(annotated_type) for annotated_type in annotation_args if get_workflow_annotation(annotated_type)
+        ]
+        if len(workflow_args) == len(annotation_args):
```
elliotgunton, line 561:
```suggestion
        # If all tuple elements are annotated as Parameter/Artifact
        if len(workflow_args) == len(annotation_args):
```

src/hera/workflows/script.py
```diff
@@ -555,8 +555,15 @@ def _extract_return_annotation_output(source: Callable) -> List:
...
+            output.extend(workflow_args)
+        elif workflow_args:
```
elliotgunton, line 563:
```suggestion
        elif workflow_args:  # Only some tuple elements are annotated as Parameter/Artifact
```

src/hera/workflows/script.py
```diff
@@ -555,8 +555,15 @@ def _extract_return_annotation_output(source: Callable) -> List:
...
+            raise ValueError(
+                f"Function '{source.__name__}' output has partially annotated return type. Hera allows tuple annotation to be fully annotated or not."
```
elliotgunton, line 565:
```suggestion
                f"Function '{source.__name__}' output has partially annotated tuple return type. Tuple elements must be all Annotated as Parameter/Artifact, or contain no Parameter/Artifact annotations for a raw tuple return type."
```

elliotgunton (commented):
Minor fixes then LGTM

src/hera/workflows/script.py
jeongukjae: thanks!

src/hera/workflows/script.py
```diff
@@ -558,11 +558,12 @@ def _extract_return_annotation_output(source: Callable) -> List:
...
             get_args(annotated_type) for annotated_type in annotation_args if get_workflow_annotation(annotated_type)
         ]
+        # If all tuple elements are annotated as Parameter/Artifact
         if len(workflow_args) == len(annotation_args):
             output.extend(workflow_args)
-        elif workflow_args:
+        elif workflow_args:  # Only some tuple elements are annotated as Parameter/Artifact
             raise ValueError(
-                f"Function '{source.__name__}' output has partially annotated return type. Hera allows tuple annotation to be fully annotated or not."
+                f"Function '{source.__name__}' output has partially annotated tuple return type. Tuple elements must be all Annotated as Parameter/Artifact, or contain no Parameter/Artifact annotations for a raw tuple return type."
```
alicederyn: nit: could we split this onto two lines? It's got pretty long.
```suggestion
                f"Function '{source.__name__}' output has partially annotated tuple return type."
                " Tuple elements must be all Annotated as Parameter/Artifact, or contain no"
                " Parameter/Artifact annotations for a raw tuple return type."
```

alicederyn: Ah, you split it already. I think 3 lines would still be better.

elliotgunton (approved):

src/hera/workflows/script.py
jeongukjae: Oh thanks! but merged

Issue #1162, opened 2024-08-20
New dag/steps decorators issues/inconsistencies


elliotgunton:
<h2>Bug report</h2>

**Describe the bug**
_A clear and concise description of what the bug is:_

There are a few issues with the new dag decorator.

- [x] It doesn't accept artifacts because of the `path` existing in the yaml. I'm not sure if DAGs can even accept artifacts in the first place if they are used as tasks themselves. Tried submitting `examples/workflows/experimental/new_dag_decorator_artifacts.py` but got
> Server returned status code 400 with message: `templates.worker.inputs.artifacts.artifact_a.path only valid in container/script templates`

- [x] The mandated Input for a DAG function doesn't really make sense in the current form - if I don't have any inputs I still need one due to the check here https://github.com/argoproj-labs/hera/blob/3470e8e8ec6f416d66e1e9a2a35970f3cbcf1261/src/hera/workflows/_meta_mixins.py#L820

i.e. my function must be:
```py
@w.dag()
def my_dag(_: Input):
```

- [x] The Output is not mandated but you can e.g. set `result` or `exit_code` on it which should probably be an error within a DAG/Steps function.

- [x] DAG tasks must have alphanumeric hyphenated names, meaning the python variable names can't be directly used via `varname`. E.g for the code:

```py
from hera.shared import global_config
from hera.workflows import Input, Output, Workflow

global_config.experimental_features["decorator_syntax"] = True


w = Workflow(generate_name="my-workflow-")


class MyInput(Input):
    user: str


@w.script()
def hello_world(my_input: MyInput) -> Output:
    output = Output()
    output.result = f"Hello Hera User: {my_input.user}!"
    return output


@w.set_entrypoint
@w.dag()
def worker(_: Input):
    task_name = hello_world(my_input=MyInput(user="test"))
```
`argo lint basic-dag.yaml` gets 

```
argo lint manifests/basic-dag.yaml
manifests/basic-dag.yaml:
   ✖ in "my-workflow-" (Workflow): templates.worker.tasks[0].name: 'task_name' is invalid: name must consist of alpha-numeric characters or '-', and must start with an alpha-numeric character (e.g. My-name1-2, 123-NAME)

✖ 1 linting errors found!
```

- [x] A result of the above is **local running doesn't work** when using alternative names as **task kwargs aren't accepted**:
```
python -m hera_demo.advanced_dag
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
[...]

Issue #711, opened 2023-07-14
Automatic Conversion of Argo YAML Files to Hera Code


Leewaldoe:
**Is your feature request related to a problem? Please describe.**
Yes, the problem is that currently, there is no straightforward way to convert Argo yaml files to Hera code. This can be a time-consuming process, especially for large or complex workflows and for those more familiar with yaml but would like to start using Hera. 

**Describe the solution you'd like**
I would like a feature that can automatically convert Argo yaml files into equivalent Hera code. This would involve parsing the yaml file, interpreting its structure and contents, and generating corresponding Hera code that represents the same workflow.

**Describe alternatives you've considered**
The current alternative to this feature is to manually convert Argo yaml files to Hera code. This involves going through examples, understanding the mapping between yaml and Hera syntax, and spending hours rewriting and testing the code. However, this process is time-consuming, labor-intensive, and prone to human error. It also requires a deep understanding of both yaml and Hera, which might be a barrier for some users.

**Additional context**
Adding this feature would not only simplify the process of creating and managing Argo workflows but also provide the following benefits:

- It would make it easier for users to transition from using pure yaml to Hera. By providing a way to automatically convert their existing Argo yaml files into Hera code, we can lower the barrier to entry and encourage more users to adopt Hera.
- It would increase the speed of understanding Hera for those coming from a pure yaml implementation. By seeing how their familiar YAML workflows translate into Hera code, users can more quickly grasp the structure and syntax of Hera, accelerating their learning process.

This feature would serve as a valuable tool for onboarding new users to Hera.

schmidtnz:
Hi hera Team,
How is this ticket tracking in your priority order?
We are looking at potentially migrating to hera for describing our workflows and having a way to convert our YAML instead of implementing them from scratch would be a huge bonus.
Many thanks for giving a quick update.
Cheers & have a happy festive season :-)

gyanprakash48:
i have same situation as @schmidtnz . Is there any estimated timeline , when this feature might  be available

gyanprakash48:
@elliotgunton  by any chance , you can help with approximated timeline . Ticket appears open since 2 yrs . This will help us in making  our decision to switch to Hera. Many Thanks!

elliotgunton:
[...]

PR #1189, opened 2024-09-04
Enable Pydantic I/O types in workflow context

### LINKED ISSUE #1192: Allow using Pydantic I/O in with-based code
Migrating to the new decorator syntax (or, presumably, off it) is made more challenging by having to completely rewrite pipelines from the old `arguments=` syntax to the new Pydantic syntax in one go.

Allow a mix of the new Pydantic syntax and the old `arguments=` syntax within "legacy" `with`-based pipelines. This would let users verify that they are creating the same YAML at more points along the journey, catching mistakes earlier.

We could allow a mixture inside the new syntax instead (or as well). I haven't investigated how much work doing it this way around would be.

**Additional context**
The `script_runner_io.py` example workflow could be rewritten as:
```python
with Workflow(generate_name="pydantic-io-") as w:
    with Steps(name="use-pydantic-io"):
        write_step = writer()
        pydantic_io(
            MyInput(
                artifact_int="{{steps.writer.outputs.artifacts.int-artifact}}",
                param_int=101,
                an_object=MyObject(a_dict={"my-new-key": "my-new-value"}),
            )
        )
```
The `artifact_int` parameter isn't ideal, though there may be a cleaner way to write it. I've been rewriting workflows top-down however, which makes this simpler as you could have used `write_step.int_artifact` if `write_step` had been migrated first.

alicederyn:
**Pull Request Checklist**
- [X] Fixes #1192
- [X] Tests added
- [X] Documentation/examples added

**Description of PR**
Currently, `@script`-decorated functions that use Pydantic I/O types must still use pre-Pydantic syntax when creating workflows. This makes it harder to migrate to the new dag and steps decorators, as more code must be migrated in a single step before the YAML can be regenerated and tested.

This PR allows Pydantic input types to be passed in when in an active context, and will then configure the returned subnode to mimic the Pydantic output type. Tasks using fields from the returned object will automatically have a dependency added, if no dependency is explicitly specified.

### DIFF
--- a/examples/workflows/experimental/pydantic_io_in_steps_context.py
+++ b/examples/workflows/experimental/pydantic_io_in_steps_context.py
@@ -11,7 +11,7 @@ from hera.shared import global_config
 from hera.workflows import Parameter, Steps, WorkflowTemplate, script
 from hera.workflows.io.v1 import Input, Output
 
-global_config.experimental_features["script_pydantic_io"] = True
+global_config.experimental_features["decorator_syntax"] = True
 
 
 class CutInput(Input):
--- a/examples/workflows/experimental/pydantic_io_in_steps_context.py
+++ b/examples/workflows/experimental/pydantic_io_in_steps_context.py
@@ -1,0 +1,53 @@
+import sys
+from typing import List
+
+if sys.version_info >= (3, 9):
+    from typing import Annotated
+else:
+    from typing_extensions import Annotated
+
+
+from hera.shared import global_config
+from hera.workflows import Parameter, Steps, WorkflowTemplate, script
+from hera.workflows.io.v1 import Input, Output
+
+global_config.experimental_features["script_pydantic_io"] = True
+
+
+class CutInput(Input):
+    cut_after: Annotated[int, Parameter(name="cut-after")]
+    strings: List[str]
+
+
+class CutOutput(Output):
+    first_strings: Annotated[List[str], Parameter(name="first-strings")]
+    remainder: List[str]
+
+
+class JoinInput(Input):
+    strings: List[str]
+    joiner: str
+
+
+class JoinOutput(Output):
+    joined_string: Annotated[str, Parameter(name="joined-string")]
+
+
+@script(constructor="runner")
+def cut(input: CutInput) -> CutOutput:
+    return CutOutput(
+        first_strings=input.strings[: input.cut_after],
+        remainder=input.strings[input.cut_after :],
+        exit_code=1 if len(input.strings) <= input.cut_after else 0,
+    )
+
+
+@script(constructor="runner")
+def join(input: JoinInput) -> JoinOutput:
+    return JoinOutput(joined_string=input.joiner.join(input.strings))
+
+
+with WorkflowTemplate(generate_name="pydantic-io-in-steps-context-v1-", entrypoint="d") as w:
+    with Steps(name="d"):
+        cut_result = cut(CutInput(strings=["hello", "world", "it's", "hera"], cut_after=1))
+        join(JoinInput(strings=cut_result.first_strings, joiner=" "))
--- a/examples/workflows/experimental/pydantic_io_in_steps_context.py
+++ b/examples/workflows/experimental/pydantic_io_in_steps_context.py
@@ -11,7 +11,7 @@ from hera.shared import global_config
 from hera.workflows import Parameter, Steps, WorkflowTemplate, script
 from hera.workflows.io.v1 import Input, Output
 
-global_config.experimental_features["decorator_syntax"] = True
+global_config.experimental_features["context_manager_pydantic_io"] = True
 
 
 class CutInput(Input):
--- a/src/hera/workflows/_meta_mixins.py
+++ b/src/hera/workflows/_meta_mixins.py
@@ -377,7 +377,6 @@ class CallableTemplateMixin(BaseMixin):
                 current_task_depends = _context.pieces[-1]._current_task_depends
                 if current_task_depends and "depends" not in kwargs:
                     kwargs["depends"] = " && ".join(sorted(current_task_depends))
-                current_task_depends.clear()
 
                 return Task(template=self, **kwargs)
 
@@ -582,7 +581,6 @@ class TemplateDecoratorFuncsMixin(ContextMixin):
                 depends=" && ".join(sorted(_context.pieces[-1]._current_task_depends)) or None,
                 **kwargs,
             )
-            _context.pieces[-1]._current_task_depends.clear()
 
         subnode._build_obj = HeraBuildObj(subnode._subtype, output_class)
         return subnode
--- a/src/hera/workflows/_meta_mixins.py
+++ b/src/hera/workflows/_meta_mixins.py
@@ -564,7 +564,8 @@ class TemplateDecoratorFuncsMixin(ContextMixin):
             # Set template to None as it cannot be set alongside template_ref
             template = None  # type: ignore
 
-        if isinstance(_context.pieces[-1], (Steps, Parallel)):
+        current_context = _context.pieces[-1]
+        if isinstance(current_context, (Steps, Parallel)):
             subnode = Step(
                 name=subnode_name,
                 template=template,
@@ -572,13 +573,14 @@ class TemplateDecoratorFuncsMixin(ContextMixin):
                 arguments=subnode_args,
                 **kwargs,
             )
-        elif isinstance(_context.pieces[-1], DAG):
+        elif isinstance(current_context, DAG):
+            if current_context._current_task_depends and "depends" not in kwargs:
+                kwargs["depends"] = " && ".join(sorted(current_context._current_task_depends))
             subnode = Task(
                 name=subnode_name,
                 template=template,
                 template_ref=template_ref,
                 arguments=subnode_args,
-                depends=" && ".join(sorted(_context.pieces[-1]._current_task_depends)) or None,
                 **kwargs,
             )
 
--- a/src/hera/workflows/_meta_mixins.py
+++ b/src/hera/workflows/_meta_mixins.py
@@ -373,6 +373,12 @@ class CallableTemplateMixin(BaseMixin):
                 return Step(template=self, **kwargs)
 
             if isinstance(_context.pieces[-1], DAG):
+                # Add dependencies based on context if not explicitly provided
+                current_task_depends = _context.pieces[-1]._current_task_depends
+                if current_task_depends and "depends" not in kwargs:
+                    kwargs["depends"] = " && ".join(sorted(current_task_depends))
+                current_task_depends.clear()
+
                 return Task(template=self, **kwargs)
 
         raise InvalidTemplateCall(
--- a/src/hera/workflows/_meta_mixins.py
+++ b/src/hera/workflows/_meta_mixins.py
@@ -347,12 +347,12 @@ class CallableTemplateMixin(BaseMixin):
 
         from hera.workflows.dag import DAG
         from hera.workflows.script import Script
-        from hera.workflows.steps import Parallel, Step, Steps
-        from hera.workflows.task import Task
+        from hera.workflows.steps import Parallel, Steps
         from hera.workflows.workflow import Workflow
 
         if _context.pieces:
-            if isinstance(_context.pieces[-1], Workflow):
+            current_context = _context.pieces[-1]
+            if isinstance(current_context, Workflow):
                 # Notes on callable templates under a Workflow:
                 # * If the user calls a script directly under a Workflow (outside of a Steps/DAG) then we add the script
                 #   template to the workflow and return None.
@@ -369,16 +369,8 @@ class CallableTemplateMixin(BaseMixin):
                 raise InvalidTemplateCall(
                     f"Callable Template '{self.name}' is not callable under a Workflow"  # type: ignore
                 )
-            if isinstance(_context.pieces[-1], (Steps, Parallel)):
-                return Step(template=self, **kwargs)
-
-            if isinstance(_context.pieces[-1], DAG):
-                # Add dependencies based on context if not explicitly provided
-                current_task_depends = _context.pieces[-1]._current_task_depends
-                if current_task_depends and "depends" not in kwargs:
-                    kwargs["depends"] = " && ".join(sorted(current_task_depends))
-
-                return Task(template=self, **kwargs)
+            if isinstance(current_context, (Steps, Parallel, DAG)):
+                return current_context._create_leaf_node(template=self, **kwargs)
 
         raise InvalidTemplateCall(
             f"Callable Template '{self.name}' is not under a Workflow, Steps, Parallel, or DAG context"  # type: ignore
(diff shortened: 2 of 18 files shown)


src/hera/workflows/_meta_mixins.py
```diff
@@ -539,8 +545,7 @@ def _create_subnode(
         assert _context.pieces
 
         template_ref = None
-        _context.declaring = False
-        if _context.pieces[0] != self and isinstance(self, WorkflowTemplate):
+        if _context.pieces[0] is not self and isinstance(self, WorkflowTemplate):
```
alicederyn, line 548: Note: `!=` in Pydantic raises an "incorrect type" validation error due to having been returned a string from TemplateInvocatorSubNodeMixin's modified `__getattribute__` implementation. We could temporarily disable the active context instead, but switching to `is not` is simpler (and faster). As a side-benefit, we can remove the temporary assignment of `False` to `_context.declaring`, which was previously working around this issue.

src/hera/workflows/script.py
```diff
@@ -760,6 +755,27 @@ def script_wrapper(func: Callable[FuncIns, FuncR]) -> Callable:
         def task_wrapper(*args, **kwargs) -> Union[FuncR, Step, Task, None]:
             """Invokes a `Script` object's `__call__` method using the given SubNode (Step or Task) args/kwargs."""
             if _context.active:
+                if len(args) == 1 and isinstance(args[0], (InputV1, InputV2)):
```
alicederyn, line 758: Note this new behaviour only affects calls made to a `@script`-decorated function when passed a Pydantic I/O object by argument, which previously would not have functioned meaningfully (the object was ignored).

jeongukjae, line 758: These looks like it doesn't support kwargs.

```suggestion
                if (
                    (len(args) == 1 and len(kwargs) == 0 and isinstance(args[0], (InputV1, InputV2))) or
                    (len(args) == 0 and len(kwargs) == 1 and isinstance(next(iter(kwargs.values())), (InputV1, InputV2)))
                ):
                    if args:
                        arguments = args[0]._get_as_arguments()
                    else:
                        arguments = next(iter(kwargs.values()))._get_as_arguments()

```

src/hera/workflows/script.py
```diff
@@ -760,6 +755,27 @@ def script_wrapper(func: Callable[FuncIns, FuncR]) -> Callable:
...
+                    subnode = s.__call__(arguments=arguments_list, **kwargs)
+                    if not subnode:
+                        raise SyntaxError("Cannot use Pydantic I/O outside of a DAG, Steps or Parallel context")
+
+                    output_class = inspect.signature(func).return_annotation
+                    if not output_class or output_class is NoneType:
+                        return None
+
+                    if not issubclass(output_class, (OutputV1, OutputV2)):
+                        raise SyntaxError("Cannot use Pydantic input type without a Pydantic output type")
```
jeongukjae, line 774: it should be better to check these in tests.
seems to be submission tests?

examples/workflows/experimental/pydantic_io_in_steps_context.py
```diff
@@ -0,0 +1,53 @@
...
+
+@script(constructor="runner")
+def join(input: JoinInput) -> JoinOutput:
+    return JoinOutput(joined_string=input.joiner.join(input.strings))
+
+
+with WorkflowTemplate(generate_name="pydantic-io-in-steps-context-v1-", entrypoint="d") as w:
+    with Steps(name="d"):
+        cut_result = cut(CutInput(strings=["hello", "world", "it's", "hera"], cut_after=1))
+        join(JoinInput(strings=cut_result.first_strings, joiner=" "))
```
jeongukjae, line 53: How about calling one of these calls to use keyword arguments?

src/hera/workflows/io/_io_mixins.py
```diff
@@ -39,21 +40,29 @@
...
             # Intercept the declaration to avoid validation on the templated strings
-            # We must then turn off declaring mode to be able to "construct" an instance
+            # We must then disable the active context to be able to "construct" an instance
             # of the InputMixin subclass.
-            _context.declaring = False
-            instance = cls.construct(**kwargs)
-            _context.declaring = True
-            return instance
+            with no_active_context():
+                return cls.construct(**kwargs)
```
jeongukjae, line 60: I don't understand hera's context management 100%, so deferring mixins' review to others.
For me, it seems refine the context state checking, and looks fine.

src/hera/workflows/_meta_mixins.py
```diff
@@ -372,6 +372,12 @@ def __call__(self, *args, **kwargs) -> Union[None, Step, Task]:
                 return Step(template=self, **kwargs)
 
             if isinstance(_context.pieces[-1], DAG):
+                # Add dependencies based on context if not explicitly provided
+                current_task_depends = _context.pieces[-1]._current_task_depends
+                if current_task_depends and "depends" not in kwargs:
+                    kwargs["depends"] = " && ".join(sorted(current_task_depends))
+                current_task_depends.clear()
+
```
jeongukjae, line 380: Not 100% sure.
But this seems more discussions are needed.

e.g. in this pr's example, following code works as expected

```py
        cut_task = cut(CutInput(strings=["hello", "world", "it's", "hera"], cut_after=1))
        join_task = join(JoinInput(strings=cut_task.first_strings, joiner=" "))

        cut_task >> join_task
```

This code can possibly restrict the construction of DAGs.
for example, let's think about following example

```py
A = task_a_fn(SomePydanticClass1(...))
B = task_b_fn(SomePydanticClass2(arg_1=A.return_1, ...)) # B depends on A's output
C = task_c_fn(SomePydanticClass3(...)) # C doesn't depends on A's output.

# But user want to construct DAG as followings.
A >> [B, C]
```

In above case, hera will raise an error. (e.g. already in some task's depends)
https://github.com/argoproj-labs/hera/blob/2eec4c282557823c661dae6b410b470d13a90be5/src/hera/workflows/task.py#L83-L84

src/hera/workflows/script.py
alicederyn: Agreed. I copied that limitation from the existing Pydantic I/O logic here: https://github.com/argoproj-labs/hera/blob/2eec4c282557823c661dae6b410b470d13a90be5/src/hera/workflows/_meta_mixins.py#L530-L532

I think if we want to lift that limitation and support the user passing their Input type as a named parameter, we should do it in both places at once in a follow-up PR.

src/hera/workflows/_meta_mixins.py
alicederyn: In your example, you could just write `A >> C` instead of `A >> [B, C]`

However, I agree that this is an existing limitation of the new Pydantic I/O approach which will need solving separately. My preference would be for user-provided dependency edges to not raise duplication errors (like in your example &mdash; it should be fine to be explicit), and to override automatic ones (e.g. if you do `A.on_success(B)`, that should overwrite the existing dependency between A and B). In the meantime, users can:
 - use the old `A.get_parameter` API to avoid adding dependency edges
 - pass in `requires=""` to disable the automatic generation of requires from dependency edges
 - do `B.requires=""` later

src/hera/workflows/script.py
alicederyn: Added tests

src/hera/workflows/_meta_mixins.py
```diff
@@ -372,6 +372,12 @@ def __call__(self, *args, **kwargs) -> Union[None, Step, Task]:
                 return Step(template=self, **kwargs)
 
             if isinstance(_context.pieces[-1], DAG):
+                # Add dependencies based on context if not explicitly provided
```
alicederyn, line 375: * Note this is different from the implementation in the experimental decorator code, which does not allow a `depends` to be passed at all: https://github.com/argoproj-labs/hera/blob/2eec4c282557823c661dae6b410b470d13a90be5/src/hera/workflows/_meta_mixins.py#L568-L569
* I think this logic should probably be made common -- perhaps moved into DAG?

src/hera/workflows/_meta_mixins.py
```diff
@@ -372,6 +372,12 @@ def __call__(self, *args, **kwargs) -> Union[None, Step, Task]:
...
+                current_task_depends = _context.pieces[-1]._current_task_depends
+                if current_task_depends and "depends" not in kwargs:
+                    kwargs["depends"] = " && ".join(sorted(current_task_depends))
+                current_task_depends.clear()
```
alicederyn, line 379: Pre-existing issue: if the user creates a Task explicitly rather than via a `@script`-decorated function, the dependencies list won't be cleared. Perhaps this should be moved to the point where a Task is registered?

src/hera/workflows/script.py
jeongukjae: sounds great!

elliotgunton: >looks like it doesn't support kwargs

This is by design as we need a way to separate the function input arg from the Task kwargs, i.e. the `Input` should be positional-only so we don't have a kwarg name clash

src/hera/workflows/_meta_mixins.py
jeongukjae: Sounds great, but for me it looks better to have same/expected behavior even with Pydantic I/O.
User might expect to be able to write `A >> [B, C]`, but this raises an error.
I assume (checked very limited cases) other nodes are also doesn't generate `depends` attribute based on the returns/arguments.

How about separating this feature to another PR, and support this for regardless of the pydantic io?
Maybe @elliotgunton can give more opinions?

elliotgunton: > User might expect to be able to write A >> [B, C]

I agree - it would be good to write all my dependencies in one place (and as a bonus in one statement) rather than spread out

> user-provided dependency edges to not raise duplication errors

Yes I think it _might_ make sense (not sure about precedent/existing patterns) to make dependency-setting idempotent, and then use the last call (ignoring whether it was automatic or not):

```py
A.on_success(B)
A.on_failure(B)
```
Will end up only with B's `depends: A.Failed`.

src/hera/workflows/_meta_mixins.py
elliotgunton: > Would it be reasonable for the existing code to change to match?

Yes that would be good! 🙏

src/hera/workflows/_meta_mixins.py
elliotgunton: Yes, makes sense to move it to DAG's `_add_sub` https://github.com/argoproj-labs/hera/blob/e7f38fa958a70e3f8cdf66471263f2656511f7ff/src/hera/workflows/dag.py#L50-L56

src/hera/workflows/script.py
```diff
@@ -760,6 +755,27 @@ def script_wrapper(func: Callable[FuncIns, FuncR]) -> Callable:
...
+
+                    subnode = s.__call__(arguments=arguments_list, **kwargs)
+                    if not subnode:
+                        raise SyntaxError("Cannot use Pydantic I/O outside of a DAG, Steps or Parallel context")
+
+                    output_class = inspect.signature(func).return_annotation
+                    if not output_class or output_class is NoneType:
+                        return None
+
+                    if not issubclass(output_class, (OutputV1, OutputV2)):
```
elliotgunton, line 773: I'm not sure this should be a requirement, especially for gradual migrations and issues like https://github.com/argoproj-labs/hera/issues/1191#issue-2507988901

alicederyn: I'm not sure what to return that would match the type signature though. Perhaps this can be addressed in a later PR?

elliotgunton: Setting the `subnode._build_obj` means users can do `my_task.parameter_name` when passes values between tasks - that behaviour should really only be available for the new decorator contexts as it's very experimental and unproven (see #1162). I think this PR is mixing up the two features which should be kept separate IMO, so I'm struggling to understand the motivation of the PR.

tests/test_pydantic_io_syntax.py
```diff
@@ -0,0 +1,62 @@
...
+def test_output_field_contains_argo_template(global_config_fixture):
+    @script()
+    def triple(input: IntInput) -> IntOutput:
+        return IntOutput(field=input.field * 3)
+
+    with Workflow(name="foo"):
+        with Steps(name="bar"):
+            result = triple(IntInput(field=5)).field
+
+    assert result == "{{steps.triple.outputs.parameters.field}}"
```
elliotgunton, line 29: This is not what the Pydantic IO feature flag was designed for and shouldn't be possible in a normal `Steps` context

elliotgunton (changes requested):
I would like to discuss the motivation behind this PR and how we might tidy it up as it's mixing up the two experimental features in a way that is non-obvious

elliotgunton:
I think I just have 2 issues to resolve for the general direction of the PR - the main one being the code implementing the new decorator behaviour i.e. `result = triple(IntInput(field=5)).field` style, should be asserting on the "new decorators" feature flag.

The second is just if you could motivate the PR with an issue showing how the change in the existing example would work here https://github.com/argoproj-labs/hera/blob/e7f38fa958a70e3f8cdf66471263f2656511f7ff/examples/workflows/experimental/script_runner_io.py#L54-L62

I think the feature is a good stepping stone for users to migrate, but shouldn't be backported to the `script_pydantic_io` flag (as that's been out a while, whereas our funky new syntax is kinda buggy, and I actually wouldn't recommend migrating yet, having tried it out myself).

src/hera/workflows/_meta_mixins.py
alicederyn: Done.

src/hera/workflows/_meta_mixins.py
alicederyn: Done.

src/hera/workflows/script.py
alicederyn: Moved the new behaviour into the decorator_syntax experimental feature for now.

tests/test_pydantic_io_syntax.py
alicederyn: Moved behind the decorator_syntax flag. I think it's important to incremental migration to allow this in a `Steps` context though.

elliotgunton (changes requested):
Updating blocking status of the PR for discussion in #1192. I don't think we should make this drastic of a feature change to the old syntax

elliotgunton:
Hey @alicederyn, we discussed this PR and motivation behind it in the maintainer's meeting ([notes here](https://docs.google.com/document/d/1IpHkxsxWdE0lhgpDj_pXYGotsa3s38koCzawlHyH860/edit)), @samj1912 agreed that we don't want to mix the two syntaxes as that will confuse users, so I've moved the PR into the "draft" state.

It seems the main goal is to avoid a big-bang migration, or at least have a checkpoint approach as you explained. An alternative would be to work on #711 as the new decorators mean we can create a full Argo Workflow spec as Python stubs in Hera. This should mean we can construct the DAG too, so the only work to do would be convert old script templates to use Pydantic IO classes, which should match the YAML (because it would be generated _from it_ (super awesome!!)) and then you'd just copy/paste the function body into the stubs.

Let's discuss when you're back, no rush to reply! 😄

elliotgunton:
Closing as agreed in maintainer meeting discussion (I'd rather not leave the PR open indefinitely) - we need to consider the graduation criteria for the new decorators and then how to migrate once they are matured and ready to graduate.

PR #1197, opened 2024-09-11
Error on Parameter default kwarg usage for inputs

### LINKED ISSUE #812: Input Parameter annotations should only be able to set default via standard Python

elliotgunton:
**Pull Request Checklist**
- [x] Fixes #812
- [x] Tests added

**Description of PR**
Currently, using the `default` kwarg for a `Parameter` in an input Annotation or Hera's Pydantic `Input` class logs a warning message, added in https://github.com/argoproj-labs/hera/pull/1102. This PR makes the code path raise an error, which can be supressed using the `suppress_parameter_default_error` experimental feature flag for 1 minor version until it is removed.


src/hera/workflows/io/_io_mixins.py
```diff
@@ -71,10 +73,16 @@ def _get_parameters(cls, object_override: Optional[Self] = None) -> List[Paramet
...
-                    )
+                    if not global_config.experimental_features[_SUPPRESS_PARAMETER_DEFAULT_ERROR_FLAG]:
+                        warnings.warn(
+                            "Using the default field for Parameters in Annotations is deprecated since v5.16"
+                            "and will be removed in a future minor version, use a Python default value instead. "
+                        )
+                        raise ValueError(
+                            "default cannot be set via the Parameter's default, use a Python default value instead. "
+                            "You can suppress this error by setting "
+                            f'global_config.experimental_features["{_SUPPRESS_PARAMETER_DEFAULT_ERROR_FLAG}"] = True'
```
samj1912, line 84: Should this be an experimental flag?

elliotgunton: What do you mean? Should the error be supressed another way?

samj1912: Yeah, I think so. This is a flag but not really an experiment.

elliotgunton: I was intending this to only be a one-off flag for 1 minor version, so I'd rather not add another feature flag syntax just for 1 flag for 1 version (and then have to keep it around). Plus the supression flag is a nicety instead of the intended purpose of "experimental" letting us make breaking changes, but `Annotated` became a mainstay so I understand the need in this case.

In any case, the `global_config.experimental_features` dictionary should be able to hold whatever we need it to hold related to any experimental features - we haven't put any enforced rules on it.
https://github.com/argoproj-labs/hera/blob/77a214dc92828783b693d823b9c6ad0d69481cbd/src/hera/shared/_global_config.py#L77-L78

flaviuvadan: FWIW I think experimental flag is ok as long as we have a clear deadline/version by which we're going to remove this. Small but do we need a warning and a ValueError? I sense the warning will get lost when the exception is raised

elliotgunton: @flaviuvadan Good idea - I can add an explicit version (5.18?) in the warning and error text?

> the warning will get lost

For the first time seeing the error, yes, the warning will probably not be seen, but then after turning the flag on, we still want the user to see the logged warning.

src/hera/workflows/io/_io_mixins.py
```diff
@@ -71,10 +73,16 @@ def _get_parameters(cls, object_override: Optional[Self] = None) -> List[Paramet
                 if param.name is None:
                     param.name = field
                 if param.default is not None:
-                    warnings.warn(
-                        "Using the default field for Parameters in Annotations is deprecated since v5.16"
-                        "and will be removed in a future minor version, use a Python default value instead"
-                    )
+                    if not global_config.experimental_features[_SUPPRESS_PARAMETER_DEFAULT_ERROR_FLAG]:
+                        warnings.warn(
```
alicederyn, line 90: Should this not be outside the if to match the existing behaviour? (Or perhaps on the else path?)

elliotgunton: Yes good catch, I intended to keep the warning no matter the flag value

samj1912 (approved):

Issue #892, opened 2023-12-12
On-cluster testing improvements


elliotgunton:
Follow up on #652 and related PR #887 summarised from the PR description:

* I'd like to use a subset of our examples
* But we want to assert on more than the status.phase, e.g. output values, running order, skipped nodes etc, which would require bespoke tests for each
* It would also be good to have a wrapper for fetching these values, e.g. getting a node requires iterating through the status.nodes dictionary, which is keyed by hashes, so you have to check displayName from the values.

This issue is then for:
- [ ] beginning a helper class for fetching values etc as described
- [ ] adding more on-cluster tests

elliotgunton:
>adding more on-cluster tests

It would be good to target the main features of Argo/Hera, WIP list:

- [ ] WorkflowTemplates
- [ ] DAGs/Steps
- [ ] Artifacts
- [ ] Volumes
- [ ] Script runner (would need to build an image)

PR #1212, opened 2024-09-27
fix: update dask use case example

Joibel:
dask_kubernetes.classic is deprecated, so migrated this example by following https://kubernetes.dask.org/en/latest/kubecluster_migrating.html


elliotgunton (approved):
Thanks for fixing @Joibel, can confirm this example now actually runs!

Maintainer note: another reason to run more on-cluster tests so we catch these out-of-date examples sooner, added to list in #892

PR #1214, opened 2024-09-27
Disable warnings from model_dump

alicederyn:
**Pull Request Checklist**
- [ ] Fixes #
- [X] Tests added

**Description of PR**
Currently, using Pydantic V2 with the Input/Output mixin types and a non-`str` field prints a `UserWarning` when the class is declared, of the form "UserWarning: Pydantic serializer warning: Expected `int` but got `str` with value `'{{...}}'` - serialized value may not be as expected". This is caused by using `model_dump` on an object constructed with all string values, i.e. one which does not pass the type-checking performed by the method, as part of class creation.

This PR just adds `warnings="none"` to the two uses of `model_dump`, to disable these warnings.

### DIFF
--- a/src/hera/workflows/io/_io_mixins.py
+++ b/src/hera/workflows/io/_io_mixins.py
@@ -138,7 +138,7 @@ class InputMixin(BaseModel):
         if isinstance(self, V1BaseModel):
             self_dict = self.dict()
         elif _PYDANTIC_VERSION == 2 and isinstance(self, V2BaseModel):
-            self_dict = self.model_dump()
+            self_dict = self.model_dump(warnings="none")
 
         for field, _, annotation in _construct_io_from_fields(type(self)):
             # The value may be a static value (of any time) if it has a default value, so we need to serialize it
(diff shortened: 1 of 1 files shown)


tests/conftest.py
```diff
@@ -8,3 +10,12 @@ def global_config_fixture():
     global_config.reset()
     yield global_config
     global_config.reset()
+
+
+@pytest.fixture(autouse=True)
+def fail_on_user_warnings():
```
alicederyn, line 16: I'm in two minds about this fixture &mdash; it does catch this issue and prevent any similar issues in future, but I don't like having to suppress the one deprecation warning that Pydantic appears to have accidentally left as a UserWarning. LMK if you'd rather leave it out, it's a separate commit so easy to revert.

elliotgunton: I would rather not optimise early here as it might get difficult/annoying when we move Hera's core code to Pydantic V2

alicederyn: Sure thing. Removed the fixture.

elliotgunton (approved):

PR #1224, opened 2024-10-03
Fix TypeError in script.py for functions returning non-types

alicederyn:
**Pull Request Checklist**
- [ ] Fixes #
- [X] Tests added

**Description of PR**
Currently, `_get_outputs_from_return_annotation` is always passing the return type annotation (if one exists) to `issubclass`, to check if it is a Pydantic type. Some valid type annotations like `typing.Dict` are not valid types at runtime, i.e. cannot be used with `subclass`, resulting in a runtime `TypeError`.

This PR changes the precondition for entering this code from truthiness to `isinstance(..., type)`, matching the code a few lines above.

### DIFF
--- a/src/hera/workflows/script.py
+++ b/src/hera/workflows/script.py
@@ -407,7 +407,7 @@ def _get_outputs_from_return_annotation(
 
             if param_or_artifact := get_workflow_annotation(annotation):
                 append_annotation(param_or_artifact)
-    elif return_annotation and issubclass(return_annotation, (OutputV1, OutputV2)):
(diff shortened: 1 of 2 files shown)


elliotgunton (approved):
>resulting in a runtime TypeError.

It should be noted this is quite a niche case which is why it hasn't been found til now - the runtime error would only happen if you are using `global_config.experimental_features["script_annotations"] = True` but want to return a non `Annotated` `Dict` directly. (I was confused because I thought I fixed this in #791, but that's for Annotated generics). LGTM! 🚀

PR #1225, opened 2024-10-03
Update contributing installation guide

elliotgunton:
* Users should run `make install` which adds `--all-extras` to the poetry commands
* Also updated some make target text for the `make help` text

Ran into some setup problems for @Joibel on #1212, this PR highlights the need for `make` and `poetry` being installed, and updates the initial command to be `make install` instead of `poetry install`


samj1912 (approved):

Joibel:
Thanks Elliot

Issue #1173, opened 2024-08-23
Support Literal types in script runner


alicederyn:
<h2>Pre-bug-report checklist</h2>

**1. This bug can be reproduced using pure Argo YAML**
- [ ] Yes [👉 Please report a bug to the Argo Workflows GitHub 👈](https://github.com/argoproj/argo-workflows/issues/new/choose)
- [X] No

_If yes, it is more likely to be an Argo bug unrelated to Hera. Please double check before submitting an issue to Hera._

**2. This bug occurs in Hera when...**
- [ ] exporting to YAML
- [ ] submitting to Argo
- [X] running on Argo with the Hera runner
- [ ] other: 

<h2>Bug report</h2>

**Describe the bug**
_A clear and concise description of what the bug is:_

Using a function with a Literal argument in the runner results in the following trace:

_Error log if applicable_:
```
Traceback (most recent call last):
  File "/.../python3.9/runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File ".../python3.9/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File ".../python3.9/site-packages/hera/workflows/runner.py", line 6, in <module>
    _run()
  File "/.../python3.9/site-packages/hera/workflows/_runner/util.py", line 254, in _run
    result = _runner(args.entrypoint, kwargs_list)
  File ".../python3.9/site-packages/hera/workflows/_runner/util.py", line 228, in _runner
    return function(**kwargs)
  File ".../python3.9/site-packages/hera/workflows/_runner/util.py", line 58, in inner
    filtered_kwargs = {key: _parse(value, key, f) for key, value in kwargs.items() if _is_kwarg_of(key, f)}
  File ".../python3.9/site-packages/hera/workflows/_runner/util.py", line 58, in <dictcomp>
    filtered_kwargs = {key: _parse(value, key, f) for key, value in kwargs.items() if _is_kwarg_of(key, f)}
  File ".../python3.9/site-packages/hera/workflows/_runner/util.py", line 89, in _parse
    if _is_str_kwarg_of(key, f) or _is_artifact_loaded(key, f) or _is_output_kwarg(key, f):
  File ".../python3.9/site-packages/hera/workflows/_runner/util.py", line 142, in _is_str_kwarg_of
    return issubclass(type_, str)
TypeError: issubclass() arg 1 must be a class
```

**To Reproduce**
Full Hera code to reproduce the bug:
```py
from typing import Literal
[...]

PR #1236, opened 2024-10-10
Fix issues with with_param

### LINKED ISSUE #861: Script annotation inputs doesn't play well with `with_param`

- [x] No

<h2>Bug report</h2>

**Describe the bug**
When using script annotations, an alternative name can be specified. When using `with_param`, it checks through the python names only, [here](https://github.com/argoproj-labs/hera/blob/ce9f04d8baa4241148317e681095b820ae894d83/src/hera/workflows/_mixins.py#L1102-L1107). When using the alternative name in an `arguments={"param-name": "a value"}` alongside a `with_param`, the `with_param` will populate the other arguments as `param_name`.

This will then fail on submission to Argo with
```
inputs.parameters.param-name was not supplied
```

_I'll add a reproducible example here soon!_

**Expected behavior**
Annotated names are also checked/linked with the python names (then probably always use the alternative name if specified)

**Environment**
- Hera Version: 5.9.0
- Python Version: 3.11.6
- Version of Argo: 3.X.X

**Additional context**

### LINKED ISSUE #1234: Issue using with_param and Pydantic IO

- [X] No

**2. I have searched for [existing issues](https://github.com/argoproj-labs/hera/issues)**
- [X] Yes

**3. This bug occurs in Hera when...**
- [X] exporting to YAML
- [ ] submitting to Argo
- [ ] other: 

<h2>Bug report</h2>

**Describe the bug**
When using `with_param` with a function using Pydantic IO, an extra argument is added with a name matching the incidental argument name used by the script-decorated function, not the parameters from the Pydantic Input type. 

**To Reproduce**
```py
from hera.workflows import Workflow, Input, Output, script, Steps

global_config.experimental_features["script_pydantic_io"] = True

class ListOutput(Output):
    values: list[str]

class PrintInput(Input):
    value: str

@script()
def produce_list() -> ListOutput:
    return ListOutput(values=["item1", "item2"])

@script()
def print_value(input: PrintInput) -> None:
    print("Value:", input.value)


    with Steps(name="steps"):
        list_step = produce_list(arguments={})
        print_value(with_param=list_step.get_parameter("values"))
```

**Expected behavior**
```yaml
- arguments:
    parameters:
    - name: value  # This is actually "input"
      value: '{{item}}'
```

**Environment**
- Hera Version: 5.17.1
- Python Version: 3.11.8
- Argo Version: N/A

**Additional context**
This is the same root cause as https://github.com/argoproj-labs/hera/issues/861, i.e. the code in `_get_params_items_from_source`, but can't be worked around by avoiding using alternative names. https://github.com/argoproj-labs/hera/blob/a08f744cd9d7ca30091e7ab51e80bd18e3116010/src/hera/workflows/_meta_mixins.py#L266-L288

alicederyn:
**Pull Request Checklist**
- [X] Fixes #861 and fixes #1234
- [X] Tests added
- [X] Documentation/examples added

**Description of PR**
Currently, the logic implementing `with_param` assumes that Argo Parameters exactly match script function parameters, including the Python parameter name. This does not support the experimental `script_annotations` and `script_pydantic_io` features.

This PR extracts common logic for creating an IO object from parameter/field metadata into a `construct_io_from_annotation` function, which copies any IO annotation it finds, sets the name if missing, and defaults to a Parameter if no annotation is found. It then uses this to generalize `_get_param_items_from_source`, including:

* pulling the parameters from a Pydantic Input type instead, if present
* respecting the name in the Parameter annotation, if present
* ignoring `output=True` annotations
* ignoring `Artifact` annotations


src/hera/workflows/_meta_mixins.py
```diff
@@ -275,17 +286,24 @@ def _get_param_items_from_source(source: Callable) -> List[Parameter]:
...
+            if parameter.default is None:
+                non_default_parameters.append(parameter)
+    else:
+        for p in inspect.signature(source).parameters.values():
+            if p.default == inspect.Parameter.empty and p.kind == inspect.Parameter.POSITIONAL_OR_KEYWORD:
+                # only add positional or keyword arguments that are not set to a default value
+                # as the default value ones are captured by the automatically generated `Parameter` fields for positional
+                # kwargs. Otherwise, we assume that the user sets the value of the parameter via the `with_param` field
+                io = construct_io_from_annotation(p.name, p.annotation)
+                if isinstance(io, Parameter) and io.default is None and not io.output:
```
alicederyn, line 301: I wasn't sure whether it made sense to include `Artifact`s here or not; if this was the wrong decision, LMK and I'll fix it.

alicederyn:
It looks like the code coverage is broken, because it's complaining about lines not being tested that definitely *are*.

src/hera/workflows/io/_io_mixins.py
```diff
@@ -45,18 +45,12 @@
 def _construct_io_from_fields(cls: Type[BaseModel]) -> Iterator[Tuple[str, FieldInfo, Union[Parameter, Artifact]]]:
     """Constructs a Parameter or Artifact object for all Pydantic fields based on their annotations.
 
-    If a field has a Parameter or Artifact annotation, a copy will be returned, with name added if missing.
-    Otherwise, a Parameter object will be constructed.
+    If a field has a Parameter or Artifact annotation, a copy will be returned, with missing
+    fields filled out based on other metadata. Otherwise, a Parameter object will be constructed.
```
alicederyn, line 49: Note: I changed this docstring because the core logic is now in a different function, so it will be easy to miss changes in future and bitrot. I intend to change the behaviour as part of fixing #1173, for instance, to set the enum field for Literals.

src/hera/workflows/_meta_mixins.py
jeongukjae: for me it looks reasonable while reading callee function (it is `_get_deduped_params_from_source`, and returns List[Parameter])

src/hera/workflows/io/_io_mixins.py
```diff
@@ -45,18 +45,12 @@
...
     annotations = get_field_annotations(cls)
     for field, field_info in get_fields(cls).items():
-        if annotation := get_workflow_annotation(annotations[field]):
-            # Copy so as to not modify the fields themselves
-            annotation_copy = annotation.copy()
-            annotation_copy.name = annotation.name or field
-            yield field, field_info, annotation_copy
-        else:
-            yield field, field_info, Parameter(name=field)
+        yield field, field_info, construct_io_from_annotation(field, annotations[field])
```
jeongukjae, line 53: :+1:

src/hera/workflows/script.py
```diff
@@ -434,25 +439,19 @@ def _get_outputs_from_parameter_annotations(
     artifacts: List[Artifact] = []
 
     for name, p in inspect.signature(source).parameters.items():
-        annotation = get_workflow_annotation(p.annotation)
-        if not annotation or not annotation.output:
+        annotation = construct_io_from_annotation(name, p.annotation)
```
jeongukjae, line 442: seems better to have tests for the changes in `src/hera/workflows/script.py` (`_get_outputs_from_parameter_annotations` and `_extract_all_output_annotations`)
even if someone updates `construct_io_from_annotation`, we can make sure this is working as expected.

jeongukjae (commented):
other things looks good

src/hera/workflows/_meta_mixins.py
alicederyn: Perhaps we should be raising an error instead?

jeongukjae (approved):
looks good to me :)

src/hera/workflows/_meta_mixins.py
elliotgunton: We only want to inspect parameters in this function. We shouldn't raise an error (as Artifact defaults work differently)

src/hera/workflows/_meta_mixins.py
```diff
@@ -275,17 +287,24 @@ def _get_param_items_from_source(source: Callable) -> List[Parameter]:
...
+                    non_default_parameters.append(io)
+
+    for param in non_default_parameters:
+        param.value = "{{" + ("item" if len(non_default_parameters) == 1 else f"item.{param.name}") + "}}"
```
elliotgunton, line 306: This seems harder to read/understand IMO

examples/workflows/experimental/script_runner_io_dynamic_fanout.py
```diff
@@ -0,0 +1,38 @@
+"""
```
elliotgunton, line 1: I don't think we need to add these two examples - they are not demonstrating something unique - script annotations/pydantic IO have their own examples, and the syntax for `with_param` is more simply shown in the loops examples, where the syntax in the DAG construction is the same.

alicederyn: Fair, I added them to exercise the code rather than as examples per se. Is there a better place to put this kind of end-to-end YAML output test?

elliotgunton: You could do a roundtrip `to_dict`/`from_dict` like in the `test_script_annotations` tests https://github.com/argoproj-labs/hera/blob/a08f744cd9d7ca30091e7ab51e80bd18e3116010/tests/test_script_annotations.py#L373-L374

src/hera/workflows/_meta_mixins.py
alicederyn: It can't be as succinct as before because we're doing a side-effect. A couple of options spring to mind, which do you prefer?

```code
# Just move the if/else out into a separate line
for param in non_default_parameters:
    param_label = "item" if len(non_default_parameters) == 1 else f"item.{param.name}"
    param.value = "{{" + param_label + "}}"
```

```code
# Move the if/else back out to the top
if len(non_default_parameters) == 1:
    non_default_parameters[0].value = "{{item}}"
else:
    for param in non_default_parameters:
        param.value = "{{{item." + param.name + "))
```

sambhav (approved):

PR #1239, opened 2024-10-15
Fix deserialization for str unions

alicederyn:
**Pull Request Checklist**
- [ ] Fixes #
- [X] Tests added

**Description of PR**
PR #1168 changed the logic of `map_runner_input` and `_parse` to no longer pass incoming values to `json.loads` if their annotated type was a union that included str, rather than only when given a subtype of `Optional[str]`. This makes it impossible to, for instance, pass an int to a `Union[str, int]`.

This PR splits `origin_type_issubclass` into two functions, `origin_type_issupertype` (which matches the previous behaviour) and `origin_type_issubtype`, and use the latter instead to restore the original behaviour.

Additionally, it fixes a bug where we were passing an annotation to `issubclass` without first checking if it's a type, resulting in a `TypeError` (partially addresses #1173).

### DIFF
--- a/src/hera/shared/_type_util.py
+++ b/src/hera/shared/_type_util.py
@@ -92,13 +92,22 @@ def get_unsubscripted_type(t: Any) -> Any:
     return t
 
 
-def origin_type_issubclass(cls: Any, type_: type) -> bool:
-    """Return True if cls can be considered as a subclass of type_."""
-    unwrapped_type = unwrap_annotation(cls)
+def origin_type_issubtype(annotation: Any, type_: Union[type, Tuple[type, ...]]) -> bool:
+    """Return True if annotation is a subtype of type_."""
+    unwrapped_type = unwrap_annotation(annotation)
     origin_type = get_unsubscripted_type(unwrapped_type)
     if origin_type is Union or origin_type is UnionType:
-        return any(origin_type_issubclass(arg, type_) for arg in get_args(cls))
+        return all(origin_type_issubtype(arg, type_) for arg in get_args(annotation))
     return issubclass(origin_type, type_)
+
+
+def origin_type_issupertype(annotation: Any, type_: type) -> bool:
+    """Return True if annotation is a supertype of type_."""
+    unwrapped_type = unwrap_annotation(annotation)
+    origin_type = get_unsubscripted_type(unwrapped_type)
+    if origin_type is Union or origin_type is UnionType:
+        return any(origin_type_issupertype(arg, type_) for arg in get_args(annotation))
+    return issubclass(type_, origin_type)
 
 
 def is_subscripted(t: Any) -> bool:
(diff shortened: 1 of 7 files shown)


tests/test_runner.py
```diff
@@ -65,6 +65,24 @@
...
+                    id="str-or-int-given-str",
+                ),
+                pytest.param(
+                    "tests.script_runner.parameter_inputs:str_or_int_parameter",
+                    [{"name": "my_str_or_int", "value": "3"}],
+                    "type given: int",
+                    id="str-or-int-given-int",
+                ),
+            ]
+            if _PYDANTIC_VERSION > 1
```
alicederyn, line 83: The `parse_obj_as` function we use for Pydantic v1 does not appear to correctly handle `Union[int, str]`; it is serializing ints back to strs: https://github.com/argoproj-labs/hera/blob/e7ed95b692fad12ae49bc2598116d89dc5ee8abd/src/hera/workflows/_runner/util.py#L103

Rather than dig into this, I've disabled these two tests on the Pydantic v1 suite for now.

src/hera/workflows/_runner/util.py
```diff
@@ -125,7 +131,7 @@ def _get_unannotated_type(key: str, f: Callable) -> Optional[type]:
 def _is_str_kwarg_of(key: str, f: Callable) -> bool:
     """Check if param `key` of function `f` has a type annotation that can be interpreted as a subclass of str."""
     if func_param_annotation := _get_function_param_annotation(key, f):
-        return origin_type_issubclass(func_param_annotation, str)
+        return origin_type_issubtype(func_param_annotation, (str, NoneType))
```
jeongukjae, line 134: :+1: Thanks

src/hera/shared/_type_util.py
```diff
@@ -92,15 +92,24 @@ def get_unsubscripted_type(t: Any) -> Any:
     return t
 
 
-def origin_type_issubclass(cls: Any, type_: type) -> bool:
-    """Return True if cls can be considered as a subclass of type_."""
-    unwrapped_type = unwrap_annotation(cls)
+def origin_type_issubtype(annotation: Any, type_: Union[type, Tuple[type, ...]]) -> bool:
+    """Return True if annotation is a subtype of type_."""
```
jeongukjae, line 96: should be better to write a note for tuple args.

maybe
`type_ may be tuple of types. In that case, return true if annotation is a subtype of union of items in type_.`

jeongukjae (commented):
Looks great to me. just small one.

jeongukjae (approved):
looks great

tests/test_runner.py
sambhav: Probably related to https://docs.pydantic.dev/1.10/usage/model_config/#smart-union

sambhav, line 83: Why not use pytest.skipif with pytest.params?

alicederyn: Amazing! Fixed, thanks.

alicederyn: Yep, that would have made sense. No longer needed though as I reordered the union elements so the test passes.

sambhav (approved):

PR #1249, opened 2024-10-28
Support Literal types in script runner

### LINKED ISSUE #1173: Support Literal types in script runner

alicederyn:
**Pull Request Checklist**
- [X] Fixes #1173
- [X] Tests added

**Description of PR**
Currently, the script runner does not support `Literal` types, as `origin_type_issubtype` does not special-case them, so `_is_str_kwarg_of` does not return True.

This PR adds supports for Literals to `origin_type_issubtype`, and additionally copies their values from the annotation into the `enum` field on input Parameters. It also combines two paths in `_get_inputs_from_callable` into one by using `construct_io_from_annotation` instead of having a fallback branch if `get_workflow_annotation` returns `None`; this fixes a bug where the default was being verified as None for optional strings only if there was no IO annotation.

### DIFF
--- a/src/hera/shared/_type_util.py
+++ b/src/hera/shared/_type_util.py
@@ -83,11 +83,16 @@ def get_workflow_annotation(annotation: Any) -> Optional[Union[Artifact, Paramet
     return metadata[0]
 
 
-def add_metadata_from_type(parameter: Parameter, annotation: Any) -> None:
-    if not parameter.enum:
-        type_ = unwrap_annotation(annotation)
-        if get_origin(type_) is Literal:
-            parameter.enum = list(get_args(type_))
+def set_enum_based_on_type(parameter: Parameter, annotation: Any) -> None:
+    """Sets the enum field of a Parameter based on its type annotation.
+
+    Currently, only supports Literals.
+    """
+    if parameter.enum:
+        return
+    type_ = unwrap_annotation(annotation)
+    if get_origin(type_) is Literal:
+        parameter.enum = list(get_args(type_))
 
 
 def construct_io_from_annotation(python_name: str, annotation: Any) -> Union[Parameter, Artifact]:
@@ -107,7 +112,7 @@ def construct_io_from_annotation(python_name: str, annotation: Any) -> Union[Par
 
     io.name = io.name or python_name
     if isinstance(io, Parameter):
-        add_metadata_from_type(io, annotation)
+        set_enum_based_on_type(io, annotation)
 
     return io
 
--- a/src/hera/shared/_type_util.py
+++ b/src/hera/shared/_type_util.py
@@ -6,6 +6,7 @@ from typing import (
     Any,
     Iterable,
     List,
+    Literal,
     Optional,
     Tuple,
     Type,
@@ -120,6 +121,8 @@ def origin_type_issubtype(annotation: Any, type_: Union[type, Tuple[type, ...]])
     origin_type = get_unsubscripted_type(unwrapped_type)
     if origin_type is Union or origin_type is UnionType:
         return all(origin_type_issubtype(arg, type_) for arg in get_args(unwrapped_type))
+    if origin_type is Literal:
+        return all(isinstance(value, type_) for value in get_args(unwrapped_type))
     return isinstance(origin_type, type) and issubclass(origin_type, type_)
 
 
--- a/src/hera/shared/_type_util.py
+++ b/src/hera/shared/_type_util.py
@@ -83,6 +83,13 @@ def get_workflow_annotation(annotation: Any) -> Optional[Union[Artifact, Paramet
     return metadata[0]
 
 
+def add_metadata_from_type(parameter: Parameter, annotation: Any) -> None:
+    if not parameter.enum:
+        type_ = unwrap_annotation(annotation)
+        if get_origin(type_) is Literal:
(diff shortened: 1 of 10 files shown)


src/hera/workflows/script.py
```diff
@@ -379,6 +380,7 @@ def _get_parameters_from_callable(source: Callable) -> List[Parameter]:
             default = MISSING
 
         param = Parameter(name=p.name, default=default)
+        add_metadata_from_type(param, p.annotation)
```
alicederyn, line 383: I wasn't sure if we wanted to populate param.enum from the Literal metadata in the non-experimental path, or just when the `script_annotations` experimental feature is enabled, so I put this change into a separate commit to make it easy to revert.

sambhav: I would prefer to do it just for script annotations for now

alicederyn: That does mean that the generated YAML for a non-Annotated script will change depending on whether an unrelated experimental feature is enabled though.

src/hera/shared/_type_util.py
```diff
@@ -82,6 +83,13 @@ def get_workflow_annotation(annotation: Any) -> Optional[Union[Artifact, Paramet
     return metadata[0]
 
 
+def add_metadata_from_type(parameter: Parameter, annotation: Any) -> None:
```
jeongukjae, line 86: should be great if docstring is added here

src/hera/shared/_type_util.py
```diff
@@ -91,13 +99,19 @@ def construct_io_from_annotation(python_name: str, annotation: Any) -> Union[Par
...
-        return annotation_copy
+        io = workflow_annotation.copy()
+    else:
+        io = Parameter()
+
+    io.name = io.name or python_name
+    if isinstance(io, Parameter) and not io.enum:
+        type_ = unwrap_annotation(annotation)
+        if get_origin(type_) is Literal:
+            io.enum = list(get_args(type_))
```
jeongukjae, line 112: seems we can reuse above function

```suggestion
    if isinstance(io, Parameter):
        add_metadata_from_type(io, annotation)
```

src/hera/shared/_type_util.py
jeongukjae: are there any other things you want to do in this function? this function name looks generic.

alicederyn: That was the intention, but looking at the Parameter type, I can't see any other fields that we would want to set based on the type annotation, so renamed to be more specific -- thanks!

src/hera/shared/_type_util.py
alicederyn: Whoops! This was indeed the intention, must have lost it in my git history.... Fixed, thanks!

jeongukjae (approved):
:+1:

sambhav (approved):

PR #1264, opened 2024-11-13
Add a local build-docs make target

elliotgunton:
* Make it easier to check doc changes locally before making a PR
* Also removed deprecated options in mkdocs.yml


Makefile
```diff
@@ -124,6 +124,11 @@ init-files:
 examples:  ## Generate documentation files for examples
 	@(cd docs && poetry run python generate.py)
 
+.PHONY: build-docs
+build-docs: ## Generate (and host) documentation locally
+	@python -m pip install --exists-action=w --no-cache-dir -r docs/requirements.txt
```
alicederyn, line 129: Would it be worth making a separate venv for this so we don't accidentally disrupt the Poetry one?

elliotgunton: Not worth the effort for now as it's not supported natively by poetry https://stackoverflow.com/questions/70739858/how-to-create-a-brand-new-virtual-environment-or-duplicate-an-existing-one-in-po

sambhav (approved):

PR #1278, opened 2024-11-27
Remove override for `default` usage in `Parameter` annotations

### LINKED ISSUE #812: Input Parameter annotations should only be able to set default via standard Python

elliotgunton:
**Pull Request Checklist**
- [x] Fixes #812 (final step)
- [x] Tests adjusted
- [x] Documentation/examples added previously

**Description of PR**
Currently, when using the default attribute in Parameters (`Parameter(default="x")`) users can supress the error with the `suppress_parameter_default_error` experimental feature flag. If users have not updated their code (which is a very easy change) they will now get errors when upgrading Hera, the warning has been there for 2 minor versions and this is an experimental feature - we will be able to graduate it in the next version once this is released.

### DIFF
--- a/tests/test_script_annotations.py
+++ b/tests/test_script_annotations.py
@@ -7,6 +7,7 @@ import pytest
 
 from hera.shared._pydantic import _PYDANTIC_VERSION
 from hera.workflows import Workflow, script
+from hera.workflows.io import Input
 from hera.workflows.parameter import Parameter
 from hera.workflows.steps import Steps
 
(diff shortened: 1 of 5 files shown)


tests/test_script_annotations.py
```diff
@@ -103,11 +80,7 @@ def echo_int(an_int: Annotated[int, Parameter(default=1)]):
 
         w.to_dict()
 
-    assert (
-        "default cannot be set via the Parameter's default, use a Python default value instead"
-        "You can suppress this error by setting "
-        'global_config.experimental_features["suppress_parameter_default_error"] = True'
-    ) in str(e.value)
+    assert ("default cannot be set via the Parameter's default, use a Python default value instead") in str(e.value)
```
alicederyn, line 83: Note: you can do this directly in `pytest.raises` with the `match=` flag.

alicederyn (approved):

sambhav (approved):

PR #1285, opened 2024-12-04
Fix artifact_repository_ref example and regenerate

elliotgunton:
Examples are out of date


examples/workflows/upstream/artifact_repository_ref.py
```diff
@@ -6,7 +6,7 @@
 )
 
 with Workflow(
-    generate_name="artifactory-repository-ref-",
+    generate_name="artifact-repository-ref-",
```
elliotgunton, line 9: This is the only manual change

sambhav (approved):

PR #1324, opened 2025-03-01
Add validation on assignment

### LINKED ISSUE #1169: Arguments do not build correctly when using a dict

- [x] No

- [x] exporting to YAML
- [x] submitting to Argo
- [ ] other: 

<h2>Bug report</h2>

**Describe the bug**

TODO - basically using `w.arguments = {"param_name": "value"}` before doing `w.create()` doesn't get the argument correctly because we don't check if `self.arguments` is itself a dict here https://github.com/argoproj-labs/hera/blob/523a9b91d5af34a6f797907d5d02ac4a89240d6a/src/hera/workflows/_mixins.py#L562

_Error log if applicable_:
```
error: something broke!
```

**To Reproduce**
```py

    # Add your code here
```

**Expected behavior**

**Environment**
- Hera Version: 5.X.X
- Python Version: 3.X.X
- Argo Version: 3.X.X

**Additional context**

elliotgunton:
**Pull Request Checklist**
- [x] Fixes #1169 
- [ ] Tests added

**Description of PR**
Currently, any assignment to fields of a Pydantic model is not revalidated. This causes problems in e.g. `arguments` as the `normalize_to_list_or(ModelArguments)` doesn't fire, so we have a dictionary instead of a list of one (of the dict).

### DIFF
--- a/src/hera/shared/_pydantic.py
+++ b/src/hera/shared/_pydantic.py
@@ -92,6 +92,9 @@ class BaseModel(PydanticBaseModel):
         smart_union = True
         """uses smart union for matching a field's specified value to the underlying type that's part of a union"""
 
+        validate_assignment = True
+        """validates on assignment to a field"""
+
 
 __all__ = [
     "BaseModel",


elliotgunton:
We build up the model objects in different places, and `validate_assigment` re-validates the whole model, not just the field being assigned so it isn't what we're looking for.

PR #1337, opened 2025-03-05
Fix edge case for model parameter conversion

elliotgunton:
**Description of PR**
Noticed the tests were incorrect while checking related logic on #1336. `ModelInputs`, `ModelOutputs` and `ModelArguments` accepts Hera `Parameter`s as they are a subclass of the model `Parameter`. We can convert to the parent class easily using `parse_obj`. If anyone is mixing Hera's Parameter and ModelInputs (or the others), they will get invalidate YAML dumped that Argo will not accept downstream (as Hera's `Parameter` has the `output` field):

```yml
  arguments:
    parameters:
    - name: test
      output: false  # should not be here!
      value: value
```
- Added logic to convert `Parameter` instances to `ModelParameter` in `IOMixin._build_inputs` and `IOMixin._build_outputs`, and in `ArgumentsMixin._build_arguments`
- Updated unit tests to cover the new conversion logic.
- Renamed `test_io_mixins.py` to `test_pydantic_io_mixins.py` for clarity.


src/hera/workflows/_mixins.py
```diff
@@ -285,6 +285,13 @@ def _build_inputs(self) -> Optional[ModelInputs]:
         if self.inputs is None:
             return None
         elif isinstance(self.inputs, ModelInputs):
+            # Special case as Parameter is a subclass of ModelParameter
+            # We need to convert Parameters to ModelParameters
+            if self.inputs.parameters:
+                self.inputs.parameters = [
+                    ModelParameter.parse_obj(p.dict()) if isinstance(p, Parameter) else p
+                    for p in self.inputs.parameters
+                ]
```
alicederyn, line 294: Worth extracting this repeated code into a separate method?

alicederyn (approved):

src/hera/workflows/_mixins.py
elliotgunton: Yep makes sense

Issue #1166, opened 2024-08-21
Custom serialization for non-user types and non-serializable types for Hera runner (Parameter/Artifact inputs and outputs)


elliotgunton:
# Proposal for custom serialisation and deserialisation

A largely common-sense design: we add the `loader` and `dumper` fields to `Parameter` and `Artifact`, allowing `loader` to be a `Callable` for the existing `loader: Optional[ArtifactLoader] = None` in `artifact.py`.

The attributes will be:

```py
    dumper: Optional[Callable[[Any], str]] = None
    """used to specify a dumper function to serialise the parameter value for Annotated parameters"""

    loader: Optional[Callable[[str], Any]] = None
    """used to specify a loader function to load the parameter value for Annotated parameters"""
```

Then, I've split the problem into two main cases to deal with (which are duplicated for Artifacts, but are identical to Parameters from the user perspective):

1. The value to be (de)serialised is a plain data type or `BaseModel` - this is already possible without any further implementation (tests exist).
2. The value to be (de)serialised is a custom class that the user may not control, e.g. a `pandas.DataFrame` from the original issue motivation. The (de)serialisation function should be passed in the annotation to the Parameter/Artifact's `loader`/`dumper`.

Then the third "case" worth pointing out is if the script function is using the Pydantic IO feature (so a class inheriting from `Input` or `Output` from `hera.workflows.io`), any attributes needing custom dumpers/loaders should have their (de)serialisation function passed in the annotation of the attributes within the class, not in the annotation of the function. The BaseModel will then need to add `allow_arbitrary_types = True` in the model config.

The examples below should demonstrate the intention from a user perspective.

<details><summary>Preamble code (imports and custom classes)</summary>

```py
import json

from pydantic import BaseModel

from hera.shared import global_config
from hera.workflows import Parameter, script
from hera.workflows.io.v1 import Input, Output

try:
    from typing import Annotated
except ImportError:
    from typing_extensions import Annotated
try:
    from pydantic.v1 import BaseModel
except ImportError:
    from pydantic import BaseModel

global_config.experimental_features["script_pydantic_io"] = True


# Classes used in examples:
class MyBaseModel(BaseModel):
    a: str = "a"
    b: str = "b"


class NonUserNonBaseModelClass:
    """Represents a non-user-defined class (e.g. pandas DataFrame) that does not inherit from BaseModel."""

    def __init__(self, a: str, b: str):
        self.a = a
        self.b = b

    @classmethod
[...]

PR #1339, opened 2025-03-07
Refactor runner utils for clarity

elliotgunton:
* Among other changes, rename _map_argo_inputs_to_function to _map_function_annotations and move the function to `util.py` from `script_annotations_util.py`
* Prepares us for adding `Parameter` loader functionality (#1166)

### DIFF
--- a/src/hera/workflows/_runner/script_annotations_util.py
+++ b/src/hera/workflows/_runner/script_annotations_util.py
@@ -21,18 +21,12 @@ from hera.shared._type_util import (
 from hera.shared.serialization import serialize
 from hera.workflows import Artifact, Parameter
 from hera.workflows.artifact import ArtifactLoader
-from hera.workflows.io.v1 import (
-    Output as OutputV1,
-)
+from hera.workflows.io.v1 import Output as OutputV1
 
 try:
-    from hera.workflows.io.v2 import (  # type: ignore
-        Output as OutputV2,
-    )
+    from hera.workflows.io.v2 import Output as OutputV2
 except ImportError:
-    from hera.workflows.io.v1 import (  # type: ignore
-        Output as OutputV2,
-    )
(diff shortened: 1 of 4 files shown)


alicederyn:
Tests are failing; will wait until that's addressed to review.

src/hera/workflows/_runner/script_annotations_util.py
```diff
@@ -16,26 +15,22 @@
...
-    is_subscripted,
     origin_type_issubtype,
     unwrap_annotation,
 )
 from hera.shared.serialization import serialize
 from hera.workflows import Artifact, Parameter
 from hera.workflows.artifact import ArtifactLoader
 from hera.workflows.io.v1 import (
-    Input as InputV1,
     Output as OutputV1,
```
alicederyn, line 25: nit: would you be averse to removing the trailing comma so this reverts to a single import? If so, we can set `split-on-trailing-comma = False` [in the isort settings](https://docs.astral.sh/ruff/settings/#lint_isort_split-on-trailing-comma)

alicederyn (approved):

src/hera/workflows/_runner/script_annotations_util.py
elliotgunton: I can fix these to be on one line, but I'd prefer not changing the setting – tried it and got this kind of diff:

```diff
--- a/examples/workflows/misc/template_level_volume.py
+++ b/examples/workflows/misc/template_level_volume.py
@@ -2,15 +2,7 @@
 
 See https://argo-workflows.readthedocs.io/en/latest/walk-through/volumes/"""
 
-from hera.workflows import (
-    Container,
-    ExistingVolume,
-    Parameter,
-    Resource,
-    Step,
-    Steps,
-    Workflow,
-)
+from hera.workflows import Container, ExistingVolume, Parameter, Resource, Step, Steps, Workflow
```

Which is uhhhh... not nice :)

PR #1351, opened 2025-03-28
Add a recursive dag example

elliotgunton:
* Prompted by discussion https://github.com/argoproj-labs/hera/discussions/1349

### DIFF
--- a/docs/examples/workflows/dags/recursive_dag.md
+++ b/docs/examples/workflows/dags/recursive_dag.md
@@ -1,0 +1,63 @@
+# Recursive Dag
+
+
+
+
+
+
+=== "Hera"
+
+    ```python linenums="1"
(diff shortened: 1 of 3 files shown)


alicederyn (approved):
How would we do this with the decorator-based approach?

elliotgunton:
Decorator syntax doesn't seem possible without some work. It would be

```py
@w.dag()
def recursive_dag():
    roll_task = roll()
    recursive_dag(when=f"{roll_task.result} != 6")
```
But when we collect the dag tasks by running the function, the call to itself doesn't seem to find the function
```
    @w.dag()
    def recursive_dag():
        roll_task = roll()
>       recursive_dag(when=f"{roll_task.result} != 6")
E       NameError: name 'recursive_dag' is not defined
```
🤷

PR #1363, opened 2025-04-11
Update deps and examples

elliotgunton:
**Description of PR**
Currently, there are a load of deps to be updated, including mypy which raised a complaint about the missing `**kwargs` in the new decorators (which is by design). I wanted to avoid polluting the git history with all the dependabot PRs so have closed them.

This PR updates deps, regenerates docs, fixes test/linting complaints.


elliotgunton:
We can ignore the codecov CI failure, I've only added `# type: ignore` to the file it's complaining about

alicederyn (approved):

PR #1367, opened 2025-04-15
Add custom Artifact serialisation

### LINKED ISSUE #1191: Cannot output a non-serialisable Artifact using Pydantic IO
A complete blocker to using the new decorators - I have no way to output a bytes Artifact from a template - 

Using
```py
class ModelTrainingInput(Input):
    X_train: Annotated[list, Artifact(name="X_train", loader=ArtifactLoader.json)]
    y_train: Annotated[dict, Artifact(name="y_train", loader=ArtifactLoader.json)]
    model: Annotated[Path, Artifact(name="model", output=True)]  # Note the `output=True`
```

Gets the following error when building the workflow

```
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/Users/elliot/projects/ds-blog/ds_blog/__main__.py", line 14, in <module>
    from ds_blog.workflow import w
  File "/Users/elliot/projects/ds-blog/ds_blog/workflow.py", line 138, in <module>
    @w.dag()
     ^^^^^^^
  File "/Users/elliot/Library/Caches/pypoetry/virtualenvs/ds-blog-IWZkzs9u-py3.12/lib/python3.12/site-packages/hera/workflows/_meta_mixins.py", line 826, in decorator
    func_return = func(input_obj)
                  ^^^^^^^^^^^^^^^
  File "/Users/elliot/projects/ds-blog/ds_blog/workflow.py", line 144, in run_training
    model_training(
  File "/Users/elliot/Library/Caches/pypoetry/virtualenvs/ds-blog-IWZkzs9u-py3.12/lib/python3.12/site-packages/hera/workflows/_meta_mixins.py", line 670, in script_call_wrapper
    return self._create_subnode(subnode_name, func, script_template, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/elliot/Library/Caches/pypoetry/virtualenvs/ds-blog-IWZkzs9u-py3.12/lib/python3.12/site-packages/hera/workflows/_meta_mixins.py", line 550, in _create_subnode
    subnode_args = args[0]._get_as_arguments()
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/elliot/Library/Caches/pypoetry/virtualenvs/ds-blog-IWZkzs9u-py3.12/lib/python3.12/site-packages/hera/workflows/io/_io_mixins.py", line 152, in _get_as_arguments
    templated_value = serialize(self_dict[field])
                                ~~~~~~~~~^^^^^^^
KeyError: 'model'
```

And using

```py
class ModelTrainingOutput(Output):
    model: Annotated[bytes, Artifact(name="model", archive=NoneArchiveStrategy())]

@w.script()
def model_training(model_training_input: ModelTrainingInput) -> ModelTrainingOutput:
    X_train = np.array(model_training_input.X_train)
    y_train = pd.Series(model_training_input.y_train)
    model = LogisticRegression(random_state=42)
    model.fit(X_train, y_train)
    return ModelTrainingOutput(model=pickle.dumps(model))
```

gets the following error when running on the cluster
```
  File "/usr/local/lib/python3.12/site-packages/hera/workflows/_runner/util.py", line 222, in _runner
    output = _save_annotated_return_outputs(function(**kwargs), output_annotations)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/hera/workflows/_runner/script_annotations_util.py", line 250, in _save_annotated_return_outputs
    _write_to_path(path, value)
  File "/usr/local/lib/python3.12/site-packages/hera/workflows/_runner/script_annotations_util.py", line 326, in _write_to_path
    output_string = serialize(output_value)
                    ^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/hera/shared/serialization.py", line 56, in serialize
    return json.dumps(value, cls=PydanticEncoder)  # None serialized as `null`
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/json/__init__.py", line 238, in dumps
    **kw).encode(obj)
          ^^^^^^^^^^^
  File "/usr/local/lib/python3.12/json/encoder.py", line 200, in encode
    chunks = self.iterencode(o, _one_shot=True)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/json/encoder.py", line 258, in iterencode
    return _iterencode(o, 0)
           ^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/hera/shared/serialization.py", line 42, in default
    return super().default(o)
           ^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/json/encoder.py", line 180, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
TypeError: Object of type bytes is not JSON serializable
```

Workaround is to use the old syntax with an "output" artifact in the function inputs i.e.

```py
@script(constructor="runner")
def model_training(
    X_train: Annotated[list, Artifact(name="X_train", loader=ArtifactLoader.json)],
    y_train: Annotated[dict, Artifact(name="y_train", loader=ArtifactLoader.json)],
    model_path: Annotated[Path, Artifact(name="model", archive=NoneArchiveStrategy(), output=True)],
):
```

And doing
```py
    model_path.write_bytes(pickle.dumps(model))
```

_Originally posted by @elliotgunton in https://github.com/argoproj-labs/hera/issues/1166#issuecomment-2305085253_

elliotgunton:
Separated artifact implementation for #1166 
Also fixes #1191


src/hera/workflows/_runner/script_annotations_util.py
```diff
@@ -95,40 +95,54 @@ def get_annotated_output_param(param_annotation: Parameter) -> Path:
...
+    path.parent.mkdir(parents=True, exist_ok=True)
+    return path
+
+
+def get_annotated_artifact_value(artifact_annotation: Artifact, func_param_annotation: type) -> Union[Path, Any]:
     """Get the value of the given Artifact annotation.
 
     If the artifact is an output, return the path it will write to.
-    If the artifact is an input, return the loaded value (json, path or string) using its ArtifactLoader.
+    If the artifact is an input, return the  loaded value (json obj, path or string)
```
alicederyn, line 112:
```suggestion
    If the artifact is an input, return the loaded value (json obj, path or string)
```

src/hera/workflows/_runner/script_annotations_util.py
```diff
@@ -95,40 +95,54 @@ def get_annotated_output_param(param_annotation: Parameter) -> Path:
...
+    using its ArtifactLoader, unless ArtifactLoader is a Callable, in which case it
+    is used to load the value.
```
alicederyn, line 114: I don't understand from this text what is different between the if and the else -- they both seem to say it will be loaded using the ArtifactLoader

src/hera/workflows/_runner/script_annotations_util.py
```diff
@@ -95,40 +95,54 @@ def get_annotated_output_param(param_annotation: Parameter) -> Path:
...
-        path = Path(artifact_annotation.path)
-        with path.open() as f:
-            return json.load(f)
+    actual_type = unwrap_annotation(func_param_annotation)
 
-    if artifact_annotation.loader == ArtifactLoader.file.value:
-        path = Path(artifact_annotation.path)
+    if artifact_annotation.loader is None:
+        if issubclass(actual_type, Path):
+            return artifact_annotation.path
```
alicederyn, line 131: nit, maybe use `and` here since there's nothing else in the outer if?
```suggestion
    if artifact_annotation.loader is None and issubclass(actual_type, Path):
        return artifact_annotation.path
```
(in passing, I'm not sure why we're doing `issubclass` here &mdash; if it's not a Path, we're unlikely to be passing whatever subtype the user specified)

src/hera/workflows/_runner/script_annotations_util.py
```diff
@@ -95,40 +95,54 @@ def get_annotated_output_param(param_annotation: Parameter) -> Path:
...
+
+    path = Path(artifact_annotation.path)
+    if artifact_annotation.loader == ArtifactLoader.file:
         return path.read_text()
 
-    if artifact_annotation.loader is None:
-        return artifact_annotation.path
+    if artifact_annotation.loader == ArtifactLoader.json or artifact_annotation.loader is None:
+        with path.open() as f:
+            return json.load(f)
```
alicederyn, line 139: This is a change of behaviour for when loader is `None`, correct? We never previously decoded it. Won't that risk breaking people?

src/hera/workflows/_runner/util.py
```diff
@@ -190,8 +190,10 @@ def _map_function_annotations(function: Callable, template_inputs: Dict[str, str
                     function_kwargs[func_param_name] = load_param_input(
                         param_value, func_param.annotation, param_or_artifact.loader
                     )
-            else:
-                function_kwargs[func_param_name] = get_annotated_artifact_value(param_or_artifact)
+            elif isinstance(param_or_artifact, Artifact):
```
alicederyn, line 193: nit: why did this change? We already know it's an Artifact here based on what `get_workflow_annotation` can return -- if we want to defend against the case it might not be, it would be clearer to make this an `assert(isinstance(param_or_artifact, Artifact))` in an unconditional `else`; failing that, there should be an else/raise branch.

src/hera/workflows/artifact.py
```diff
@@ -84,10 +84,16 @@ class Artifact(BaseModel):
...
 
-    loader: Optional[ArtifactLoader] = None
+    loader: Union[Optional[ArtifactLoader], Optional[Callable[[str], Any]]] = None
     """used for input Artifact annotations for determining how to load the data.
 
-    Note: A loader value of 'None' must be used with an underlying type of 'str' or Path-like class."""
+    Use a callable to specify a dumper function to serialise the Artifact value for Annotated Artifacts.
+    Otherwise, use an ArtifactLoader to automatically load the Artifact value (deserialising the JSON string
+    or loading the file contents as a string). A loader value of 'None' must be used with an underlying type
+    of 'str' or Path-like class to load the Artifact as a Path."""
```
alicederyn, line 93: This comment doesn't seem to match the code -- we don't necessarily pass in a `str` any more.

src/hera/workflows/_runner/script_annotations_util.py
elliotgunton: This would only be breaking if someone is using a `str` type input with a loader of `None`, so they want the Path but as a `str`, which I would consider a rare use case. I was considering if/how to use experimental flags for this feature, which I think make sense as a way to deprecate the old `ArtifactLoader` enum, or at least its `None` behaviour, so we make it require a `Path`, or add a new `ArtifactLoader.Path`. The main loader/dumper feature would not need an experimental flag.

src/hera/workflows/_runner/script_annotations_util.py
```diff
@@ -43,7 +43,7 @@ def _get_outputs_path(destination: Union[Parameter, Artifact]) -> Path:
 
 
 def _get_dumper_function(destination: Union[Parameter, Artifact]) -> Callable:
-    if isinstance(destination, Parameter) and destination.dumper is not None:
+    if isinstance(destination, (Artifact, Parameter)) and destination.dumper is not None:
```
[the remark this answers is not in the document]
elliotgunton: True, was a leftover from trying to do the Parameter/Artifact implementation in separate PRs.

src/hera/workflows/_runner/util.py
elliotgunton: No functional reason for this change, it was mainly for readability while I was working on the feature. I've changed it back for now as I think I managed to avoid too much indentation through some refactoring.

src/hera/workflows/_runner/script_annotations_util.py
elliotgunton: This was in anticipation of adding the override flag

src/hera/shared/_global_config.py
```diff
@@ -201,13 +201,15 @@ def _set_defaults(cls, values):
...
 _DECORATOR_SYNTAX_FLAG = "decorator_syntax"
+_ARTIFACT_PATH_AS_STRING_FLAG = "artifact_path_as_string"
 
 # A dictionary where each key is a flag that has a list of flags which supersede it, hence
 # the given flag key can also be switched on by any of the flags in the list. Using simple flat lists
 # for now, otherwise with many superseding flags we may want to have a recursive structure.
 _SUPERSEDING_FLAGS: Dict[str, List] = {
     _SCRIPT_PYDANTIC_IO_FLAG: [_DECORATOR_SYNTAX_FLAG],
     _DECORATOR_SYNTAX_FLAG: [],
+    _ARTIFACT_PATH_AS_STRING_FLAG: [],
```
alicederyn, line 212: would it be simpler to do `_SUPERSEDING_FLAGS.get(flag, [])` so we don't need to add empty lists here?

src/hera/workflows/_runner/script_annotations_util.py
```diff
@@ -99,30 +107,37 @@ def get_annotated_artifact_value(artifact_annotation: Artifact) -> Union[Path, A
...
+    if artifact_annotation.loader is None:
+        if os.environ.get("hera__artifact_path_as_string", None) is not None and issubclass(actual_type, str):
+            warnings.warn(
+                "The artifact path is being passed as a string. This is deprecated and will be removed in future versions. "
+                "Please use a Path type for the function parameter.",
+            )
+            return artifact_annotation.path
+
+        if issubclass(actual_type, Path):
+            return artifact_annotation.path
```
alicederyn, line 128: 1. Don't we still need to ensure this is a `Path` instance before returning it?
2. If this is a *subclass* of `Path`, we need to ensure the object we actually have is the right type.
```suggestion
        if issubclass(actual_type, Path):
            path = Path(artifact_annotation.path)
            if not isinstance(path, actual_type):
              raise RuntimeError(f"Artifact parameter annotated with {actual_type.__name__}, but Paths are {type(path).__name__} instances on this OS")
            return path
```

src/hera/workflows/_runner/script_annotations_util.py
```diff
@@ -99,30 +107,37 @@ def get_annotated_artifact_value(artifact_annotation: Artifact) -> Union[Path, A
...
 
-    if artifact_annotation.loader is None:
-        return artifact_annotation.path
+    if artifact_annotation.loader == ArtifactLoader.json or artifact_annotation.loader is None:
+        with path.open() as f:
+            return json.load(f)
+
+    if isinstance(artifact_annotation.loader, Callable):  # type: ignore
+        loaded_value = artifact_annotation.loader(path.read_text())
+        return loaded_value
```
alicederyn, line 140: What about codecs that want to work on `bytes`?

alicederyn: If we renamed the attributes `loads` and `dumps` then we could have `load` and `dump` that take Paths.

elliotgunton: Sounds good – we should use `loads` and `dumps` in Parameter as well then

src/hera/workflows/_runner/script_annotations_util.py
elliotgunton: Technically no, Pydantic will validate and convert a `str` path into the user's type when calling the function. Note, the old code worked this way (returning the `artifact_annotation.path` which is a `str`).

src/hera/workflows/_runner/script_annotations_util.py
```diff
@@ -46,6 +46,9 @@ def _get_dumper_function(destination: Union[Parameter, Artifact]) -> Callable:
     if isinstance(destination, Parameter) and destination.dumps is not None:
         return destination.dumps
 
+    if isinstance(destination, Artifact) and destination.dumps is not None:
+        return destination.dumps
+
     return serialize
```
alicederyn, line 52: I think we can simplify this?
```suggestion
    return destination.dumps or serialize
```

alicederyn (approved):

src/hera/workflows/_runner/script_annotations_util.py
elliotgunton: Just pushed the `dump` to bytes change which needed this code layout

src/hera/workflows/_runner/script_annotations_util.py
elliotgunton: I chose to use `bytes` rather than `Path` to mirror the `str` behaviour. Also then we have the user having to pick the `Path` to write to for the `dump` function.

elliotgunton:
Hey @alicederyn - sorry for the push after approving, I was intending to add the `bytes` de/serialisation to this PR. If you're happy with the changes in https://github.com/argoproj-labs/hera/pull/1367/commits/385bb15dfdae141880deaf02d274b62adda3539b then I can merge. The commit after that was tidying up and undoing the changes made from previous assumptions that we'd change the old `ArtifactLoader` behaviour.

src/hera/workflows/_runner/script_annotations_util.py
```diff
@@ -46,6 +46,12 @@ def _get_dumper_function(destination: Union[Parameter, Artifact]) -> Callable:
     if isinstance(destination, Parameter) and destination.dumps is not None:
         return destination.dumps
 
+    if isinstance(destination, Artifact):
+        if destination.dumps is not None:
+            return destination.dumps
+        if destination.dump is not None:
+            return destination.dump
```
alicederyn, line 53: I'm not comfortable with how this is discarding whether the expected input is str or not :/ It doesn't seem like the call site can have enough information to call this safely.

src/hera/workflows/_runner/script_annotations_util.py
```diff
@@ -87,38 +93,48 @@ def get_annotated_output_param(param_annotation: Parameter) -> Path:
...
         # Path is added to the spec automatically when built. As it isn't present in the annotation itself,
         # we need to add it back in for the runner.
         artifact_annotation.path = artifact_annotation._get_default_inputs_path()
 
+    path = Path(artifact_annotation.path)
+    if artifact_annotation.loads is not None and isinstance(artifact_annotation.loads, Callable):  # type: ignore
+        return artifact_annotation.loads(path.read_text())
+
+    if artifact_annotation.load is not None and isinstance(artifact_annotation.load, Callable):  # type: ignore
+        return artifact_annotation.load(path.read_bytes())
```
alicederyn, line 131: Any reason not to just pass the `path` directly, in case this is a very large file the user doesn't want loaded into memory? Similarly, can we pass the `path` into the dump function? I think this more closely matches e.g. `json.load` and `json.dump`.

alicederyn (changes requested):

src/hera/workflows/_runner/script_annotations_util.py
elliotgunton: Yes, a couple:
* I think we should keep things simple and consistent for the majority of use cases; the only choice for the user in this case is deserialising from `str` or `bytes`, plus easier handling for us as `read_text` and `read_bytes` exist
* If they want a path for a big-file-use-case (a very good point), I think it makes more sense that the function parameter would be a `Path` and the user creates and deals with a streamed file in the function itself
  * They can then test their file streaming business logic themselves
  * Simplifies our handling as there's only one way to do things (I am no expert in low level file things)
  * I think there is also potential confusion for us (and users) of whether we are dealing with a file pointer or a partially-streamed section of the file contents after `load`ing using the annotation function, depending on the type of the variable

src/hera/workflows/_runner/script_annotations_util.py
elliotgunton: Not sure if you mixed up input/output or dumping/loading itself there - if you meant output, then I don't _think_ it matters - we just call the `dumper` function passed to `_write_to_path` and see what it returned, then call `path.write_text` or `path.write_bytes`.

src/hera/workflows/_runner/script_annotations_util.py
alicederyn: Then perhaps call this `loadb` and `dumpb`

elliotgunton: I was having the same thought yesterday! 😆 But only didn't suggest it because I've never seen a `loadb`/`dumpb` in the wild

src/hera/workflows/_runner/script_annotations_util.py
alicederyn: Yep, I did.

src/hera/workflows/_runner/script_annotations_util.py
alicederyn: I think that's probably because you'd normally go straight to pipe if you were dealing with bytes?

elliotgunton: Updated the names

src/hera/workflows/_runner/script_annotations_util.py
```diff
@@ -46,6 +46,12 @@ def _get_dumper_function(destination: Union[Parameter, Artifact]) -> Callable:
     if isinstance(destination, Parameter) and destination.dumps is not None:
         return destination.dumps
 
+    if isinstance(destination, Artifact):
+        if destination.dumps is not None:
+            return destination.dumps
+        if destination.dumpb is not None:
+            return destination.dumpb
```
alicederyn, line 53: Note we can simplify this
```suggestion
    if destination.dumps is not None:
        return destination.dumps
    if isinstance(destination, Artifact) and destination.dumpb is not None:
        return destination.dumpb
```

src/hera/workflows/_runner/script_annotations_util.py
```diff
@@ -87,38 +93,48 @@ def get_annotated_output_param(param_annotation: Parameter) -> Path:
...
         # Path is added to the spec automatically when built. As it isn't present in the annotation itself,
         # we need to add it back in for the runner.
         artifact_annotation.path = artifact_annotation._get_default_inputs_path()
 
+    path = Path(artifact_annotation.path)
+    if artifact_annotation.loads is not None and isinstance(artifact_annotation.loads, Callable):  # type: ignore
+        return artifact_annotation.loads(path.read_text())
+
+    if artifact_annotation.loadb is not None and isinstance(artifact_annotation.loadb, Callable):  # type: ignore
+        return artifact_annotation.loadb(path.read_bytes())
```
alicederyn, line 131: It looks like we only support callables for loads/loadb, and we have no alternate path for these, so I think we can simplify this? It'll raise a TypeError if the user provided a non-Callable, which is what we want.

```suggestion
    if artifact_annotation.loads is not None:
        return artifact_annotation.loads(path.read_text())

    if artifact_annotation.loadb is not None:
        return artifact_annotation.loadb(path.read_bytes())
```

src/hera/workflows/artifact.py
```diff
@@ -89,6 +89,18 @@ class Artifact(BaseModel):
...
     Note: A loader value of 'None' must be used with an underlying type of 'str' or Path-like class."""
 
+    loadb: Union[Optional[Callable[[bytes], Any]]] = None
+    """used to specify a loader function to deserialise from bytes for Annotated Artifact function parameters"""
+
+    dumpb: Optional[Callable[[Any], bytes]] = None
+    """used to specify a dumper function to serialise the Artifact value as bytes for Annotated Artifact function parameters"""
+
+    loads: Union[Optional[Callable[[str], Any]]] = None
+    """used to specify a loader function to deserialise a string representation of an object for Annotated Artifact function parameters"""
```
alicederyn, line 99: These Unions only have one element.

```suggestion
    loadb: Optional[Callable[[bytes], Any]] = None
    """used to specify a loader function to deserialise from bytes for Annotated Artifact function parameters"""

    dumpb: Optional[Callable[[Any], bytes]] = None
    """used to specify a dumper function to serialise the Artifact value as bytes for Annotated Artifact function parameters"""

    loads: Optional[Callable[[str], Any]] = None
    """used to specify a loader function to deserialise a string representation of an object for Annotated Artifact function parameters"""
```

elliotgunton:
Thank you @alicederyn - great suggestions to tidy up 🚀

PR #1379, opened 2025-05-07
Resolve deprecation warnings of regex library

emmanuel-ferdman:
# PR Summary
This small PR resolves the `re` deprecation warnings:
```python
DeprecationWarning: 'count' is passed as positional argument
```


elliotgunton (approved):
Thanks @emmanuel-ferdman - we only run this code on py3.9 to generate other files, so we don't see the deprecation warning usually

PR #1392, opened 2025-05-21
Use a `readiness_probe` on daemon example

elliotgunton:
* Exemplify best practice which is used in [the upstream docs](https://argo-workflows.readthedocs.io/en/latest/walk-through/daemon-containers/)

### DIFF
--- a/examples/workflows/misc/daemon.py
+++ b/examples/workflows/misc/daemon.py
@@ -1,12 +1,14 @@
-"""Enablement of https://argoproj.github.io/argo-workflows/variables/
-This example creates two tasks, one of the Tasks is a deamond task and its IP address is shared with the second task
-The daemoned task operates as server, serving an example payload, with the second task operating as a client, making
-http requests to the server."""
+"""Demonstrates usage of [daemon container IP](https://argoproj.github.io/argo-workflows/variables/).
+
+This example creates two tasks, one of the Tasks is a daemon task and its IP address is shared with the second task. The
+daemon task operates as server, serving an example payload, with the second task operating as a client, making http
+requests to the server."""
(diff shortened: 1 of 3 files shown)


examples/workflows/misc/daemon.py
```diff
@@ -1,12 +1,14 @@
-"""Enablement of https://argoproj.github.io/argo-workflows/variables/
-This example creates two tasks, one of the Tasks is a deamond task and its IP address is shared with the second task
+"""Demonstrates usage of [daemon container IP](https://argoproj.github.io/argo-workflows/variables/).
+
+This example creates two tasks, one of the Tasks is a deamond task and its IP address is shared with the second task.
```
alicederyn, line 3:
```suggestion
This example creates two tasks, one of the Tasks is a daemond task and its IP address is shared with the second task.
```

examples/workflows/misc/daemon.py
```diff
@@ -36,6 +38,6 @@ def consumer(ip):
 
 with Workflow(generate_name="daemon-", entrypoint="d") as w:
     with DAG(name="d"):
-        s = server()
+        s: Task = server()
```
alicederyn, line 41: why did you add this?

elliotgunton: So that intellisense knows `ip` exists on `s`. Otherwise:

image

alicederyn: I think this might be down to it selecting the "calling the original function" overload you're using when you call `server`, how does it look if you do `s = server(arguments={})` ?

alicederyn (approved):

PR #1440, opened 2025-07-16
Add YAML-to-Python converter to CLI

### LINKED ISSUE #711: Automatic Conversion of Argo YAML Files to Hera Code

elliotgunton:
**Pull Request Checklist**
- [x] Fixes #711 
- [x] Tests added

**Description of PR**
This PR adds a YAML converter to the Hera CLI. It includes a test suite which includes all `upstream` examples, and performs a roundtrip test on them, by creating an actual Python file containing the workflow code, then importing the workflow from it to convert it back to YAML and compare it to the original.

### DIFF
--- a/src/hera/_cli/generate/hera.py
+++ b/src/hera/_cli/generate/hera.py
@@ -1,19 +1,20 @@
 import os
 import sys
-from collections import ChainMap
 from dataclasses import (
     dataclass,
     field as dc_field,
 )
 from pathlib import Path
-from typing import Any, Generator, Iterable, List, Optional, Tuple, Type, Union, cast, get_args
+from typing import Any, Generator, Iterable, List, Optional, Tuple, Type, Union, cast
 
 import yaml
 
 from hera._cli.base import GeneratePython
 from hera._cli.generate.util import YAML_EXTENSIONS, expand_paths, filter_paths
 from hera.shared._pydantic import BaseModel
-from hera.shared._type_util import get_annotated_metadata, is_optionally_subtype, unwrap_annotation
+from hera.shared._type_util import (
+    get_annotated_metadata,
+)
 from hera.workflows._meta_mixins import ModelMapperMixin, _get_model_attr
 from hera.workflows.cluster_workflow_template import ClusterWorkflowTemplate
 from hera.workflows.container import Container
@@ -38,10 +39,10 @@ from hera.workflows.task import Task
 from hera.workflows.workflow import Workflow
 from hera.workflows.workflow_template import WorkflowTemplate
 
-try:
-    from inspect import get_annotations  # type: ignore
-except ImportError:
-    from hera.shared._inspect import get_annotations  # type: ignore
+if sys.version_info >= (3, 10):
+    from types import NoneType
+else:
+    NoneType = type(None)
 
 DEFAULT_EXTENSION = ".py"
 
@@ -121,19 +122,14 @@ def load_yaml_workflows(path: Path) -> Generator[ModelWorkflow, None, None]:
             raise ValueError(f"Invalid YAML workflow: {yaml_workflow}")
 
 
-def is_basic_type(annotation: type) -> bool:
-    """Check if the annotation is a basic JSON type."""
-    return is_optionally_subtype(annotation, (str, int, float, bool))
-
-
-def python_obj_to_repr(value: Any, annotation: Optional[type] = None) -> Tuple[str, List[str], List[str]]:
+def python_obj_to_repr(value: Any) -> Tuple[str, List[str], List[str]]:
     """Convert a JSON value to a Python string representation, without "None" values.
 
     This function also collects the assumed Hera/model imports and returns them.
     """
-    if annotation is not None and is_basic_type(annotation) or isinstance(value, (str, int, float, bool)):
+    if isinstance(value, (str, int, float, bool)):
         return repr(value), [], []
-    elif annotation is not None and is_optionally_subtype(annotation, list) or isinstance(value, list):
+    elif isinstance(value, list):
         python_repr = []
         model_imports = []
         hera_imports = []
@@ -143,7 +139,7 @@ def python_obj_to_repr(value: Any, annotation: Optional[type] = None) -> Tuple[s
             hera_imports.extend(sub_hera_imports)
             model_imports.extend(sub_imports)
         return "[" + ", ".join(python_repr) + "]", hera_imports, model_imports
-    elif annotation is not None and is_optionally_subtype(annotation, dict) or isinstance(value, dict):
+    elif isinstance(value, dict):
         python_repr = []
         model_imports = []
         hera_imports = []
@@ -153,10 +149,10 @@ def python_obj_to_repr(value: Any, annotation: Optional[type] = None) -> Tuple[s
             hera_imports.extend(sub_hera_imports)
             model_imports.extend(sub_imports)
         return "{" + ", ".join(python_repr) + "}", hera_imports, model_imports
-    elif annotation is not None and is_optionally_subtype(annotation, BaseModel) or isinstance(value, BaseModel):
+    elif isinstance(value, BaseModel):
         return model_to_python(value)
     else:
-        raise ValueError(f"Unsupported type: {annotation} for value {value}")
+        raise ValueError(f"Unsupported type: {type(value)} for value {value}")
 
 
 def convert_to_hera_template(
@@ -339,11 +335,11 @@ def model_to_python(model: BaseModel) -> Tuple[str, List[str], List[str]]:
     hera_imports = []
     model_imports = [model_name]
     class_def = [f"{model_name}("]
-    for attr in ChainMap(*(get_annotations(c) for c in model.__class__.__mro__)):
+    for attr in model.__fields__:
         if attr == "__slots__" or getattr(model, attr) is None:
             continue
         value = getattr(model, attr)
-        assign_value, sub_hera_imports, sub_model_imports = python_obj_to_repr(value, type(value))
+        assign_value, sub_hera_imports, sub_model_imports = python_obj_to_repr(value)
         hera_imports.extend(sub_hera_imports)
         model_imports.extend(sub_model_imports)
 
@@ -385,16 +381,12 @@ def build_file(
 
                 if attr == "templates":
                     for template in value:
-                        t_repr, sub_hera_imports, sub_model_imports = python_obj_to_repr(
-                            template, get_args(unwrap_annotation(annotation))[0]
-                        )
+                        t_repr, sub_hera_imports, sub_model_imports = python_obj_to_repr(template)
                         file_builder.hera_imports.extend(sub_hera_imports)
                         file_builder.model_imports.extend(sub_model_imports)
                         file_builder.context_def.append("    " + t_repr)
                 else:
-                    val_repr, sub_hera_imports, sub_model_imports = python_obj_to_repr(
-                        value, unwrap_annotation(annotation)
-                    )
+                    val_repr, sub_hera_imports, sub_model_imports = python_obj_to_repr(value)
                     file_builder.hera_imports.extend(sub_hera_imports)
                     file_builder.model_imports.extend(sub_model_imports)
                     file_builder.class_def.append(f"{attr}={val_repr},")
--- a/src/hera/shared/_type_util.py
+++ b/src/hera/shared/_type_util.py
@@ -19,11 +19,13 @@ from typing import (
 )
 
 if sys.version_info >= (3, 10):
-    from types import UnionType
+    from types import NoneType, UnionType
 else:
     UnionType = Union
+    NoneType = type(None)
 
 if TYPE_CHECKING:
+    # Avoid circular import
     from hera.workflows.artifact import Artifact
     from hera.workflows.parameter import Parameter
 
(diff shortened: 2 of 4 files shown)


src/hera/_cli/generate/hera.py
```diff
@@ -0,0 +1,434 @@
...
+            parallel_steps_list = parallel_steps.__root__
+            if len(parallel_steps_list) > 1:
+                hera_template_str.append(f"{indent}with invocator.parallel():")
+                indent = " " * 12
+            for step in parallel_steps.__root__:
+                val, sub_hera_imports, sub_model_imports = convert_to_hera_equivalent(step, Step)
+                hera_imports.extend(sub_hera_imports)
+                model_imports.extend(sub_model_imports)
+                hera_template_str.append(f"{indent}{val}")
+            indent = " " * 8
```
alicederyn, line 286: I'm worried this pattern is fragile. We're relying on `convert_to_hera_equivalent` only emitting a single function call, for instance, since we're only indenting the start of the block. I'd prefer to create a class encapsulating "adding code + imports" and "do an indented block", and pass a mutable instance into the various functions. Something more like:

```python
code_writer.add_import(hera_template_class)
code_writer.add_line("with invocator.parallel():")
with code_writer.indent(4) as with_block:
    write_hera_equivalent(block, task, Task)
```

I'm aware this is a significant rewrite, though, so up to you if/when you want to do it.

src/hera/_cli/generate/hera.py
```diff
@@ -0,0 +1,434 @@
...
+    model_name = model.__class__.__name__
+    if isinstance(model, Template):
+        return template_to_python(model)
+
+    hera_imports = []
+    model_imports = [model_name]
+    class_def = [f"{model_name}("]
+    for attr in ChainMap(*(get_annotations(c) for c in model.__class__.__mro__)):
+        if attr == "__slots__" or getattr(model, attr) is None:
+            continue
```
alicederyn, line 344: is there not a Pydantic API exposing this? what is the difference from `__fields__`?

src/hera/_cli/generate/hera.py
```diff
@@ -0,0 +1,434 @@
...
+    in that folder
+    """
+    paths = sorted(expand_paths(options.from_, YAML_EXTENSIONS, recursive=options.recursive))
+
+    # Generate a collection of source file paths and their resultant python.
+    path_to_output: list[tuple[str, str]] = []
+    for path in filter_paths(paths, includes=options.include, excludes=options.exclude):
+        python_outputs = []
+        for workflow in load_yaml_workflows(path):
+            python_outputs.append(workflow_to_python(workflow))
```
alicederyn, line 77: We probably want to add the filename and workflow name to any validation error we raise here, otherwise users are going to have a hard time knowing where errors came from.

src/hera/_cli/generate/hera.py
```diff
@@ -0,0 +1,434 @@
...
+        return convert_to_hera_invocator_template(template, DAG, template.dag)
+    elif template.steps is not None:
+        return convert_to_hera_invocator_template(template, Steps)
+    elif template.container_set is not None:
+        return convert_to_hera_invocator_template(template, ContainerSet, template.container_set)
+    else:
+        # No-op (convert Template to itself) and rearrange imports
+        val, hera_imports, model_imports = convert_to_hera_equivalent(template, Template)
+        hera_imports.remove("Template")
+        model_imports.append("Template")
```
alicederyn, line 330: nit: if we stored a single set of fully-qualified imports, I think the code might be simplified a bit.

src/hera/_cli/generate/hera.py
```diff
@@ -0,0 +1,434 @@
...
+        raise ValueError("Unrecognised model workflow class")
+
+    file_builder = FileBuilder(
+        hera_imports=[hera_workflow_class.__name__],
+        class_def=[f"with {hera_workflow_class.__name__}("],
+    )
+    file_builder = build_file(file_builder, hera_workflow_class, model)
+
+    imports = list(map(lambda x: f"from hera.workflows import {x}", set(file_builder.hera_imports)))
+    imports.extend(map(lambda x: f"from hera.workflows.models import {x}", set(file_builder.model_imports)))
```
alicederyn, line 427: nit: sort the imports?

src/hera/_cli/generate/util.py
```diff
@@ -0,0 +1,49 @@
+import os
+from fnmatch import fnmatch
+from pathlib import Path
+from typing import Generator, Iterable, List, Set
+
+YAML_EXTENSIONS = {".yml", ".yaml"}
+
+
+def expand_paths(source: Path, suffixes: Set[str], recursive: bool = False) -> Generator[Path, None, None]:
```
alicederyn, line 9: Can we simplify this with https://docs.python.org/3/library/pathlib.html#pathlib.Path.glob or https://docs.python.org/3/library/pathlib.html#pathlib.Path.walk ?

src/hera/shared/_type_util.py
```diff
@@ -70,11 +72,14 @@ def get_annotated_metadata(annotation, type_):
...
 
 
-def get_workflow_annotation(annotation: Any) -> Optional[Union[Artifact, Parameter]]:
+def get_workflow_annotation(annotation: Any) -> "Optional[Union[Artifact, Parameter]]":
     """If given annotation has Artifact or Parameter metadata, return it.
 
     Note that this function will raise the error when multiple Artifact or Parameter metadata are given.
     """
+    from hera.workflows.artifact import Artifact
+    from hera.workflows.parameter import Parameter
```
alicederyn, line 81: Why did we move the imports into a function?

src/hera/shared/_type_util.py
```diff
@@ -127,6 +134,22 @@ def get_unsubscripted_type(t: Any) -> Any:
...
+    """
+    unwrapped_type = unwrap_annotation(annotation)
+    origin_type = get_unsubscripted_type(unwrapped_type)
+    if origin_type is Union or origin_type is UnionType:
+        if len(get_args(unwrapped_type)) == 2 and type(None) in get_args(unwrapped_type):
+            return any(origin_type_issubtype(arg, type_) for arg in get_args(unwrapped_type))
+    else:
+        return isinstance(origin_type, type) and issubclass(origin_type, type_)
+
+    return False
```
alicederyn, line 150: Is this different from calling `origin_type_issubtype(annotation, (type_, type(None)))`?

src/hera/workflows/_mixins.py
```diff
@@ -449,7 +449,7 @@ class TemplateMixin(SubNodeMixin, HookMixin, MetricsMixin):
     name: Optional[str] = None
     node_selector: Optional[Dict[str, str]] = None
     parallelism: Optional[int] = None
-    http: Optional[HTTP] = None
+    http: Optional[HTTP] = None  # THIS SHOULD NOT BE HERE
```
alicederyn, line 452: can we be a bit more detailed on this comment?

tests/cli/test_generate_python.py
```diff
@@ -0,0 +1,127 @@
...
+    runner.invoke(str(yaml_file), "--to", str(output_path))
+
+    loader = SourceFileLoader("workflow_output", str(output_path))
+    workflow_module = ModuleType(loader.name)
+    loader.exec_module(workflow_module)
+    assert hasattr(workflow_module, "w") and isinstance(workflow_module.w, Workflow)
+    workflow = workflow_module.w
+
+    if isinstance(workflow, ClusterWorkflowTemplate):
+        workflow_from_yaml = cast(ClusterWorkflowTemplate, ClusterWorkflowTemplate.from_file(yaml_file))
```
alicederyn, line 69: why is this cast needed?

tests/test_remaining_examples.py
```diff
@@ -113,13 +113,13 @@ def test_upstream_examples_roundtrip(file_name):
         print(f"UNKNOWN KIND: {kind}")
         return
 
-    if kind == "Workflow":
-        kind_class = Workflow
-    elif kind == "WorkflowTemplate":
+    kind_class = Workflow
```
alicederyn, line 116: nit: assign this in an `else`?

src/hera/_cli/generate/hera.py
elliotgunton: This function was mostly a copy of `generate_yaml` from src/hera/_cli/generate/yaml.py, which I was intending to refactor with this change but wanted to get it working first. I would prefer to do that in a follow up minimal-functional-changes PR (but I agree telling the user what the problem was is a good idea, which is not done for YAML conversion).

src/hera/_cli/generate/hera.py
```diff
@@ -0,0 +1,434 @@
...
+    path_to_output: list[tuple[str, str]] = []
+    for path in filter_paths(paths, includes=options.include, excludes=options.exclude):
+        python_outputs = []
+        for workflow in load_yaml_workflows(path):
+            python_outputs.append(workflow_to_python(workflow))
+
+        if not python_outputs:
+            continue
+
+        path_to_output.append((path.name, join_workflows(python_outputs)))
```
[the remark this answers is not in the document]
elliotgunton: I believe it is preserved by line 98 below

src/hera/_cli/generate/hera.py
elliotgunton: Yes I agree it's not the best. I started adding the `FileBuilder` to the code below, but didn't want to refactor this bit 😅 I like the `with_block` though, I intended to do something similar.

src/hera/_cli/generate/hera.py
elliotgunton: I think this was a mindless copy-paste from the `ModelMapperMixin` 🤔 (And now I'm trying to remember why we had to use `__mro__` in the first place...).

Changed to `__fields__` and all tests still pass so 👍 from me

src/hera/shared/_type_util.py
elliotgunton: I think this was from the POC code - I had a circular import at some point. Reverted these changes and tests still working 🚀

src/hera/_cli/generate/hera.py
```diff
@@ -0,0 +1,434 @@
...
+    """Check if the annotation is a basic JSON type."""
+    return is_optionally_subtype(annotation, (str, int, float, bool))
+
+
+def python_obj_to_repr(value: Any, annotation: Optional[type] = None) -> Tuple[str, List[str], List[str]]:
+    """Convert a JSON value to a Python string representation, without "None" values.
+
+    This function also collects the assumed Hera/model imports and returns them.
+    """
+    if annotation is not None and is_basic_type(annotation) or isinstance(value, (str, int, float, bool)):
```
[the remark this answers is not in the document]
elliotgunton: All the `if`s have `or` so actually we can't do this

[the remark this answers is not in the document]
elliotgunton: Turns out we don't need the `annotation` at all 🤔

src/hera/shared/_type_util.py
elliotgunton: This function was unnecessary in the end so have removed. It is indeed equivalent to `origin_type_issubtype(annotation, (type_, type(None)))` though 👍

src/hera/workflows/_mixins.py
elliotgunton: Ah yes lol, forgot to add the reason why

tests/cli/test_generate_python.py
elliotgunton: `from_file` returns a `ModelMapperMixin`, which was avoiding a circular import. I don't intend to fix that in this PR.

elliotgunton:
@alicederyn would you be okay to approve this based on the conversion code logic itself (i.e. mainly looking at `src/hera/_cli/generate/hera.py`)? I will then do a refactor on the path handling for both `generate_yaml` and `generate python` as that will be a small behaviour change. Also hoping to refactor the logic in `src/hera/_cli/generate/hera.py` if I can find time for it. I think it's fine to merge as the CLI is an experimental feature that we can iterate on (as long as you agree the fundamentals in this PR are correct).

alicederyn (approved):

src/hera/_cli/generate/util.py
elliotgunton: Took a look at this but I think `os.walk` is the easiest to follow as `Path.walk` is 3.12+ only and trying to mix the `glob` with the recursive flag and suffixes is horrendous 🫣

PR #1442, opened 2025-07-22
Fix a mypy regression in IO mixins

alicederyn:
A recent change (probably #1432) caused mypy to start erroring on any subclass of Input/Output:

> Metaclass conflict: the metaclass of a derived class must be a (non-strict) subclass of the metaclasses of all its bases  [misc]

Fix this, and the preexisting issue with v2 Input/Output types, by moving the `__new__`/`__init__` logic required to disable validation in declaring context into two small, dedicated V1/V2 base classes, and making InputMixin and OutputMixin simple classes with no inherited metaclasses. This required some additional `cast` calls as the type checker no longer has visibility into these mixins being only used with Pydantic types.

Added in simple mypy tests to verify the dedicated Pydantic support is working as intended in both V1 and V2 models.

As a drive-by, I have also cleaned up some of our complex imports now we can rely on `pydantic.v1` always being available.

### DIFF
--- a/src/hera/workflows/io/_io_mixins.py
+++ b/src/hera/workflows/io/_io_mixins.py
@@ -1,5 +1,5 @@
 import sys
-from typing import TYPE_CHECKING, Iterator, List, Optional, Tuple, Type, Union
+from typing import TYPE_CHECKING, Iterator, List, Optional, Tuple, Type, Union, cast
 
 if sys.version_info >= (3, 11):
     from typing import Self
@@ -27,12 +27,9 @@ if TYPE_CHECKING:
     # We add BaseModel as a parent class of the mixins only when type checking which allows it
     # to be used with either a V1 BaseModel or a V2 BaseModel
     from hera.shared._pydantic import PydanticBaseModel as BaseModel
-else:
-    # Subclassing `object` when using the real code (i.e. not type-checking) is basically a no-op
-    BaseModel = object  # type: ignore
 
 
-def _construct_io_from_fields(cls: Type[BaseModel]) -> Iterator[Tuple[str, FieldInfo, Union[Parameter, Artifact]]]:
+def _construct_io_from_fields(cls: "Type[BaseModel]") -> Iterator[Tuple[str, FieldInfo, Union[Parameter, Artifact]]]:
     """Constructs a Parameter or Artifact object for all Pydantic fields based on their annotations.
 
     If a field has a Parameter or Artifact annotation, a copy will be returned, with missing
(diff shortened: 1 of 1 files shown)


src/hera/shared/serialization.py
```diff
@@ -31,14 +26,10 @@ def default(self, o: Any):
...
-                return o.dict(by_alias=True)
-        else:
-            if isinstance(o, BaseModel):
-                return o.model_dump(by_alias=True, mode="json")  # type: ignore
-            if isinstance(o, V1BaseModel):
-                return o.dict(by_alias=True)
+        if isinstance(o, V1BaseModel):
+            return o.dict(by_alias=True)
+        if isinstance(o, BaseModel):
+            return o.model_dump(by_alias=True, mode="json")  # type: ignore
```
elliotgunton, line 32: Not sure if it's to be trusted in this case, but I tried removing the `ignore` and `make lint` still passed
```suggestion
            return o.model_dump(by_alias=True, mode="json")
```

src/hera/shared/_pydantic.py
```diff
@@ -1,66 +1,49 @@
...
+    from inspect import get_annotations
+else:
+    from hera.shared._inspect import get_annotations
 
+from pydantic import VERSION
 
-_PYDANTIC_VERSION: int = int(VERSION.split(".")[0])
 # The pydantic v1 interface is used for both pydantic v1 and v2 in order to support
 # users across both versions.
+from pydantic.v1 import (
```
elliotgunton, line 16: Yes! 😍 🚀 I can feel the Pydantic v2 migration becoming more manageable 😂

src/hera/workflows/io/_io_mixins.py
```diff
@@ -50,31 +43,12 @@ def _construct_io_from_fields(cls: Type[BaseModel]) -> Iterator[Tuple[str, Field
...
-
-        super().__init__(**kwargs)
-
+class InputMixin:
     @classmethod
     def _get_parameters(cls, object_override: Optional[Self] = None) -> List[Parameter]:
         parameters = []
 
-        for field, field_info, param in _construct_io_from_fields(cls):
+        for field, field_info, param in _construct_io_from_fields(cls):  # type: ignore
```
elliotgunton, line 51: nit: I think I would slightly prefer `cast` over `ignore`s

```suggestion
        for field, field_info, param in _construct_io_from_fields(cast(type[BaseModel], cls)):
```

(Same for the other ignores)

elliotgunton (approved):

PR #1457, opened 2025-07-30
Add `set_next_defaults` function for `Task` and allow multiple `on` values

### LINKED ISSUE #726: Configurable rshift operator behaviour
The depends shorthand using the rshift operator `>>` uses the "enhanced depends logic" from https://argoproj.github.io/argo-workflows/enhanced-depends-logic/#depends. i.e. `task >> other_task` is equivalent to `task.Succeeded || task.Skipped || task.Daemoned`. We've found skipping tasks to be a common pattern, but as the default depends behaviour includes `skipped` tasks, you then have to also specify `when` clauses, or only use the `on_success` function, forgoing the helpful `>>` syntactic sugar.

A configurable behaviour for the `>>` operator. Sensible options would be the current default or only on success. This option could be part of the `global_config` e.g. `global_config.default_task_depends` taking an enum value.

Adding when clauses to all tasks after the skipped tasks or only using `on_success`. Not the cleanest solution.

**Additional context**
_None_

### LINKED ISSUE #1384: Setting DAG Task dependencies through `next` is too restrictive
Given the following Tasks:

```py
with DAG(name="dag"):
    A = Task(...)
    B = Task(...)
    C = Task(...)
```

I want to describe the dependencies:
* Run B if A Failed or Errored
* Run C if A Succeeded or Skipped

Which have to be described using strings in `depends`:

```py
B.depends = "A.Failed || A.Errored"
C.depends = "A.Succeeded"
```

Because trying to use `next` like so (which is valid syntactically):

```py
        A.next(B, on=TaskResult.failed)
        A.next(B, operator=Operator.or_, on=TaskResult.errored)
        A.next(C, on=TaskResult.succeeded)
```

results in:

```
>           raise ValueError(f"{self.name} already in {other.name}'s depends: {other.depends}")
E           ValueError: A already in B's depends: A.Failed

src/hera/workflows/task.py:84: ValueError
```

A better way to describe more complex conditions

Use `depends` directly

**Additional context**

elliotgunton:
**Pull Request Checklist**
- [x] Fixes #726, Fixes #1384
- [x] Tests added

**Description of PR**
* This allows users to set the default `depends` relationship to use between tasks
* It also introduces `Iterable[TaskResult]` to be able to set multiple values at once. This can be done through a list or the `__or__` operator (`|`):

```py
    with Task.set_next_defaults(on=TaskResult.succeeded | TaskResult.skipped):
        [task_a, task_b] >> task_c

    assert task_c.depends == "(task-a.Skipped || task-a.Succeeded) && (task-b.Skipped || task-b.Succeeded)"
```
> This change is also applied to the `next()` function itself, which fixes #1384 

Note the `__or__` that already exists on `Task` is untouched, and is only useful to set `depends` directly with just the task name (not `.Succeeded` etc), e.g.:
```py
    task_c.depends = (task_a | task_b)

    assert task_c.depends == "(task-a || task-b)"
```
This has been added as a test.

### DIFF
--- a/docs/walk-through/dags.md
+++ b/docs/walk-through/dags.md
@@ -271,4 +271,13 @@ C.depends = "A.Succeeded"
 D.depends = "B.Succeeded || C.Succeeded"
 ```
 
+The `set_next_defaults` function also accepts lists or values applying the `|` operator for the `on` value, meaning you can also specify conditions like `B.depends = "A.Succeeded || A.Daemoned"`, without affecting the `operator` used. E.g:
+
+```py
+with Task.set_next_defaults(operator=Operator.and_, on=TaskResult.succeeded | TaskResult.skipped):
+    [task_a, task_b] >> task_c
+
+assert task_c.depends == "(task-a.Skipped || task-a.Succeeded) && (task-b.Skipped || task-b.Succeeded)"
+```
+
 See the [DAG Configurable rshift example](../examples/workflows/dags/dag_configurable_rshift.md) for the full code!
--- a/docs/walk-through/dags.md
+++ b/docs/walk-through/dags.md
@@ -233,3 +233,42 @@ A   B
 | X |
 C   D
 ```
+
+## Configuring the Default "Next" Behaviour for `>>`
+
+Hera v5.24 added the `Task.set_next_defaults` function, allowing you to set the default `operator` and `on` values
+within a scoped context, which by extension allows you to configure the rshift (`>>`) behaviour.
+
+This is useful if you want `A >> B` to mean "run B _only if_ A succeeded", otherwise the
+[default depends logic](https://argo-workflows.readthedocs.io/en/latest/enhanced-depends-logic/) means `A >> B` is
+equivalent to "B depends on `A.Succeeded || A.Skipped || A.Daemoned`".
+
+By setting the values in `Task.set_next_defaults`, we can configure the rshift behaviour to use a different operator
+and TaskResult. Then, the following
+
+```py
+with Task.set_next_defaults(operator=Operator.or_, on=TaskResult.succeeded):
+    A >> [B, C] >> D
+```
+
+is equivalent to
+
+```py
+A.next(B, on=TaskResult.succeeded)
+A.next(C, on=TaskResult.succeeded)
+B.next(D, on=TaskResult.succeeded)
+C.next(D, operator=Operator.or_, on=TaskResult.succeeded)
+```
+
+> Note the `Operator.or_` for D's `depends` is set when calling `C.next` which can also be confusing! This is because we
+> use `next` to describe the forward relationships, while the Argo field is `depends` which describes the backward
+> relationships.
+
+Or, described using the backward relationship of `depends` (which only accepts strings):
+```py
+B.depends = "A.Succeeded"
+C.depends = "A.Succeeded"
+D.depends = "B.Succeeded || C.Succeeded"
+```
+
+See the [DAG Configurable rshift example](../examples/workflows/dags/dag_configurable_rshift.md) for the full code!
--- a/src/hera/workflows/task.py
+++ b/src/hera/workflows/task.py
@@ -130,7 +130,7 @@ class Task(
         on: OnType = None,
     ) -> Task:
         """Set self as a dependency of `other`."""
-        operator = operator if operator is not None else self.__class__._default_next_operator
+        operator = operator or self.__class__._default_next_operator
         on_list = _normalise_on(on, self.__class__._default_next_on)
 
         # Build condition string:
--- a/src/hera/workflows/task.py
+++ b/src/hera/workflows/task.py
@@ -8,7 +8,7 @@ from __future__ import annotations
 
 from contextlib import contextmanager
 from enum import Enum
-from typing import ClassVar, List, Optional, Union
+from typing import ClassVar, Iterable, Iterator, List, Optional, Set, Union
 
 from hera.workflows._mixins import (
     ArgumentsMixin,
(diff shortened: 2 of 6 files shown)


src/hera/workflows/task.py
```diff
@@ -71,22 +123,59 @@ def _get_dependency_tasks(self) -> List[str]:
     def _subtype(self) -> str:
         return "tasks"
 
-    def next(self, other: Task, operator: Operator = Operator.and_, on: Optional[TaskResult] = None) -> Task:
+    def next(
+        self,
+        other: Task,
+        operator: Optional[Operator] = None,
+        on: OnType = None,
```
elliotgunton, line 130: Note the change of types is backwards compatible and the behaviour is unchanged (the default `Operator.and_` is now stored as a class variable).

docs/walk-through/dags.md
```diff
@@ -233,3 +233,51 @@ A   B
...
+This is useful if you want `A >> B` to mean "run B _only if_ A succeeded", otherwise the
+[default depends logic](https://argo-workflows.readthedocs.io/en/latest/enhanced-depends-logic/) means `A >> B` is
+equivalent to "B depends on `A.Succeeded || A.Skipped || A.Daemoned`".
+
+By setting the values in `Task.set_next_defaults`, we can configure the rshift behaviour to use a different operator
+and TaskResult. Then, the following
+
+```py
+with Task.set_next_defaults(operator=Operator.or_, on=TaskResult.succeeded):
+    A >> [B, C] >> D
```
alicederyn, line 251: I'm not sure this is readable. "Next" isn't a Python term for `>>`, and changing the behaviour of "and" to "or" on lists on the left hand side of the operator is subtle and not obvious from the parameter name `operator`. I'd say we should only affect the `.next` function, except that now this only saves maybe a couple of redundant parameters on a few function calls.

Can we do this via a full proposal, considering all the alternatives?

elliotgunton: I think it's readable in the context of Hera, and works quite well -- Hera is effectively a DSL within Python, so the Python terms don't really matter here, "next" can be the Hera term for `>>` (or "then", as you'd read it out loud as "A then B and C, then D").

This is syntactic sugar on top of syntactic sugar, so you don't _need_ to use it, but IMO it's actually clearer than the alternative:

```py
A.next(B, on=TaskResult.succeeded)
A.next(C, on=TaskResult.succeeded)
B.next(D, on=TaskResult.succeeded)
C.next(D, operator=Operator.or_, on=TaskResult.succeeded)
```

because as described in the updated walkthrough (also in your comment), the `C.next` statement can be confusing as `operator` is referring to the `depends` of `D` (but that begs the question of what else could it be referring to? It just takes an extra moment of comprehension). Which leaves you with

```py
B.depends = "A.Succeeded"
C.depends = "A.Succeeded"
D.depends = "B.Succeeded || C.Succeeded"
```

for a no-syntactic-sugar-or-mypy-or-mental-load-problems solution.

I would consider a proposal for a V6 of Hera, as the `next` function, and `>>` behaviour, has been around since V0. Right now I'm happy with the feature, as the issue https://github.com/argoproj-labs/hera/issues/726 has been in the back of my mind since 2023.

src/hera/workflows/task.py
```diff
@@ -71,22 +123,59 @@ def _get_dependency_tasks(self) -> List[str]:
...
         return other
 
+    @classmethod
+    @contextmanager
+    def set_next_defaults(
+        cls,
+        operator: Operator = Operator.and_,
+        on: Union[TaskResult, Iterable[TaskResult], None] = None,
+    ):
+        """Temporarily override the default operator and on."""
```
alicederyn, line 166: Perhaps this might be clearer?
```suggestion
        """Temporarily modify the default behaviour of .next and >>."""
```

src/hera/workflows/task.py
```diff
@@ -71,22 +123,59 @@ def _get_dependency_tasks(self) -> List[str]:
...
+        on_list = _normalise_on(on)
+
+        old_operator = cls._default_next_operator
+        old_on = cls._default_next_on
+        cls._default_next_operator = operator
+        cls._default_next_on = on_list
```
alicederyn, line 172: nit: if I don't explicitly override `operator` or `on`, it might be surprising that any enclosing `set_next_defaults` would be overridden back to the default.

src/hera/workflows/task.py
```diff
@@ -55,6 +104,9 @@ class Task(
     dependencies: Optional[List[str]] = None
     depends: Optional[str] = None
 
+    _default_next_operator: ClassVar[Operator] = Operator.and_
+    _default_next_on: ClassVar[OnType] = None
```
alicederyn, line 108: We store the result of `_normalize_on` here, which is not the fully generic `OnType`. Storing an unknown iterable in this field would potentially be a bug, since iterables can be altered by iterating them, so we should specify the exact type to avoid a risk of issues in future.
```suggestion
    _default_next_on: ClassVar[Optional[List[TaskResult]]] = None
```

src/hera/workflows/task.py
```diff
@@ -42,6 +43,54 @@ class TaskResult(Enum):
...
+    def __iter__(self) -> Iterator[TaskResult]:
+        return iter(self._results)
+
+
+OnType = Optional[Union[TaskResult, Iterable[TaskResult]]]
+
+
+def _normalise_on(
+    on: OnType,
+    default: OnType = None,
```
alicederyn, line 76: If I'm understanding the code correctly, this should be
```suggestion
    default: Optional[List[TaskResult]] = None,
```

src/hera/workflows/task.py
```diff
@@ -42,6 +43,54 @@ class TaskResult(Enum):
...
+    Accepts:
+      - None -> return normalised default (which may also be None)
+      - TaskResult -> [TaskResult]
+      - Iterable[TaskResult] -> list(...)
+    """
+    if on is None:
+        if default is None:
+            return None
+
+        return _normalise_on(default)
```
alicederyn, line 89: If I'm understanding the code correctly, this will always be normalized already.
```suggestion
        return default
```

src/hera/workflows/task.py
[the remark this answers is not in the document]
elliotgunton: I'm not super confident with `ContextVar`, but had a go in the latest commit. Let me know if it looks right 🙏

src/hera/workflows/task.py
```diff
@@ -71,22 +122,59 @@ def _get_dependency_tasks(self) -> List[str]:
...
+
+        old_operator = _default_next_operator.get()
+        old_on = _default_next_on.get()
+        _default_next_operator.set(operator or _default_next_operator.get())
+        _default_next_on.set(on_list)
+        try:
+            yield
+        finally:
+            _default_next_operator.set(old_operator)
+            _default_next_on.set(old_on)
```
alicederyn, line 176: The preferred idiom is set/[reset](https://docs.python.org/3/library/contextvars.html#contextvars.ContextVar.reset)
```suggestion
        operator_token = _default_next_operator.set(operator or _default_next_operator.get())
        on_token = _default_next_on.set(on_list)
        try:
            yield
        finally:
            _default_next_operator.reset(operator_token)
            _default_next_on.reset(on_token)
```

alicederyn (approved):
LGTM, just a minor idiomatic cleanup

PR #1466, opened 2025-08-04
Update on-cluster testing

### LINKED ISSUE #1163: Add a test and make target for argo linting on-cluster
Quite a few issues thrown up by [new DAGs](https://github.com/argoproj-labs/hera/issues/1162) would have been surfaced by using an `argo lint`. We already have on-cluster tests so this shouldn't be too hard to add.

We should use the Hera API to do some more dogfooding, as opposed to the `argo` CLI.

### LINKED ISSUE #1350: Add a test and make target to run all workflow examples on local cluster
It would be good to ensure all custom workflows are valid and runnable on a local installation e.g. docker desktop. We could add a build step for runner scripts.

It shouldn't be a test as that would make PR checks take too long

### LINKED ISSUE #1397: On cluster tests stopped working
* Seems to be because Minio script doesn't create the `workflows` bucket that's automatically used by Artifacts (hence the DAG artifact example no longer passes) `message='Error (exit code 64): failed to put file: The specified bucket does not exist; '`

elliotgunton:
**Pull Request Checklist**
* Fixes #1397 
* Fixes #1350 
* Fixes #1163 

- [x] Tests added

**Description of PR**
This PR adds an improved "on cluster" test harness, which will run all non-upstream examples.
* CI will not run all the non-upstream examples, but will continue to run the 3 basic examples.
* CI will also  `lint` all the non-upstream examples (which runs in around 1 second on local).

Added `pytest-xdist` specifically for this test otherwise there would be a lot of sitting around waiting. I updated examples which could be easily fixed, and created a list of examples which should be skipped for linting and submission (the submission set subsumes the linting set). (Only ~one~ two examples are intentionally skipped for linting!)


tests/submissions/test_examples.py
```diff
@@ -0,0 +1,169 @@
...
+from hera.workflows.workflow_template import WorkflowTemplate
+from tests.test_examples import CI_MODE, _get_examples
+
+# Skip long-running and unsupported examples
+TIME_LIMIT_SECONDS = 120
+SKIP_LINT_EXAMPLES = {
+    "new_decorators_auto_template_refs",  # Uses TemplateRefs to non-existent WorkflowTemplates
+    "parquet_pandas",
+}
+SKIP_SUBMISSION_EXAMPLES = SKIP_LINT_EXAMPLES.union(
```
alicederyn, line 23: did we want to maybe sort these alphabetically?

tests/submissions/test_examples.py
```diff
@@ -0,0 +1,169 @@
...
+    module_name,
+    global_config_fixture,
+    workflows_service: WorkflowsService,
+):
+    global_config_fixture.host = "https://localhost:2746"
+    workflow = importlib.import_module(module_name).w
+
+    # Skip anything not a Workflow
+    if isinstance(workflow, (WorkflowTemplate, ClusterWorkflowTemplate, CronWorkflow)):
+        assert False, "Add non-workflow example to 'SKIP_SUBMISSION_EXAMPLES' set"
```
alicederyn, line 106: note: [pytest.fail](https://docs.pytest.org/en/stable/reference/reference.html#pytest.fail) is also an option

tests/submissions/test_examples.py
elliotgunton: Might as well 👍

PR #1470, opened 2025-08-05
Use `Optional` for 3.9 compatibility

elliotgunton:
Not sure why it wasn't picked up in the #1465 CI checks, but the newer union syntax broke it for other PRs.


alicederyn:
That is concerning that it passed CI :/

Issue #1254, opened 2024-11-01
Relax context handler behavior


flaviuvadan:
## Plan

1. Improve documentation on context handlers
2. Consider making Hera accept inline within context definitions of not only scripts but containers, resource, etc.

### Discussed in https://github.com/argoproj-labs/hera/discussions/1252

<div type='discussions-op-text'>

<sup>Originally posted by **daturkel** October 28, 2024</sup>
I'm a bit confused about the behavior of context managers in Hera and exactly how they behave under the hood.

If I have a python prelude like this:
```python
from hera.workflows import Container, Steps, Workflow

def get_container():
    my_container = Container(
        image="my-image",
        command=["bash", "-c"],
        args=["echo hello"],
    )
    return my_container
```

There are a couple ways I might be inclined to use this container factory function.

I could declare the Container in global scope, which works:
```python
container = get_container()
with Workflow(name="my-workflow", entrypoint="entrypoint") as my_workflow:
    with Steps(name="entrypoint") as my_steps:
        container(name="foo")
```

I could declare the Container in the Workflow scope, which also works, though I'm not sure what the advantage is:
```python
with Workflow(name="my-workflow", entrypoint="entrypoint") as my_workflow:
    container = get_container()
    with Steps(name="entrypoint") as my_steps:
        container(name="foo")
```

Lastly, I could declare the container in the Steps scope, which does *not* work:
```python
with Workflow(name="my-workflow", entrypoint="entrypoint") as my_workflow:
    with Steps(name="entrypoint") as my_steps:
        container = get_container()
        container(name="foo")

# hera.workflows.exceptions.InvalidType: <class 'hera.workflows.container.Container'>
```

I would love some information on the advantages of the Workflow and Steps context managers and when to use them, and why the third snippet here doesn't work. There doesn't seem to be a ton in the Hera docs on this topic (though maybe I've missed it) and I wouldn't be against contributing some documentation, but first I'll have to understand the behavior better!

There's some relevant discussion on documentation for idiomatic workflows in this issue: https://github.com/argoproj-labs/hera/issues/440</div>

PR #1472, opened 2025-08-06
Fix incorrect types and comments throughout mixins

elliotgunton:
Was intending to look into #1254, but found some odd types that looked wrong. This PR is mainly a refactor to fix types, with a minimal addition of the `raise ValueError("Only 'template' is supported (not inline or template_ref)")` snippets (as that would be the correct fix which was unearthed through these type fixes).

To note:
* TTemplate had the wrong description and so was used incorrectly in places where the `isinstance` check would never be true
* Prefer `Templatable` as a type over `TemplateMixin`


src/hera/workflows/_context.py
```diff
@@ -22,14 +20,12 @@ class SubNodeMixin(BaseMixin):
     """SubNodeMixin ensures that the class gets added to the Hera context on initialization.
 
     The mixin implements the core Hera `__hera_init__`, which is invoked post Hera object initialization. Anything
-    that inherits from this mixin has the capacity to manage a context via either being added to a context (like being
-    added to a Workflow/DAG) or managing a context itself (like holding Container, Script, etc).
+    that inherits from this mixin will add itself to the managed context (e.g. added to a Workflow/DAG).
     """
 
-    def __hera_init__(self: TNode) -> TNode:
+    def __hera_init__(self: "SubNodeMixin"):
```
alicederyn, line 26: Does this work? Typing `self` isn't normally required, and not typing the return as `None` can cause issues.
```suggestion
    def __hera_init__(self) -> None:
```

src/hera/workflows/_meta_mixins.py
```diff
@@ -427,10 +425,10 @@ def __call__(self, *args, **kwargs) -> Union[None, Step, Task]:
                     f"Callable Template '{self.name}' is not callable under a Workflow"  # type: ignore
                 )
             if isinstance(_context.pieces[-1], (Steps, Parallel)):
-                return Step(template=self, **kwargs)
+                return Step(template=cast(Templatable, self), **kwargs)
```
alicederyn, line 428: should this class maybe extend Templatable?

elliotgunton: I originally wanted to have `TemplateMixin` itself extend `Templatable`, but `Protocol` seems incompatible with `BaseModel`:

```
TypeError: metaclass conflict: the metaclass of a derived class must be a (non-strict) subclass of the metaclasses of all its bases
```

alicederyn (approved):

PR #1490, opened 2025-08-19
Add example using `with_items` for step/dag decorators

elliotgunton:
Received a question on how to perform fan outs using the decorator syntax. It is already possible, so I added an example and expanded the user guide. Also added a note on what's keeping decorator syntax as "experimental", as I've received a few questions around that too - essentially we just want more feedback and to improve the UX as much as possible before graduating the feature.


docs/user-guides/decorators.md
```diff
@@ -168,5 +168,44 @@ def my_steps() -> None:
...
+def print_message(inputs: PrintMessageInput):
+    print(inputs.message)
+
+
+@w.steps()
+def loop_example():
+    print_message(
+        PrintMessageInput(message="{{item}}"),
+        name="print-message-loop-with-items",
+        with_items=["hello world", "goodbye world"],
```
alicederyn, line 198: Will this call `print_message` twice if I run `loop_example` directly in Python?

elliotgunton: No, it will just print `{{item}}`:
```py
@w.steps()
def loop_example():
    print_message(
        PrintMessageInput(message="{{item}}"),
        name="print-message-loop-with-items",
        with_items=["hello world", "goodbye world"],
    )


if __name__ == "__main__":
    loop_example()
```

Output:
```
$ python -m hera_scratch.workflow
{{item}}
```

alicederyn: Then I think this violates the intention of the decorator syntax. Quoting the guide:

> The dag and steps decorator bring brand-new functionality to Hera, as they allow you to run and test your dag and steps functions entirely locally.

As well as not working as intended at runtime, the `PrintMessageInput` call only happens to work at all in this example because `message` is typed as a `string`. If it was typed as anything else, Pydantic would fail validation.

I think it would be much closer to the intent of decorator syntax if we just said this wasn't supported, and left people to use the old syntax.

elliotgunton: There is also the next line:

> Note that as you are running pure Python code, features of Argo Workflows are not supported when running locally, such as expressions.

This includes loops. We also already have https://github.com/argoproj-labs/hera/issues/1172 for addressing the fact that local running is not a great experience with the dag/steps decorators (yet). But it is at least a step forwards from the context syntax.

I agree it's only because `message` is a `str`, but the functionality is already there, and was designed to work this way, so it is technically supported, just never documented.

In the case of a non-str, you can `construct` the Pydantic class:

```py
@w.script()
def print_message(inputs: PrintMessageInput):
    print(inputs.an_int)


@w.set_entrypoint
@w.steps()
def loop_example():
    print_message(
        PrintMessageInput.construct(an_int="{{item}}"),
        name="print-message-loop-with-items",
        with_items=[42, 123],
    )
```

And it will work on the cluster, which is the main thing Hera is for, and as a custom DSL we can break some rules :)

And as an experimental feature, I'm fine with the above until we find better solutions.

alicederyn: Then I shall [disagree and commit](https://en.wikipedia.org/wiki/Disagree_and_commit). I think we should explain how this will behave if run locally in the docs before we merge.

alicederyn (approved):

PR #1500, opened 2025-08-29
Ensure namespace is present for list_workflow_templates API

jaseemabid:
**Pull Request Checklist**

- [X] Fixes #
- [ ] Tests added

**Description of PR**

Neither global config or this client ensures that namespace is non empty here, so you end up making the wrong API call to
https://argo-workflows/api/v1/workflow-templates/None; which is IMO worse than blowing up here with an assertion.

Fail rather than be incorrect


jaseemabid:
Turns out the argo workflows docs are wrong, this param is not needed. 

https://github.com/argoproj/argo-workflows/issues/14799

If the namespace is empty, hera client should omit it rather than set it to None, which is a separate bug.

jaseemabid:
Thanks @elliotgunton, let me see if I can get the upstream schema fixed.

PR #1534, opened 2025-12-03
Migrate to mkdocstrings 2.0

elliotgunton:
Fixes build breakages in rtd. Maybe we should add bounds in `docs/requirements.txt`? Not too bothered.


docs/requirements.txt
```diff
@@ -5,3 +5,4 @@ mkdocs-include-markdown-plugin
 mkdocs-click
 mkdocstrings[python]
 mkdocs-awesome-pages-plugin
+ruff
```
elliotgunton: Now required for formatting function sigs

Issue #943, opened 2024-01-27
Robust validation for k8s resource requirements


KengoA:
**Is your feature request related to a problem? Please describe.**

Currently the validation for `Resources` class https://github.com/argoproj-labs/hera/blob/main/src/hera/workflows/resources.py misses some cases of invalid configurations in k8s.

**Describe the solution you'd like**
- Decimal units are parsed for string values for CPU requirements such as `500m` and `0.5`
- Ensuring `request <= limit` for memory, ephemeral storage, and CPU specified in the Resources class
  - For CPU, float values are compared and not just int values, currently `cpu_request = 0.4` and `cpu_limit = 0.2` does NOT raise AssertionError as float instances are not compared.

PR #1584, opened 2026-05-21
fix: validate mixed CPU resource units

immanuwell:
## What

Tiny fix for a sneaky Resources edge case.

`cpu_request` and `cpu_limit` were compared only when both values used the same Python type. So mixed valid Kubernetes CPU units could slip through, kinda yikes:

```py
from hera.workflows import Resources

Resources(cpu_request=1, cpu_limit="500m").build()
Resources(cpu_request="1000m", cpu_limit=0.5).build()
```

Both build invalid manifests where the CPU request is greater than the limit. Kubernetes rejects that at submit time.

This normalizes both CPU values before comparing them, while keeping valid mixed equal values working:

```py
Resources(cpu_request=0.5, cpu_limit="500m")
Resources(cpu_request="500m", cpu_limit=0.5)
```

Related context: #943 / #944

## Checks

```text
pytest tests/test_unit/test_resources.py --no-cov
ruff format --check src/hera/workflows/resources.py tests/test_unit/test_resources.py
ruff check src/hera/workflows/resources.py tests/test_unit/test_resources.py
```


src/hera/workflows/resources.py
```diff
@@ -25,6 +25,12 @@ def _merge_dicts(a: Dict, b: Dict, path=None):
     return a
 
 
+def _convert_cpu_value(value: Union[float, int, str]) -> float:
```
elliotgunton:
```suggestion
def _normalise_cpu_value(value: Union[float, int, str]) -> float:
```

src/hera/workflows/resources.py
```diff
@@ -84,17 +90,15 @@ def __post_init__(self):
...
         if self.cpu_request is not None and isinstance(self.cpu_request, str):
             validate_cpu_units(self.cpu_request)
         if self.cpu_limit is not None and isinstance(self.cpu_limit, str):
             validate_cpu_units(self.cpu_limit)
-            if self.cpu_request is not None and isinstance(self.cpu_request, str):
-                if convert_cpu_units(self.cpu_request) > convert_cpu_units(self.cpu_limit):
-                    raise ValueError("CPU request must be smaller or equal to limit")
+
+        if self.cpu_request is not None and self.cpu_limit is not None:
+            if _convert_cpu_value(self.cpu_request) > _convert_cpu_value(self.cpu_limit):
```
elliotgunton:
```suggestion
            if _normalise_cpu_value(self.cpu_request) > _normalise_cpu_value(self.cpu_limit):
```

elliotgunton (commented):
Just gonna update the naming here and then approve, nice fix!

elliotgunton (approved):
🚀

PR #1588, opened 2026-05-28
fix: converters reject signed resource values to match validator behavior

immanuwell:
**Pull Request Checklist**
- [ ] Fixes #
- [x] Tests added

**Description of PR**

Currently, `_convert_decimal_units` and `_convert_binary_units` accept signed values like `-500m` or `+1Gi` due to `[+-]?` in their regex patterns. The validators (`_validate_decimal_units`, `_validate_binary_units`) correctly reject these, so there's a mismatch - same input, different outcome depending on which function you hit first.

This PR removes `[+-]?` from both converter patterns so they're consistent with validators. Also fixes `_merge_dicts` raising bare `Exception` instead of `ValueError`.

Reproduce:

```python
from hera.workflows.converters import _convert_decimal_units, _convert_binary_units
from hera.workflows.validators import _validate_decimal_units

_validate_decimal_units("-500m")   # raises ValueError (correct)
_convert_decimal_units("-500m")    # returns -0.5 (bug)

_convert_binary_units("-512Mi")    # returns -536870912.0 (bug)
```

Negative resource values aren't valid in Kubernetes anyway, so rejecting them in both layers makes sense.

### DIFF
--- a/src/hera/workflows/converters.py
+++ b/src/hera/workflows/converters.py
@@ -35,7 +35,7 @@ def _convert_decimal_units(value: str) -> float:
     Returns:
         float: Float value of the given decimal units.
     """
-    pattern = r"^\s*([+-]?\d+(?:\.\d+)?)([mkMGTPE]?)\s*$"
+    pattern = r"^\s*(\d+(?:\.\d+)?)([mkMGTPE]?)\s*$"
     matches = re.match(pattern, value)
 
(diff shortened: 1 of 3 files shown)


elliotgunton:
Hi @immanuwell - It doesn't really make sense for a converter function to perform validation, so I won't be merging this PR.

Issue #1446, opened 2025-07-23
YAML-to-Python code generator improvements


elliotgunton:
The "v1" of the YAML-to-Python generator is due to be released. It is a bit rough around the edges in terms of generating Pythonic code, so this is a list of potential issues described in https://github.com/argoproj-labs/hera/pull/1444, roughly listed in low-to-high effort order:

* Skip `api_version` and `kind` in the `Workflow` params
* Split long parameter value strings
* Generate arguments in dict format
* Sort Python arguments using the same algorithm as YAML keys (`name` first, then basic types, then lists/dicts)
* Generate script-decorated functions for `script` templates using a Python image/command

Completed:
* ~Update `invocator` context name to use `steps`/`dag`/`container_set`~ Completed in #1465
* ~Skip trailing comma to allow formatter to decide whether to collapse list of args into one line~ Completed in #1465

PR #1592, opened 2026-06-19
fix(cli): omit default workflow params in generated Python

immanuwell:
- [x] Part of #1446
- [x] Tests added

Description of PR
This trims redundant args from `hera generate python` output.

Right now a tiny workflow like `tests/cli/examples/single_workflow.yaml` turns into:
```python

with Workflow(api_version="argoproj.io/v1alpha1", kind="Workflow", name="single") as w:
    pass
```

That works, but its kinda noisy. Hera already fills those defaults for us.

This PR skips `kind` when it matches the generated Hera class, and skips `api_version` when it matches the default config. Non default `apiVersion` still stays in the output, so roundtrip behavior is fine.

Repro:
```bash
. /tmp/hera-pr-venv/bin/activate
PYTHONPATH=src python - <<'PY'
from pathlib import Path
from hera._cli.generate.python import load_yaml_workflows, workflow_to_python
model = next(load_yaml_workflows(Path('tests/cli/examples/single_workflow.yaml')))
print(workflow_to_python(model))
PY
```

Checks run:
```bash
. /tmp/hera-pr-venv/bin/activate
PYTHONPATH=src pytest tests/cli/test_generate_python.py -q
```

### DIFF
--- a/src/hera/_cli/generate/python.py
+++ b/src/hera/_cli/generate/python.py
@@ -11,6 +11,7 @@ from pydantic import RootModel
 
 from hera._cli.base import GeneratePython
 from hera._cli.generate.util import YAML_EXTENSIONS, convert_code, expand_paths, write_output
+from hera.shared import global_config
 from hera.shared._pydantic import APIBaseModel
 from hera.shared._type_util import (
     get_annotated_metadata,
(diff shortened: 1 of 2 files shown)


elliotgunton (approved):
Nice one! 🚀

PR #1593, opened 2026-06-22
fix(cli): sort generated Python kwargs for readability

immanuwell:
Pull Request Checklist
- [x] Part of #1446
- [x] Tests added

Description of PR
Part of #1446.

`hera generate python` already emits valid code, but the arg order is kinda noisy rn. Nested bits like `inputs` can show up before `name` and simple scalar fields, so the generated code is harder to scan.

This keeps behavior the same and just sorts generated kwargs for readability. `name` goes first, then scalar values, then nested calls, then lists and dicts. Small QoL fix, cleans things up nicely.

Repro:
```bash
.venv/bin/python -m hera generate python examples/workflows/upstream/http-success-condition.upstream.yaml
```

Before:
```python
HTTP(
    inputs=Inputs(parameters=[Parameter(name="url")]),
    name="http-status-is-201",
    success_condition="response.statusCode == 201",
    url="{{inputs.parameters.url}}",
)
```

After:
```python
HTTP(
    name="http-status-is-201",
    success_condition="response.statusCode == 201",
    url="{{inputs.parameters.url}}",
    inputs=Inputs(parameters=[Parameter(name="url")]),
)
```

Checks run:
```bash
.venv/bin/python -m pytest tests/cli/test_generate_python.py --no-cov -q
```

### DIFF
--- a/src/hera/_cli/generate/python.py
+++ b/src/hera/_cli/generate/python.py
@@ -217,7 +217,7 @@ class WorkflowPythonBuilder:
             context_expr=ast.Call(
                 func=ast.Name(id=hera_workflow_class.__name__, ctx=ast.Load()),
                 args=[],
-                keywords=keywords,
+                keywords=self._sort_keywords(keywords),
             ),
             optional_vars=ast.Name(id="w", ctx=ast.Store()),
         )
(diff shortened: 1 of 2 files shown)


elliotgunton (approved):
🚀

PR #1610, opened 2026-09-07
fix(cli): handle empty and missing-kind YAML documents

emecii:
## Summary

- Ignore implicit empty YAML documents created by leading, consecutive, or trailing separators.
- Raise a clear `ValueError` when a YAML mapping does not declare `kind`, instead of exposing a `KeyError`.
- Preserve validation for explicit nulls, scalars, lists, and unsupported kinds.

## Validation

- `pytest -k cli`: 260 passed, 2 skipped, 11 xfailed
- Focused YAML-generator regressions: 18 passed
- `ruff check` and `ruff format --check` on changed files
- `git diff --check`

The non-CLI suite has six pre-existing/environmental failures in upstream-example round-trips and a missing `pip` executable; the generator path is not involved.

### DIFF
--- a/src/hera/_cli/generate/python.py
+++ b/src/hera/_cli/generate/python.py
@@ -78,20 +78,37 @@ def generate_python(options: GeneratePython):
 
 def load_yaml_workflows(path: Path) -> Generator[ModelWorkflow, None, None]:
     """Load the YAML file containing a Workflow(s)."""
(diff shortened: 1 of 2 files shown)


elliotgunton:
IMO this is code bloat and unnecessary for a simple YAML loading script

